PR #14, opened 2025-10-28
[RSPEED-2069] Simplify tool definition and logging

samdoran:
Create a decorator that can be used to easily wrap tool calls instead of having to call a wrapper function only for the purpose of logging.

Reorganize imports so that tools are automatically added instead of having to write new wrapped functions for every tool call.

Use a consistent name for the logger.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -1,6 +1,6 @@
 codecov:
   notify:
-    after_n_builds: 7
+    after_n_builds: 6
     wait_for_ci: false
 
   require_ci_to_pass: false


src/linux_mcp_server/audit.py
```diff
@@ -90,43 +94,68 @@ def process(self, msg, kwargs):
...
-        message += f" | {params_str}"
+        func(*args, **kwargs)
 
-    logger.info(message, extra=extra)
+    @functools.wraps(func)
+    async def awrapper(*args, **kwargs):
+        execution_mode = "remote" if kwargs.get("host") else "local"
+        safe_params = sanitize_parameters(kwargs)
+
+        extra = {
```
narmaku: Could we extract this to a helper function, even if it's within log_tool_call scope?
I see this repeated in both wrappers. same with formatting of `params_str` and `message`.

src/linux_mcp_server/server.py
```diff
@@ -1,438 +1,17 @@
...
-        result = await handler(**kwargs)
-        duration = time.time() - start_time
-        log_tool_complete(tool_name, status="success", duration=duration)
-        return result
-
-    except Exception as e:
-        duration = time.time() - start_time
-        log_tool_complete(tool_name, status="error", duration=duration, error=str(e))
-        raise
+from linux_mcp_server.tools import *  # noqa: E402, F403
```
narmaku, line 14: Amazing :clap:

src/linux_mcp_server/tools/__init__.py
```diff
@@ -1 +1,44 @@
-"""Tool modules for Linux diagnostics."""
+from linux_mcp_server.tools.logs import get_audit_logs
```
narmaku, line 2: Call me picky, but I would prefer to have a new line separating the modules. It groups them visually so it's easier for humans to read. Not sure if that will interfere with linting/formatting though.

src/linux_mcp_server/tools/logs.py
```diff
@@ -4,10 +4,14 @@
 
 from pathlib import Path
 
-from .ssh_executor import execute_command
-from .validation import validate_line_count
+from linux_mcp_server.audit import log_tool_call
+from linux_mcp_server.server import mcp
+from linux_mcp_server.tools.ssh_executor import execute_command
```
narmaku, line 9: Maybe not for this MR which is already big enough, but something to consider in another refactoring session:
This ssh_executor might be taken out of the tools dir. It's more a helper/util that is used by all the tools, but not a tool set by itself. WDYT?

src/linux_mcp_server/tools/processes.py
```diff
@@ -6,22 +6,28 @@
...
+from linux_mcp_server.tools.ssh_executor import execute_command
+from linux_mcp_server.utils import format_bytes
+from linux_mcp_server.utils.validation import validate_pid
 
 
+@log_tool_call
+@mcp.tool(
+    title="List processes",
+    description="List running processes",
+    annotations=ToolAnnotations(readOnlyHint=True),
```
narmaku, line 22: Interesting. Does this mean we are letting the LLM know that this tool performs read-only operations?

narmaku: I think this answers my question: https://modelcontextprotocol.io/legacy/concepts/tools

Should we then add this to all the tools? Or would that be too much of metadata overload?

src/linux_mcp_server/tools/ssh_executor.py
```diff
@@ -5,6 +5,8 @@
 either local or remote execution based on the provided parameters.
 """
 
+# FIXME: Move this out of tools and into its own package.
```
narmaku, line 8: LOL this solves one of my previous comments.

narmaku (changes requested):
Good stuff! Adding a comment that applies to all the tools:

The tools docstrings/descriptions (in the mcp decorator) must be more verbose and explanatory so the LLMs can understand them better.
The way it was defined before was to have a docstring in the async functions decorated with `@mcp.tool()` in `server.py`.

E.g. for `get_hardware_info`, the docstring said: "Get hardware information including CPU architecture, PCI devices, USB devices, and memory hardware."

But now, in the `system_info.py` module, the same tool just says "Get hardware information.". --> here is the issue. The LLMs will not know how much information they may retrieve. Also I believe we should add it in the mcp decorator `description` param.

Suggestion: To preserve current behavior with LLMs, let's make sure to use tool title/name and descriptions for each of them (in mcp decorator), using the same docstrings as in `server.py` before this refactoring.

src/linux_mcp_server/server.py
samdoran: I spent a couple of days during the hackathon week trying figure out how to do this.

src/linux_mcp_server/tools/__init__.py
samdoran: The formatter will complain, but we can work around that. I believe in making formatters work _for_ us, not the other way around.

src/linux_mcp_server/tools/logs.py
samdoran: My thoughts exactly! You found my `FIXME`.

src/linux_mcp_server/tools/processes.py
samdoran: Yes! Rather than doc strings we should use parameters to the `tools` decorator. The lets us separate what's for human consumption and what is meant for the LLM.

I did a few as an example and wasn't sure if we should handle the rest in a followup PR since this once is already getting huge.

src/linux_mcp_server/audit.py
samdoran: Potentially. It gets a bit tricky because the two functions are identical, but one awaits a coroutine while the other just runs the function. I haven't come up with a way to reduce the duplication.

src/linux_mcp_server/tools/__init__.py
samdoran: I got this fixed. It does look way better now.

src/linux_mcp_server/audit.py
samdoran: I see how to do it. Fix incoming.

samdoran: It should be much better now. I also added a follow up call which logs when the function completes and adjusted the duration calculation and display a bit.

src/linux_mcp_server/tools/processes.py
samdoran: Here are the [arguments that tool accepts](https://gofastmcp.com/servers/tools#decorator-arguments).

samdoran: Though we're using the [Python SDK](https://github.com/modelcontextprotocol/python-sdk?tab=readme-ov-file#tools), not FastMCP. They're somewhat related.

samdoran: I will take care of this in another PR.

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +96,109 @@ def process(self, msg, kwargs):
...
-
-    # Determine if local or remote execution
-    execution_mode = "remote" if parameters.get("host") else "local"
-
-    # Sanitize parameters
-    safe_params = sanitize_parameters(parameters)
+def _log_event_start(
+    logger: logging.Logger,
+    tool_name: str,
+    execution_mode: str,
```
subpop: Doesn't the value of this parameter immediately get discarded by the first line of the function? Should this argument exist at all?

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +96,109 @@ def process(self, msg, kwargs):
...
+    params: dict[t.Any, t.Any],
+) -> int:
+    execution_mode = "remote" if params.get("host") else "local"
+    safe_params = sanitize_parameters(params)
 
-    # Build log record with extra fields
     extra = {
         "event": "TOOL_CALL",
```
subpop: I feel like this event value should match the name of the method somehow. `_log_event_tool_call`? Also would it make sense to move these strings into an enum?

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +96,109 @@ def process(self, msg, kwargs):
...
-    execution_mode = "remote" if parameters.get("host") else "local"
-
-    # Sanitize parameters
-    safe_params = sanitize_parameters(parameters)
+def _log_event_start(
+    logger: logging.Logger,
+    tool_name: str,
+    execution_mode: str,
+    params: dict[t.Any, t.Any],
+) -> int:
```
subpop, line 121: Does this need a docstring? Especially considering it's returning a timespan of some sort. We should definitely document what the timespan means.

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +96,109 @@ def process(self, msg, kwargs):
...
-        extra["host"] = parameters["host"]
-    if "username" in parameters:
-        extra["username"] = parameters["username"]
+    if "username" in params:
+        extra["username"] = params["username"]
 
-    # Add sanitized parameters as string
     params_str = ", ".join(f"{k}={v}" for k, v in safe_params.items() if k not in ["host", "username"])
-
     message = f"TOOL_CALL: {tool_name}"
```
subpop: There's a redundant string here. We have the event name "TOOL_CALL" and this static string value. Are they meant to be the same?

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +96,109 @@ def process(self, msg, kwargs):
...
+    logger: logging.Logger,
+    tool_name: str,
+    start: int,
+    error: Exception | None = None,
+) -> None:
+    stop = time.perf_counter_ns()
+    duration = timedelta(microseconds=(stop - start) / 1_000)
+    status = "error" if error else "success"
     extra = {
         "event": "TOOL_COMPLETE",
```
subpop: Similar comment here about moving events into an enum.

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +96,109 @@ def process(self, msg, kwargs):
...
+def log_tool_call(func: t.Callable) -> Function:
+    """Decorator to log tool calls
+
+    Works with sync or async functions.
+    """
+    logger = logging.getLogger("linux-mcp-server")
+    tool_name = func.__name__
+
+    @functools.wraps(func)
+    def wrapper(*args, **kwargs):
```
subpop, line 189: The code in `wrapper` and `awrapper` looks identical. Could these be combined somehow?

src/linux_mcp_server/audit.py
samdoran: Yup. Looks like I only have removed that. Every time I read through this code I find more ways it could be cleaned up.

src/linux_mcp_server/audit.py
samdoran: The full log message includes the name of the tool being called. The actual function doing the calling seems more of an implementation detail more useful to the developer than the user.

```
2025-11-05 16:20:41 | INFO | linux-mcp-server | TOOL_CALL: list_services | event=TOOL_CALL | tool=list_services | execution_mode=local
2025-11-05 16:20:41 | INFO | linux-mcp-server | TOOL_COMPLETE: list_services | event=TOOL_COMPLETE | tool=list_services | status=success | duration=0:00:00.000001s
```

samdoran: I wouldn't mind making these values enums. Currently it's only used in this one file apart from tests.

samdoran: I was thinking that `execution_mode` should be an enum.

samdoran: I added a `StrEnum` class and made these values into enums.

src/linux_mcp_server/audit.py
samdoran: I added a doc string.

src/linux_mcp_server/audit.py
samdoran: Yeah, that's funny. I noticed that when I was switching these to enums.

I think the logging machinery needs reworking as well. We shouldn't be building big long strings. We should be passing around structured data through the logging pipeline and have the emitter handle presentation.

I'll remove it from the `extra` parameters for now.

src/linux_mcp_server/audit.py
samdoran: I found a bunch of these and made them all enums. Thanks for pointing this out.

src/linux_mcp_server/audit.py
samdoran: I pulled out all I could into the `_log_event_*` functions. I don't think it can be further simplified mainly because they have to be two separate function definitions, one sync and the other async.

It may be possible but I don't see how to do it right now.

src/linux_mcp_server/audit.py
```diff
@@ -94,6 +114,66 @@ def process(self, msg, kwargs):
...
+    params_str = ", ".join(f"{k}={v}" for k, v in safe_params.items() if k not in ["host", "username"])
+    if params_str:
+        message += f" | {params_str}"
+
+    logger.info(message, extra=extra)
+
+    return time.perf_counter_ns()
+
+
+def _log_event_complete(
```
narmaku, line 152: Maybe add a simple docstring here?

tests/test_audit.py
```diff
@@ -118,76 +108,75 @@ def emit(self, record):
 class TestLogToolCall:
     """Test tool call logging."""
 
-    def test_log_tool_call_local(self, caplog, decorated):
-        """Test logging a local tool call."""
+    @pytest.mark.parametrize(
+        ("params", "mode"),
+        (
+            ({}, "local"),
```
narmaku:
```suggestion
            ({}, ExecutionMode.local),
```

tests/test_audit.py
```diff
@@ -118,76 +108,75 @@ def emit(self, record):
...
         params = {"password": "secret123", "username": "admin"}
         with caplog.at_level(logging.INFO):
             decorated(**params)
 
-        assert "TOOL_CALL" in caplog.text
+        assert Event.TOOL_CALL in caplog.text
         assert "secret123" not in caplog.text
         assert "REDACTED" in caplog.text
 
+    async def test_log_tool_call_async_sanitizes_parameters(self, caplog, adecorated):
```
narmaku: We could do:

```python
from linux_mcp_server.audit import SENSITIVE_FIELDS
...
    async def test_log_tool_call_async_sanitizes_parameters(self, caplog, adecorated):
        """Test that tool call logging sanitizes sensitive parameters."""
        params = {v: "secret123" for v in SENSITIVE_FIELDS}
        params["username"] = "admin"
```

narmaku: Same for the non-async test case.

src/linux_mcp_server/audit.py
```diff
@@ -31,6 +35,22 @@
...
+    REMOTE_EXEC_ERROR = "REMOTE_EXEC_ERROR"
+    SSH_AUTH_FAILED = "SSH_AUTH_FAILED"
+    SSH_CONNECT = "SSH_CONNECT"
+    SSH_CONNECTING = "SSH_CONNECTING"
+    TOOL_CALL = "TOOL_CALL"
+    TOOL_COMPLETE = "TOOL_CALL"
+
+
+class ExecutionMode(StrEnum):
+    remote = "remote"
```
narmaku: Call me picky, but if we defined Event enum strings all with uppercase, why do't we do the same for local/remote? Just for consistency. Feel free to discard this NIT comment.

src/linux_mcp_server/audit.py
```diff
@@ -94,6 +114,66 @@ def process(self, msg, kwargs):
...
+
+    logger.info(message, extra=extra)
+
+    return time.perf_counter_ns()
+
+
+def _log_event_complete(
+    logger: logging.Logger,
+    tool_name: str,
+    start: int,
```
narmaku: Maybe rename to start_time?

narmaku (commented):
LGTM, just NIT comments.

src/linux_mcp_server/audit.py
subpop: I can agree with this. Uppercasing these kind of mode values makes sense to me.

src/linux_mcp_server/audit.py
```diff
@@ -90,80 +114,111 @@ def process(self, msg, kwargs):
...
+        result = None
+
+        try:
+            result = await func(*args, **kwargs)
+        except Exception as exc:
+            error = exc
+
+        _log_event_complete(logger, tool_name, start, error)
+
+        return result
```
subpop, line 217: Does something as simple as a DRY-reduction function not work here? The special method names `wrapper` and `awrapper` are still preserved, but they share a common backing implementation.

```suggestion
    def _common_wrapper(*args, **kwargs):
        start = _log_event_start(logger, tool_name, kwargs)
        error = None
        result = None

        try:
            result = func(*args, **kwargs)
        except Exception as exc:
            # FIXME: This could potentially swallow exceptions. Maybe narrow the exception type.
            error = exc

        _log_event_complete(logger, tool_name, start, error)

        return result

    @functools.wraps(func)
    def wrapper(*args, **kwargs):
        return _common_wrapper(args, kwargs)

    @functools.wraps(func)
    async def awrapper(*args, **kwargs):
        return _common_wrapper(args, kwargs)
```

tests/test_audit.py
samdoran: Oh, that's a good idea.

src/linux_mcp_server/audit.py
samdoran: I don't feel strongly one way on another. Enums in Python are more akin to attributes than constants so the capitalization is a lot more flexible. By convention, module globals are usually the only thing that is uppercased (there are common exceptions to that, though).

I mainly prefer snake case for attributes and that's how I think of enum values in Python. I only uppercased the attributes on `Event` to match the values. I would honestly rather those be snake case as well.

src/linux_mcp_server/audit.py
samdoran: So the problem with this is the call in the middle. It needs to `await func()` for an asynchronous function. If `awrapper()` was called, the `result` value would be a coroutine and not the result of executing that coroutine.

samdoran: I did some testing at the problem with combining these is that the it's not possible to `await` inside a sync function, so the wrapper function must be `async` (returns a coro). That means it will work for `async` functions but for sync functions, the wrapper will return a coro and the caller will never execute it.

tests/test_audit.py
```diff
@@ -107,76 +109,77 @@ def emit(self, record):
...
-        record = caplog.records[-1]
-        assert hasattr(record, "execution_mode")
-        assert record.execution_mode == "remote"
-        assert hasattr(record, "host")
-        assert record.host == "server1.com"
+        assert caplog.records
+        assert getattr(record, "execution_mode", None) == mode
 
-    def test_log_tool_call_sanitizes_parameters(self, caplog):
+    @pytest.mark.parametrize("secret", SENSITIVE_FIELDS)
```
narmaku, line 150: I initially thought about parameterizing the test case.

But then I though we could pass more than 2 params to the decorated fixture, so we avoid flooding the test results for each single sensitive field and pass all params in one run.

Anyways, here we should change the logic as we are added a field name as the "sensitive" information. But those are just identifiers of what it might be followed by the actual secrete value in the data that needs to be logged.

So, right now the tests will produce:
```
params = {"password": password, "username": "admin"}
params = {"password": passwd, "username": "admin"}
params = {"password": api_key, "username": "admin"}
```
And so on.

But we want to produce:
```
params = {sensitive_field_0: secret_value, "username": "admin"}
params = {sensitive_field_1: secret_value, "username": "admin"}
```
(secret value can be the same for any sensitive field/key/name)
Etc.

I will address this in a new MR as this is the only comment I have after the latest changes.

PR #16, opened 2025-10-31
Fix filename typo in logs.py

estuart:
in logs.py this line has a typo:

`audit_log_path = "/var/log/audit/audilog"`

needs to be:

`audit_log_path = "/var/log/audit/audit.log"`


samdoran:
Thanks! That's a bad copy/replace.

PR #17, opened 2025-10-31
Convert _execute_tool to a decorator

subpop:
I'm trying to iteratively replace the audit and logging machinery while preserving functionality. This PR removes the `AuditContext` context manager and adds a `audit_tool` decorator. It doesn't really *change* anything except the usage pattern. Maybe this can get us one step closer to cleaning out some of the extra layers of abstraction?

### DIFF
--- a/src/linux_mcp_server/audit.py
+++ b/src/linux_mcp_server/audit.py
@@ -59,39 +59,6 @@ def sanitize_parameters(params: dict[str, t.Any]) -> dict[str, t.Any]:
     return sanitized
 
 
-@contextmanager
-def AuditContext(**extra_fields):
-    """
-    Context manager for adding extra fields to all log records.
-
-    Usage:
-        with AuditContext(tool="list_services", host="server1.com") as logger:
-            logger.info("Starting operation")
-
-    Args:
-        **extra_fields: Additional fields to add to log records
-
-    Yields:
-        Logger with extra fields
-    """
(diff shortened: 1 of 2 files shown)


src/linux_mcp_server/audit.py
```diff
@@ -288,3 +257,46 @@ def log_operation(operation: str, message: str, level: int = logging.INFO, **con
...
+                return result
+
+            except Exception as e:
+                duration = time.time() - start_time
+                log_tool_complete(tool_name, status="error", duration=duration, error=str(e))
+                raise
+
+        return wrapper
+
+    return decorator
```
samdoran, line 302: In order to avoid the need to use `()` when using as a decorator and accept parameters to the decorator, add a parameter to the outermost definition (`/` makes items before that positional only), check what was passed to the decorator function to determine what to return.

```python
def audit_tool(function=None, /):
...
```

```suggestion
        return wrapper
    if callable(function):
        return decorator(function)
        
    return decorator
```

samdoran: If the decorator function doesn't need to accept arguments, then you can remove the inner `decorator` definition. But it's helpful in cases where you want to be able to use the decorator both with and without arguments.

samdoran:
I fixed the type test failure but you need to rebase or push another commit in order to run the test with merged with the latest upstream commits.

PR #20, opened 2025-11-06
Update documentation for installation and development setup.

narmaku:
Added quick start instructions, clarified environment activation for different platforms, and streamlined server running commands. Enhanced integration details for Claude Desktop and included references to INSTALL.md for detailed guidance. AI-assisted by Cursor + Claude 4.5 Sonnet.

### DIFF
--- a/INSTALL.md
+++ b/INSTALL.md
@@ -277,8 +277,7 @@ sudo pacman -S python python-pip
 ```
 
 **Notes:**
-- Use `python3` and `pip3` commands instead of `python` and `pip` on most Linux distributions
-- The MCP server requires systemd, so it's designed for modern Linux distributions
+- You may have to use `python3` and `pip3` commands instead of `python` and `pip` on most Linux distributions
 
 ### macOS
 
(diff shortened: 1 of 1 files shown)


INSTALL.md
```diff
@@ -0,0 +1,514 @@
...
+```
+
+**Arch Linux:**
+```bash
+sudo pacman -S python python-pip
+```
+
+**Notes:**
+- Use `python3` and `pip3` commands instead of `python` and `pip` on most Linux distributions
+- The MCP server requires systemd, so it's designed for modern Linux distributions
```
subpop: This line seems out of place to me. The MCP server itself does not require systemd. Its tools assume the presence of systemd on the Linux systems it inspects, but it does not require systemd in order to run.

subpop (commented):
Overall, I think this is an improvement. I have just one comment clarifying something.

INSTALL.md
narmaku: You're right. This is only in the case some tools are called. As this is also commented on possible issues found when running in old/other Linux systems, I will remove this.

PR #22, opened 2025-11-07
Combine list_directories into single tool

subpop:
The new function, list_directories, takes an optional "order_by" and "sort" parameter, allowing ordering name, size and modification time, optionally sorting ascending or descending.

The reduces the number of tools exported by the server by 2.

### DIFF
--- a/tests/tools/test_storage.py
+++ b/tests/tools/test_storage.py
@@ -302,7 +302,7 @@ class TestListDirectories:
     async def test_list_directories_by_modified(self, setup_test_directory):
         """Test that list_directories returns structured output sorted by modification time."""
         dir_specs = [
-            ("newest", 100, 3000.0),
+            ("newest", 0, 3000.0),
             ("oldest", 100, 1000.0),
             ("middle", 100, 2000.0),
         ]


tests/test_audit.py
```diff
@@ -173,13 +173,15 @@ async def test_log_tool_call_async_sanitizes_parameters(self, caplog, adecorated
 
     def test_log_tool_call_failure(self, caplog, decorated_fail):
         with caplog.at_level(logging.INFO):
-            decorated_fail()
+            with pytest.raises(ValueError, match="Raised intentionally"):
+                decorated_fail()
```
samdoran: These can be ground into a single statement.
```
python
with (
    caplog.at_level(logging.INFO),
    pytest.raises(ValueError, match="Raised intentionally"),
):
```

src/linux_mcp_server/utils/clamped_int.py
```diff
@@ -0,0 +1,18 @@
+class clamped_int(int):
```
samdoran: Class names are usually camel case instead of snake case.

src/linux_mcp_server/tools/storage.py
```diff
@@ -83,287 +99,157 @@ async def list_block_devices(
...
+    order_by: t.Annotated[
+        str, Field(description="Sort order - 'size', 'name', or 'modified' (default: 'name')")
+    ] = "name",
+    sort: t.Annotated[
+        str, Field(description="Sort direction - 'ascending' or 'descending' (default: 'ascending')")
+    ] = "ascending",
+    top_n: t.Annotated[
+        int | None,
+        Field(description="Optional limit on number of directories to return (1-1000, only used with size ordering)"),
+    ] = None,
```
samdoran, line 127: Rather than validating this value later using a custom class, use Pydantic's validation machinery.
```suggestion
    top_n: t.Annotated[
        int | None,
        Field(description="Optional limit on number of directories to return (1-1000, only used with size ordering)", gt=0, lt=1_000),
    ] = None,
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -83,287 +98,155 @@ async def list_block_devices(
...
-
-    Returns:
-        Formatted string with directory names, or error message if validation fails
-    """
-    import os
-
+async def list_directories(  # noqa: C901
+    path: t.Annotated[str, Field(description="The directory path to analyze")],
+    order_by: t.Annotated[
+        str, Field(description="Sort order - 'size', 'name', or 'modified' (default: 'name')")
```
samdoran: Using `OrderBy` in the type hint means that validation isn't required inside the function.
```suggestion
        OrderBy, Field(description="Sort order - 'size', 'name', or 'modified' (default: 'name')")
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -83,287 +98,155 @@ async def list_block_devices(
...
+    ] = "name",
+    sort: t.Annotated[
+        str, Field(description="Sort direction - 'ascending' or 'descending' (default: 'ascending')")
```
samdoran: Same thing here.
```suggestion
        SortBy, Field(description="Sort direction - 'ascending' or 'descending' (default: 'ascending')")
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -83,287 +98,155 @@ async def list_block_devices(
...
+                raise ToolError(f"Error running find command: command failed with return code {returncode}")
 
-        # Format output
-        result = []
-        sort_order = "Newest First" if newest_first else "Oldest First"
-        result.append(f"=== Directories ({sort_order}) ===")
-        result.append(f"Path: {path_str}")
-        result.append(f"\nTotal subdirectories found: {len(directories)}\n")
+            # Parse and sort output
+            directories_by_name: list[tuple[None, str]] = [(None, line) for line in stdout.strip().split("\n") if line]
```
samdoran: What's the purpose of creating a list of (None, 'something') pairs? Why not just create a list of directory names?

src/linux_mcp_server/tools/storage.py
```diff
@@ -83,287 +98,155 @@ async def list_block_devices(
...
 
-        return "\n".join(result)
+            # Sort alphabetically
+            reverse = sort == SortBy.DESCENDING
+            directories_by_name.sort(reverse=reverse)
 
-    except Exception as e:
-        return f"Error listing directories: {str(e)}"
+            # Convert to tuples with single element for template compatibility
+            directories_tuples: list[tuple[None, str]] = [(name) for name in directories_by_name]
```
samdoran: The `()` around `name` is a noop. It will be a string, not a single item tuple.

Is this supposed to be a list of single items tuples? I'm still confused about the need for a list of (None, "something") pairs.

src/linux_mcp_server/tools/storage.py
subpop: It keeps the return type more consistent. It can be `Sequence[tuple[int | float | None, str]]` instead of `Sequence[tuple[int | float, str] | str]`. Either way you look at it, it's a mouthful of a type. Would it make more sense to define some explicit data structures that can be populated with the return lists instead of relying on this mashup of types?

samdoran: It sure seems like it would be better to return a typed object. It seems really messy how it is now. The contract is not well defined.

src/linux_mcp_server/tools/storage.py
```diff
@@ -1,32 +1,55 @@
...
 from linux_mcp_server.utils import format_bytes
-from linux_mcp_server.utils.validation import validate_positive_int
+from linux_mcp_server.utils import StrEnum
+
+
+@dataclass
+class DirectoryEntry:
+    size: int = 0
+    modified: float = 0.0
+    name: str = ""
```
samdoran: Since we are using Pydantic in this project, it would be better to use BaseModel.
```suggestion
class DirectoryEntry(BaseModel):
    size: int = 0
    modified: float = 0.0
    name: str = ""
```

src/linux_mcp_server/tools/storage.py
subpop: It looks much better with a typed object as the return type.

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
-    description="List directories under a specified path sorted by size (largest first).",
+    title="List directories",
+    description="List directories under a specified path with flexible sorting options.",
     annotations=ToolAnnotations(readOnlyHint=True),
 )
 @log_tool_call
-async def list_directories_by_size(  # noqa: C901
-    path: t.Annotated[str, "The directory path to analyze"],
+async def list_directories(  # noqa: C901
+    path: t.Annotated[str, Field(description="The directory path to analyze")],
```
samdoran: `path: t.Annotated[str, "description"]` is equivalent to `path: str = Field(description="description")`. If you only need to add a description, it's not necessary to use `Field`.

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
-    List directories under a specified path sorted by modification date.
-
-    This function uses efficient Linux primitives (find) to list directories with timestamps.
-    """
-    import os
-
-    from datetime import datetime
+        order_by = OrderBy(order_by)
+    except ValueError as e:
+        raise ValueError(f"Invalid order_by value '{order_by}'. Must be one of: {', '.join(OrderBy)}") from e
```
samdoran: This should be unnecessary since Pydantic will do this already based on the fact that the type hint is an enum.

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
     try:
-        # For local execution, validate path
-        if not host:
-            try:
-                path_obj = Path(path).resolve(strict=True)
-            except (OSError, RuntimeError):
-                return f"Error: Path does not exist or cannot be resolved: {path}"
+        sort = SortBy(sort)
+    except ValueError as e:
+        raise ValueError(f"Invalid sort value '{sort}'. Must be one of: {', '.join(SortBy)}") from e
```
samdoran: Same thing here: this can be removed.

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
+                username=username,
+            )
+
+            # Parse output - du may return non-zero on permission errors but still give valid data
+            lines = stdout.strip().split("\n")
+            directories_by_size: list[DirectoryEntry] = []
+
+            for line in lines:
+                if not line:
+                    continue
```
samdoran: This can be simplified by removing empty lines beforehand.
```suggestion
            lines = [line.strip() for line in stdout.strip().splitlines() if line]
            directories_by_size: list[DirectoryEntry] = []

            for line in lines:
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
+                parts = line.split("\t", 1)
+                if len(parts) == 2:
+                    try:
```
samdoran: Can the scope of this try/except be narrowed?

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
+                        dir_name = Path(dir_path_str).name
+                        if dir_path_str != path:
+                            directories_by_size.append(DirectoryEntry(size=size, modified=0.0, name=dir_name))
+                    except (ValueError, IndexError):
+                        continue
 
-            if not path_obj.is_dir():
-                return f"Error: Path is not a directory: {path}"
+            if not directories_by_size:
+                # Only error if we got no output AND a bad return code
```
samdoran: This will raise an exception if the list of directories is empty regardless of the return code. Is that the intended behavior?

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
+            for line in stdout.strip().split("\n"):
+                if line:
+                    parts = line.split("\t", 1)
+                    if len(parts) == 2:
+                        try:
+                            timestamp = float(parts[0])
+                            dir_name = parts[1]
+                            directories_by_modified.append(DirectoryEntry(modified=timestamp, name=dir_name))
+                        except ValueError:
+                            continue
```
samdoran: Same as above.
```suggestion
            lines = [line for line in stdout.strip().splitlines() if line]
            for line lines:
                parts = line.split("\t", 1)
                if len(parts) == 2:
                    try:
                        timestamp = float(parts[0])
                        dir_name = parts[1]
                        directories_by_modified.append(DirectoryEntry(modified=timestamp, name=dir_name))
                    except ValueError:
                        continue
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
-) -> str:
-    """
-    List directories under a specified path sorted by name.
-
-    This function uses efficient Linux primitives (find and sort) to list directories.
-    """
-    import os
-
+) -> t.Annotated[
+    Sequence[DirectoryEntry],
```
samdoran: I'm not sure Sequence is correct. The function returns a list which is modified after instantiation. A Sequence, while mutable, is meant to be read-only.

This is probably the reason the type hint `directories_by_size: list[DirectoryEntry]` is necessary. This can be avoided by not modifying the list or changing the return type.

src/linux_mcp_server/tools/storage.py
```diff
@@ -87,273 +109,161 @@ async def list_block_devices(
...
-                        directories.append((timestamp, dir_name))
-                    except ValueError:
-                        continue
+            if returncode != 0:
+                raise ToolError(f"Error running find command: command failed with return code {returncode}")
 
-        if not directories:
-            return f"No subdirectories found in: {path}"
+            # Parse and sort output
+            directories_by_name: list[DirectoryEntry] = [
```
samdoran: This type hint shouldn't be necessary if the return type of the function is changed or this value is not modified after creation.

tests/tools/test_storage.py
```diff
@@ -24,26 +27,116 @@ class TestStorageTools:
 
     async def test_list_block_devices_returns_string(self):
```
samdoran: This test name and doc string probably need updating.

tests/tools/test_storage.py
```diff
@@ -24,26 +27,116 @@ class TestStorageTools:
...
+        assert "result" in result[1]
+        assert isinstance(result[1]["result"], list)
+        assert len(result[1]["result"]) == len(expected_dirs)
+
+        # Verify each item is a dictionary with expected structure
+        for item in result[1]["result"]:
+            assert isinstance(item, dict)
+            assert isinstance(item["name"], str)  # Directory name
+            assert isinstance(item["size"], int)
+            assert isinstance(item["modified"], float)
```
samdoran: This could be replaced by calling the model validator on the return value.

tests/tools/test_storage.py
```diff
@@ -24,26 +27,116 @@ class TestStorageTools:
...
+        # These should either raise ToolError or resolve to safe paths
+        malicious_paths = [
+            "/tmp/../../../etc/passwd",
+            "/tmp; rm -rf /",
+            "/tmp && echo 'malicious'",
+            "/tmp`whoami`",
+            "/tmp$(whoami)",
+        ]
+
+        for path in malicious_paths:
```
samdoran: These should be parameters to the test instead of one test looping over multiple values.

tests/tools/test_storage.py
```diff
@@ -54,145 +147,156 @@ async def test_list_directories_by_size_with_temp_dirs(self, tmp_path):
...
+        assert isinstance(result[1], dict)
+        assert "result" in result[1]
+        assert isinstance(result[1]["result"], list)
+        assert len(result[1]["result"]) == 3
+
+        # Each result should be a dictionary (serialized DirectoryEntry)
+        for item in result[1]["result"]:
+            assert isinstance(item, dict)
+            assert isinstance(item["size"], int)  # Size in bytes
+            assert isinstance(item["name"], str)  # Directory name
```
samdoran: Same thing here: use the model validation tooling.

tests/tools/test_storage.py
```diff
@@ -54,145 +147,156 @@ async def test_list_directories_by_size_with_temp_dirs(self, tmp_path):
...
+            "list_directories", {"path": str(tmp_path), "order_by": "size", "sort": "ascending"}
+        )
+
+        # Verify the entire result structure
+        assert isinstance(result, tuple)
+        assert len(result) == 2
+        assert isinstance(result[1], dict)
+        assert "result" in result[1]
+        assert isinstance(result[1]["result"], list)
+        assert len(result[1]["result"]) == 2
```
samdoran: I don't know if we need to do this in every test. Maybe have one test that validates the output format and allow other tests to focus on the contents of the data.

tests/tools/test_storage.py
```diff
@@ -1,351 +1,712 @@
...
+
+        # Verify lsblk was called with correct arguments
+        mock_execute_command.assert_called_once()
+        args = mock_execute_command.call_args[0][0]
+        assert args[0] == "lsblk"
+        assert "-o" in args
+
+    @patch("linux_mcp_server.tools.storage.psutil.disk_io_counters")
+    @patch("linux_mcp_server.tools.storage.execute_command")
+    async def test_list_block_devices_with_disk_io_stats(self, mock_execute_command, mock_disk_io):
```
samdoran, line 102: It's not a blocker, but it would be better to use the `pytest-mock` fixture instead of the `patch` decorator.

tests/tools/test_storage.py
```diff
@@ -1,351 +1,712 @@
...
+        got = [DirectoryEntry(**entry) for entry in result[1]["result"]]
+
+        # Verify results are sorted by modified time
+        assert len(got) == 3
+        assert got[0].name == "oldest"
+        assert got[0].modified == 1000.0
+        assert got[1].name == "middle"
+        assert got[1].modified == 2000.0
+        assert got[2].name == "newest"
+        assert got[2].modified == 3000.0
```
samdoran, line 649: A simpler way to write this would be `assert got == result[1]["result"]` since the comparison operator will compare the version of each item in the sequence. You could either change `got` to a list of dicts by using `.model_dump()` on the objects or convert `result[1]["results"]` to a list of objects when doing the comparison.

samdoran (approved):

samdoran:
There's a partial miss on the fixture in `test_storage.py` because we never create a directory of size 0. We need a test case that does that or remove the conditional. I'm still looking at the other missing coverage.

PR #26, opened 2025-11-10
Correct enum value for tool call complete

samdoran:
Use the correct value for `Event.TOOL_COMPLETE` and add another enum for status.


src/linux_mcp_server/audit.py
```diff
@@ -43,14 +43,19 @@ class Event(StrEnum):
     SSH_CONNECT = "SSH_CONNECT"
     SSH_CONNECTING = "SSH_CONNECTING"
     TOOL_CALL = "TOOL_CALL"
-    TOOL_COMPLETE = "TOOL_CALL"
+    TOOL_COMPLETE = "TOOL_COMPLETE"
```
narmaku: Oh. This is a sign of tired reviewers :laughing:

subpop (approved):
LGTM

PR #28, opened 2025-11-12
Move SSH connection manager code outside of tools

samdoran:
This makes code inside `tools/` only the tool definitions and moves connection related code to its own package. I also updated the test structure to match the source code and added explicit tests for code that was only covered incidentally.

### DIFF
--- a/tests/utils/test_enum.py
+++ b/tests/utils/test_enum.py
@@ -1,0 +1,24 @@
+import pytest
+
+from linux_mcp_server.utils.enum import StrEnum
+
+
+@pytest.fixture
+def color():
+    class Color(StrEnum):
+        red = "red"
+        blue = "blue"
+
+    return Color
+
+
+def test_enum(color):
+    assert color.red == "red"
+    assert color.blue == "blue"
+    assert color.red in "Little red riding hood"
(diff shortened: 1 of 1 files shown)


tests/utils/test_enum.py
```diff
@@ -0,0 +1,24 @@
...
+        blue = "blue"
+
+    return Color
+
+
+def test_enum(color):
+    assert color.red == "red"
+    assert color.blue == "blue"
+    assert color.red in "Little red riding hood"
+    assert color.blue in "I'm blue, da-ba-dee, da-ba-di"
```
narmaku, line 19: Such a great song!

tests/utils/test_format.py
```diff
@@ -0,0 +1,22 @@
...
+    ("value", "expected"),
+    (
+        (512, "512.0B"),
+        (1024, "1.0KB"),
+        (2048, "2.0KB"),
+        (1536, "1.5KB"),
+        (1_073_741_824, "1.0GB"),
+        (201_073_741_824, "187.3GB"),
+        (5_201_073_741_824, "4.7TB"),
+        (10_105_201_073_741_824, "9.0PB"),
```
narmaku, line 16: :exploding_head:

PR #36, opened 2025-11-18
Convert system_info tools to structured output

subpop:
Convert the `system_info.py` tools to use structured output for responses. I also performed a fairly significant audit of the function bodies and refactored them. Of note, I removed the use of `psutil` for local data collection, so both local and remote use the same commands to collect data.

I also refactored a lot of the parsing to use a pattern of retrieving the entire remote file contents and parsing the data internally (rather than running grep as part of the remote command).

### DIFF
--- a/src/linux_mcp_server/tools/system_info.py
+++ b/src/linux_mcp_server/tools/system_info.py
@@ -408,6 +408,7 @@ async def _parse_memory_information(raw_outputs: dict[CommandKey, RawCommandOutp
     annotations=ToolAnnotations(readOnlyHint=True),
 )
 @log_tool_call
+@disallow_local_execution_in_containers
 async def get_memory_information(
     host: Host | None = None,
     fields: t.Annotated[
@@ -688,6 +689,7 @@ async def _parse_device_information(raw_outputs: dict[CommandKey, RawCommandOutp
     annotations=ToolAnnotations(readOnlyHint=True),
 )
 @log_tool_call
+@disallow_local_execution_in_containers
 async def get_device_information(
     host: Host | None = None,
     device_types: t.Annotated[


src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,144 +1,201 @@
...
-
-            # Uptime
+    result = SystemInfo()
+
+    class SystemInfoAttribute(BaseModel):
+        """System information attribute."""
+
+        names: list[str]
+        command: list[str]
+        parser: Callable[[str], t.Any]
```
r0x0d: Why not moving this to outside the function like the rest?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -147,127 +204,104 @@ async def get_system_information(  # noqa: C901
...
-            except Exception:
-                pass
-
-        return "\n".join(info)
-    except Exception as e:
-        return f"Error gathering CPU information: {str(e)}"
+        except ValueError or ConnectionError as e:
+            raise ToolError(f"Error: {str(e)}") from e
+
+        if returncode == 0 and stdout:
```
r0x0d: What should be the desired output to the user in case returncode == 0 or stdout is empty?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,144 +1,201 @@
...
-            info.append(f"Boot Time: {boot_time.strftime('%Y-%m-%d %H:%M:%S')}")
-
-        return "\n".join(info)
-    except Exception as e:
-        return f"Error gathering system information: {str(e)}"
+        except ValueError or ConnectionError as e:
+            raise ToolError(f"Error: {str(e)}") from e
+
+        if returncode == 0 and stdout:
+            parsed_output = attribute.parser(stdout)
```
r0x0d: I'm wondering if it would be wise to have some error checking around parser. It might not be the case that all of them contain incompatible data for the parsers to break, but I think we would be more on the safe side.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -279,79 +313,230 @@ async def get_cpu_information(  # noqa: C901
...
-    info = []
+    ram_stats: MemoryStats | None = None
+    swap_stats: MemoryStats | None = None
+
+    class MemoryStatsAttribute(BaseModel):
+        """Memory statistics attribute."""
+
+        names: list[str]
+        command: list[str]
+        parser: Callable[[str], t.Any]
```
r0x0d: Since the attributes of such classes are all the same, wouldn't be better to just have one `SystemInfoAttributes` or similar and re-use that? If they had any particular attributes for each of the tool calling, I guess we could just inherit from the base `SystemInfoAttributes` model instead.

r0x0d: Now that I think about it, would it also make sense to have `execute_command` inside the model? For instance: 

```python
class SystemInfoAttributes(BaseModel):
    names: list[str]
    command: list[str]
    parser: Callable[[str], t.Any]

    async def execute(self, host: str) -> str:
         try:
            returncode, stdout, _ = await execute_command(
                self.command,
                host=host
            )
        except ValueError or ConnectionError as e:
            raise ToolError(f"Error: {str(e)}") from e

        if not returncode or not stdout:
            # Throw error in case returncode/stdout is not what expected
            ...

        return stdout
```

Then, in each of the classes, for-loop, you could have: 

```python
    for attribute in attributes:
        stdout = await attribute.execute()
        attribute.parser(stdout)

        match attribute.names:
               ...
    return result
```

I'm not sure if I'm sold out on the idea of checking the returncode/stdout inside the `execute()` function, but I guess my idea is around those lines.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,144 +1,201 @@
...
+                    result.os_version = parsed_output["VERSION_ID"]
+                case ["kernel_version", "architecture"]:
+                    result.kernel_version = parsed_output["kernel_version"]
+                    result.architecture = parsed_output["architecture"]
+                case ["uptime"]:
+                    result.uptime = parsed_output
+                case ["boot_time"]:
+                    result.boot_time = parsed_output
+                case _:
+                    raise ToolError(f"Unknown attribute: {attribute.names}")
```
r0x0d: Would a dict unpacking work here instead? 

```python
try:
    system_info = SystemInfo(**attribute.names)
except ValueError as e:
    raise ToolError(f"Unknown attribute: {attribute.names}") from e

return system_info
```

r0x0d: Ah! It would not, because we are doing it in a loop and different tools return different results... Okay, nevermind.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -279,79 +313,230 @@ async def get_cpu_information(  # noqa: C901
...
+
+    class MemoryStatsAttribute(BaseModel):
+        """Memory statistics attribute."""
+
+        names: list[str]
+        command: list[str]
+        parser: Callable[[str], t.Any]
+
+    def parse_free_output(output: str) -> dict[str, MemoryStats]:
+        """Parse the output of 'free -b' command into memory statistics."""
```
r0x0d: Do you think we can document here the types of memory we are parsing?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -279,79 +313,230 @@ async def get_cpu_information(  # noqa: C901
...
+    swap_stats: MemoryStats | None = None
+
+    class MemoryStatsAttribute(BaseModel):
+        """Memory statistics attribute."""
+
+        names: list[str]
+        command: list[str]
+        parser: Callable[[str], t.Any]
+
+    def parse_free_output(output: str) -> dict[str, MemoryStats]:
```
r0x0d: The local functions are always confusing when reading/reviewing code. Is there any possibility we move them outside this function and turn it private?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -279,79 +313,230 @@ async def get_cpu_information(  # noqa: C901
...
+                if len(parts) >= 7:
+                    total = int(parts[1])
+                    used = int(parts[2])
+                    free = int(parts[3])
+                    # parts[4] is shared memory (not used in MemoryStats)
+                    buff_cache = int(parts[5]) if len(parts) > 5 else 0
+                    available = int(parts[6]) if len(parts) > 6 else free
+                    percent = (used / total * 100) if total > 0 else 0
+
+                    result["ram"] = MemoryStats(
```
r0x0d: I'm wondering, why not consider `swap` as an attribute to memory?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -279,79 +313,230 @@ async def get_cpu_information(  # noqa: C901
...
+                    ram_stats = parsed_output["ram"]
+                    swap_stats = parsed_output["swap"]
+                case _:
+                    raise ToolError(f"Unknown attribute: {attribute.names}")
+
+    return MemoryInfo(ram=ram_stats, swap=swap_stats)
+
+
+def parse_df_output(output: str) -> list[DiskPartition]:
+    """Parse the output of 'df -B1' command into disk partitions."""
```
r0x0d: Same here, can we document the different parts we parse?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,273 +1,453 @@
...
+
+    # Parse phase: Convert raw outputs to structured data
+    parsed_data = parse_cpu_information(raw_outputs)
+
+    # Filter phase: Apply field filters and build result
+    filtered_data = filter_fields(parsed_data, fields)
+
+    # Build CPUInfo object from filtered data
+    load_avg_data = filtered_data.get("load_average")
+    load_average = LoadAverage(**load_avg_data) if load_avg_data else None
```
r0x0d: Instead of assigning `None` in case of not having data for `load_avg_data`, I think we should still assign a `LoadAverage` instance, but with default values. 

This ensures that we maintain a structured output, regardless if we have data to it or not, avoiding a case where a later read to CPUInfo.load_average might be None. 

```suggestion
    load_average = LoadAverage(**load_avg_data) if load_avg_data else LoadAverage()
```

subpop:
I made another significant refactor, this time pulling the collection and filtering logic out of `storage.py` and into its own module. I created a class (named `Rummager` for lack of a better name) that can be initialized with custom collect, parse and filter functions. When `Rummager.rummage` is called, it runs through those 3 functions to collect and process the data. If no custom function is provided at initialization, and default one is used.

src/linux_mcp_server/tools/system_info.py
subpop: Swap stats and Mem stats are identical lines of data in the `free -b` output.

```
               total        used        free      shared  buff/cache   available
Mem:      6183096320   378454016  4469956608      712704  1488175104  5804642304
Swap:     6182400000           0  6182400000
```

That's probably why the original LLM that wrote the code treated them as separate items in the memory info data. `MemoryStats` has a lot of fields. Do we need all of them? Maybe we could merge "total_swap" in as a value of the regular `MemoryStat` object?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,138 +1,282 @@
...
+    """Device information model for PCI and USB devices."""
+
+    pci_devices: list[PCIDevice] = Field(default_factory=list)
+    pci_device_count: int = 0
+    usb_devices: list[USBDevice] = Field(default_factory=list)
+    usb_device_count: int = 0
+
+
+async def parse_system_information(raw_outputs: dict[CommandKey, RawCommandOutput]) -> ParsedData:
+    """
```
r0x0d, line 162: I think it's valuable to include what type of system information is expected this method to parse. I know that one can simply look at the match-case statements, but, having it also documented in the docstring helps when hovering over the method in text editors.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -142,120 +286,90 @@ async def get_system_information(  # noqa: C901
...
+        lines = raw_outputs[("free", "-b")].strip().split("\n")
+
+        for line in lines:
+            if line.startswith("Mem:"):
+                parts = line.split()
+                if len(parts) >= 7:
+                    total = int(parts[1])
+                    used = int(parts[2])
+                    free = int(parts[3])
+                    # parts[4] is shared memory (not used in MemoryStats)
```
r0x0d: why not tracking the shared memory? it's not relevant to the info?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -264,81 +378,119 @@ async def get_cpu_information(  # noqa: C901
...
-
-        return "\n".join(info)
-    except Exception as e:
-        return f"Error gathering memory information: {str(e)}"
+    # Build MemoryInfo object from filtered data
+    ram_data = filtered_data.get("ram")
+    swap_data = filtered_data.get("swap")
+
+    ram_stats = MemoryStats(**ram_data) if ram_data else None
+    swap_stats = MemoryStats(**swap_data) if swap_data else None
```
r0x0d:
```suggestion
    ram_stats = MemoryStats(**ram_data) if ram_data else MemoryStats()
    swap_stats = MemoryStats(**swap_data) if swap_data else MemoryStats()
```

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +502,266 @@ async def get_memory_information(
...
-    description="Get hardware information such as CPU details, PCI devices, USB devices, and hardware information from DMI.",
-    annotations=ToolAnnotations(readOnlyHint=True),
-)
-@log_tool_call
-@disallow_local_execution_in_containers
-async def get_hardware_information(  # noqa: C901
-    host: Host | None = None,
-) -> str:
+    io_stats_data = filtered_data.get("io_stats")
+    io_stats = DiskIOStats(**io_stats_data) if io_stats_data else None
```
r0x0d:
```suggestion
    io_stats = DiskIOStats(**io_stats_data) if io_stats_data else DiskIOStats()
```

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +502,266 @@ async def get_memory_information(
...
+            path.strip() for path in stdout.strip().split("\n") if path.strip() and path != "/sys/bus/pci/devices"
+        ]
+
+        for device_path in device_paths:
+            address = device_path.split("/")[-1]
+            cmd = (
+                "sh",
+                "-c",
+                f"cd {device_path} && for f in vendor device class label; do echo \"$f:$(cat $f 2>/dev/null || echo '')\"; done",
             )
```
r0x0d: I don't know… This doesn't feel safe at all, and it's hard to understand. Can we change this to use python functions instead? 

```python
for descriptor in ("vendor", "device", "class", "label"):
      with open(device_path, mode="r") as fd:
           fd.read() 
           ...
```

Makes it easier to understand what the code is doing, and we have the possibility to parse and give an output in case of one of those descriptors is not present for the device.

r0x0d: I'm not sure how easy it is to get a malicious device on the system that could alter the behavior of this, but I would feel safer if we used raw python in this part rather than subprocessing from execute_command like this

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +502,266 @@ async def get_memory_information(
...
+            device_name = device_path.split("/")[-1]
+            # Skip root hubs (like usb1, usb2)
+            if "-" not in device_name:
+                continue
+
+            cmd = (
+                "sh",
+                "-c",
+                f"cd {device_path} && for f in idVendor idProduct product manufacturer; do echo \"$f:$(cat $f 2>/dev/null || echo '')\"; done",
             )
```
r0x0d: Same as above https://github.com/rhel-lightspeed/linux-mcp-server/pull/36/files#r2580803194

src/linux_mcp_server/processing.py
```diff
@@ -0,0 +1,106 @@
...
+CommandKey = tuple[str, ...]
+RawCommandOutput = str
+ParsedData = dict[str, t.Any]
+CommandList = list[list[str]]
+CollectFunc = t.Callable[[CommandList, Host | None], t.Awaitable[dict[CommandKey, RawCommandOutput]]]
+ParseFunc = t.Callable[[dict[CommandKey, RawCommandOutput]], t.Awaitable[ParsedData]]
+FilterFunc = t.Callable[[ParsedData, list[str] | None], t.Awaitable[ParsedData]]
+
+
+class Rummager:
```
r0x0d: Can you document what this means and the purpose of it? `Rummager` does not mean much to me right away. 

(tried to translate, but apparently, no direct translations to my language too)

src/linux_mcp_server/processing.py
```diff
@@ -0,0 +1,106 @@
...
+        collect_func: CollectFunc | None = None,
+        parse_func: ParseFunc | None = None,
+        filter_func: FilterFunc | None = None,
+    ):
+        if collect_func is not None:
+            self._collect = collect_func  # type: ignore[method-assign]
+        if parse_func is not None:
+            self._parse = parse_func  # type: ignore[method-assign]
+        if filter_func is not None:
+            self._filter = filter_func  # type: ignore[method-assign]
```
r0x0d: I get the idea of having the default functions, and I pretty much like it, however, I feel like overriding the base methods will cause some confusion and I feel we could avoid ignoring the type error here. Maybe something like: 

```python
def __init__(self, collect_func, ...): 
     self._collect = collect_func or _default_collect
     ....
```

r0x0d: I think that will make the code easier to read and less confusing, as we can see directly that either we assign our own function to it, or we assign a default one

src/linux_mcp_server/processing.py
```diff
@@ -0,0 +1,106 @@
...
+                    if parent not in filtered:
+                        filtered[parent] = {}
+                    if isinstance(parsed_data[parent], dict) and child in parsed_data[parent]:
+                        if not isinstance(filtered[parent], dict):
+                            filtered[parent] = {}
+                        filtered[parent][child] = parsed_data[parent][child]
+            elif field in parsed_data:
+                filtered[field] = parsed_data[field]
+
+        return filtered
```
r0x0d: Based on my comment above, and if you agree with it (https://github.com/rhel-lightspeed/linux-mcp-server/pull/36/files#r2580863868), you can move all of those to outside the class as they are not using anything from the class itself. 

Usually. I like to keep such methods as internal methods of my module and not to my class since they are not depending on anything else of the class itself.

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+                case ["uptime", "-s"]:
+                    return (0, "2024-11-23 10:15:30\n", "")
+                case _:
+                    return (1, "", "Command not found")
+
+        # Patch execute_command in processing module (used by default Rummager._collect)
+        with patch(
+            "linux_mcp_server.processing.execute_command",
+            new=AsyncMock(side_effect=mock_execute_command),
+        ):
```
r0x0d: use mocker instead

```suggestion
        mocker.patch("linux_mcp_server.processing.execute_command", AsyncMock(side_effect=mock_execute_command))
```

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora/f39/system-administrators-guide/"
+SUPPORT_URL="https://ask.fedoraproject.org/"
+BUG_REPORT_URL="https://bugzilla.redhat.com/"
+REDHAT_BUGZILLA_PRODUCT="Fedora"
+REDHAT_BUGZILLA_PRODUCT_VERSION=39
+REDHAT_SUPPORT_PRODUCT="Fedora"
+REDHAT_SUPPORT_PRODUCT_VERSION=39
+SUPPORT_END=2024-11-12
+VARIANT="Server Edition"
+VARIANT_ID=server"""
```
r0x0d, line 58: For ostree systems, we have a couple of extra fields, like: 

```bash
r0x0d@nightmare ~> cat /etc/os-release 
NAME="Fedora Linux"
VERSION="43.20251201.0 (Silverblue)"
RELEASE_TYPE=stable
ID=fedora
VERSION_ID=43
VERSION_CODENAME=""
PRETTY_NAME="Fedora Linux 43.20251201.0 (Silverblue)"
ANSI_COLOR="0;38;2;60;110;180"
LOGO=fedora-logo-icon
CPE_NAME="cpe:/o:fedoraproject:fedora:43"
DEFAULT_HOSTNAME="fedora"
HOME_URL="https://silverblue.fedoraproject.org"
DOCUMENTATION_URL="https://docs.fedoraproject.org/en-US/fedora-silverblue/"
SUPPORT_URL="https://ask.fedoraproject.org/"
BUG_REPORT_URL="https://github.com/fedora-silverblue/issue-tracker/issues"
REDHAT_BUGZILLA_PRODUCT="Fedora"
REDHAT_BUGZILLA_PRODUCT_VERSION=43
REDHAT_SUPPORT_PRODUCT="Fedora"
REDHAT_SUPPORT_PRODUCT_VERSION=43
SUPPORT_END=2026-12-02
VARIANT="Silverblue"
VARIANT_ID=silverblue
OSTREE_VERSION='43.20251201.0'
```

I'm wondering if it is worth it to also catch and display those

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+                return (0, f"{load_avg_values[0]} {load_avg_values[1]} {load_avg_values[2]} 1/100 12345", "")
+
+            elif command == ["top", "-bn1"]:
+                # Mock top output
+                return (0, "Cpu(s): 10.5%us, 5.2%sy, 0.0%ni, 84.3%id, 0.0%wa", "")
+
+            return (1, "", "Command not found")
+
+        # Patch execute_command in processing module (used by default Rummager._collect)
+        with patch("linux_mcp_server.processing.execute_command", new=AsyncMock(side_effect=mock_execute_command)):
```
r0x0d:
```suggestion
        mocker.patch("linux_mcp_server.processing.execute_command", AsyncMock(side_effect=mock_execute_command))
```

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+                "free": 6182400000,
+                "percent": 0.0,
+            },
+        }
+
+        # Patch execute_command in processing module (used by default Rummager._collect)
+        with patch(
+            "linux_mcp_server.processing.execute_command",
+            new=AsyncMock(side_effect=mock_execute_command),
+        ):
```
r0x0d: mocker here too

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+            assert isinstance(result, tuple)
+            assert len(result) == 2
+            content, structured_output = result
+
+            # Content should be a TextContent object array
+            assert isinstance(content, list)
+            assert all(isinstance(item, TextContent) for item in content)
+
+            # Extract text from content and verify exact values
+            assert isinstance(content[0], TextContent)
```
r0x0d: Not needed as the above one is already verifying that all content is actually `TextContent`


```suggestion

```

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+                assert cpu_info.physical_cores_per_processor is not None
+                assert cpu_info.physical_cores_per_processor == expected_physical_cores_per_processor
+
+            # Frequency assertion
+            if expected_frequency > 0:
+                assert cpu_info.frequency_mhz is not None
+                assert cpu_info.frequency_mhz == expected_frequency
+            else:
+                # ARM doesn't have frequency, so it will be 0.0
+                assert cpu_info.frequency_mhz == 0.0
```
r0x0d:
```suggestion
            assert expected_model in cpu_info.cpu_model
            assert cpu_info.logical_cores == expected_logical_cores
            assert cpu_info.load_average.one_min == load_avg_values[0]
            assert cpu_info.load_average.five_min == load_avg_values[1]
            assert cpu_info.load_average.fifteen_min == load_avg_values[2]

            # Physical cores assertion
            if expected_physical_sockets > 0:
                assert cpu_info.physical_sockets == expected_physical_sockets

            # Physical cores per processor assertion
            if expected_physical_cores_per_processor > 0:
                assert cpu_info.physical_cores_per_processor == expected_physical_cores_per_processor

            # Frequency assertion
            if expected_frequency > 0:
                assert cpu_info.frequency_mhz == expected_frequency
            else:
                # ARM doesn't have frequency, so it will be 0.0
                assert cpu_info.frequency_mhz == 0.0
```

I would remove the `None` assert because if it is None, we will notice in the next assert anyway

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+            assert isinstance(result, tuple)
+            assert len(result) == 2
+            content, structured_output = result
+
+            # Content should be a TextContent object array
+            assert isinstance(content, list)
+            assert all(isinstance(item, TextContent) for item in content)
+
+            # Extract text from content and verify JSON contains exact values
+            assert isinstance(content[0], TextContent)
```
r0x0d:
```suggestion
```

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+            assert ram["percent"] == pytest.approx(expected_memory_stats_json["ram"]["percent"], rel=1e-2)
+
+            # Verify swap information in JSON
+            assert "swap" in data
+            swap = data["swap"]
+            assert swap is not None
+            assert swap["total"] == expected_memory_stats_json["swap"]["total"]
+            assert swap["used"] == expected_memory_stats_json["swap"]["used"]
+            assert swap["free"] == expected_memory_stats_json["swap"]["free"]
+            assert swap["percent"] == expected_memory_stats_json["swap"]["percent"]
```
r0x0d:
```suggestion
            assert "ram" in data
            ram = data["ram"]
            assert ram["total"] == expected_memory_stats_json["ram"]["total"]
            assert ram["used"] == expected_memory_stats_json["ram"]["used"]
            assert ram["free"] == expected_memory_stats_json["ram"]["free"]
            assert ram["available"] == expected_memory_stats_json["ram"]["available"]
            assert ram["buffers"] == expected_memory_stats_json["ram"]["buffers"]
            assert ram["percent"] == pytest.approx(expected_memory_stats_json["ram"]["percent"], rel=1e-2)

            # Verify swap information in JSON
            assert "swap" in data
            swap = data["swap"]
            assert swap["total"] == expected_memory_stats_json["swap"]["total"]
            assert swap["used"] == expected_memory_stats_json["swap"]["used"]
            assert swap["free"] == expected_memory_stats_json["swap"]["free"]
            assert swap["percent"] == expected_memory_stats_json["swap"]["percent"]
```

Same here. If "ram" or "swap" is not present in `data`, the data["ram"/"swap"] key access would fail

r0x0d: also, better to have a failure like `KeyError` than trying to verify if the value is None first here

tests/tools/test_system_info.py
```diff
@@ -1,72 +1,604 @@
...
+            assert first_partition_model.percent == expected_partitions[0]["percent"]
+
+            # Verify I/O stats in structured output
+            assert disk_usage.io_stats is not None
+            assert disk_usage.io_stats.read_bytes == expected_io_stats["read_bytes"]
+            assert disk_usage.io_stats.write_bytes == expected_io_stats["write_bytes"]
+            assert disk_usage.io_stats.read_count == expected_io_stats["read_count"]
+            assert disk_usage.io_stats.write_count == expected_io_stats["write_count"]
+
+    async def test_get_device_info_contains_device_data(self):
```
r0x0d: What about breaking this test in two? One that looks for `pci` and one for `usb`, as it will be easier to spot out problems in separate tests

src/linux_mcp_server/tools/system_info.py
subpop: It creates a single cache entry for the files in /sys/bus/pci/devices/$ADDRESS/{vendor, device, class, label} rather than a single cache entry for each of those files. It creates a file that looks like:

```
vendor:My Vendor, Inc.
device:1234
class:trackpad
label:My Vendor Trackpad
```

And it does this in a single call to `execute_command`. I know what you mean about using more Python to perform the collection, but we'd probably need to call `execute_function` more frequently in order to cat each file individually.

r0x0d: I get the convenience about the shell command, but I'm not super sure if that is worth it because of the complexity of it and if there is any security considerations that we need to keep in mind. I know that calling subprocesses like this may be dangerous

subpop: I rewrote it in more Python. It looks much easier to read. We'll see how it runs.

tests/tools/test_system_info.py
subpop: If we're going to do something like this, I think we should just show every value in `/etc/os-release`. That would make the code to process it much more straightforward. I'm okay the way it is, if all we care about for "system information" is the `PRETTY_NAME` and `VERSION_ID`.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
 )
 @log_tool_call
-@disallow_local_execution_in_containers
-async def get_hardware_information(  # noqa: C901
+async def get_device_information(
     host: Host | None = None,
-) -> str:
+    device_types: t.Annotated[
+        list[str],
+        Field(
```
samdoran, line 697: If this parameter only accepts a known number of values, the type hint should be an enum of those values so that the input constraint can be enforced.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,138 +1,290 @@
...
+    manufacturer: str | None = None
+
+
+class DeviceInfo(BaseModel):
+    """Device information model for PCI and USB devices."""
+
+    pci_devices: list[PCIDevice] = Field(default_factory=list)
+    pci_device_count: int = 0
+    usb_devices: list[USBDevice] = Field(default_factory=list)
+    usb_device_count: int = 0
```
samdoran: The count parameters should be properties, not parameters.
```suggestion
    pci_devices: list[PCIDevice] = Field(default_factory=list)
    usb_devices: list[USBDevice] = Field(default_factory=list)
    
    @property
    def pci_device_count(self) -> int:
        return len(self.pci_devices)
        
    @property
    def usb_device_count(self) -> int:
        return len(self.usb_devices)
```

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+    ]
+    | None = None,
+) -> DeviceInfo:
     """
-    Get hardware information.
+    Get device information for PCI and USB devices.
+
+    This tool reads device information directly from the Linux sysfs filesystem
+    (/sys/bus/pci/devices/ and /sys/bus/usb/devices/) and does not rely on
+    external utilities like lspci or lsusb which may not be installed.
```
samdoran: This is not a blocker, but a better strategy to take is using those tools if available and falling back to raw processing as a last resort. Those tools have had decades of debugging and refinement.

src/linux_mcp_server/data_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+                    if not isinstance(filtered[parent], dict):
+                        filtered[parent] = {}
+                    filtered[parent][child] = parsed_data[parent][child]
+        elif field in parsed_data:
+            filtered[field] = parsed_data[field]
+
+    return filtered
+
+
+class DataPipeline:
```
samdoran: This isn't really a pipeline, it's more a data parser.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
-            info.append("No hardware information tools available.")
-
-        return "\n".join(info)
-    except Exception as e:
-        return f"Error getting hardware information: {str(e)}"
+
+    # Create a custom filter function that wraps filter_device_information
+    # to match the FilterFunc signature
+    async def custom_filter(parsed_data: ParsedData, fields: list[str] | None) -> ParsedData:
+        return await _filter_device_information(parsed_data, device_types=device_types, limit=limit)
```
samdoran: That this local function needs to exists seems wrong. This seems like `_filter_device_information` is not adhering to the interface contract for filter functions and needs to be changed instead of adding this translation layer.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+
+    # Use Rummager with custom collect, parse, and filter functions
```
samdoran: This comment needs updating to match the class name, or it can be removed.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+    rummager = DataPipeline(
+        collect_func=_collect_device_information,
+        parse_func=_parse_device_information,
+        filter_func=custom_filter,
+    )
+
+    # Commands list is empty since collection is dynamic
+    filtered_data = await rummager.process(commands=[], host=host)
```
samdoran: Passing an empty list seems a bit odd. If the caller doesn't have any commands to run, the API should not require commands.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+    Parse phase: Convert raw device outputs into structured device data.
+
+    Returns a dictionary with pci_devices and usb_devices lists.
+    """
+    pci_devices = []
+    usb_devices = []
+
+    # Parse PCI and USB devices
+    for command_key, output in raw_outputs.items():
+        if command_key[0] == "pci_attrs" and len(command_key) == 2:
```
samdoran: `CommandKey` could be a typed object with attributes. This would make it easier to interact with.
```python
class CommandKey(BaseModel):
    command: str
    args: list[str]

    @staticmethod
    def create(command: t.Sequence[str, ...]):
        command = shutil.which(command[0])
        args = command[1:]
        return CommandKey(command=command, args=args)
```

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+            address = command_key[1]
+            pci_devices.append(_parse_pci_device_attrs(address, output))
+        elif command_key[0] == "usb_attrs" and len(command_key) == 2:
+            device_name = command_key[1]
+            usb_devices.append(_parse_usb_device_attrs(device_name, output))
+
+    return {
+        "pci_devices": pci_devices,
+        "usb_devices": usb_devices,
+    }
```
samdoran, line 683: It's a bit tedious, but these returns should be objects if possible.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+
+
+async def _filter_device_information(
+    parsed_data: ParsedData, device_types: list[str] | None = None, limit: int | None = None
+) -> ParsedData:
```
samdoran: For all these filter functions, we could use the [validation decorator](https://docs.pydantic.dev/latest/concepts/validation_decorator/#using-the-field-function-to-describe-function-parameters) to more strictly enforce the function inputs and outputs. This would allow us to use mutable defaults, something that's not possible to do safely by default.

```suggestion
DeviceTypes(StrEnum):
    pci = "pci"
    usb = "usb"
    
@validate_call
async def _filter_device_information(
    parsed_data: ParsedData,
    device_types: list[DeviceTypes] | None = None, 
    limit: t.Annotated[int | None, Field(lt=1_0000)] = None,
) -> ParsedData:
```

src/linux_mcp_server/tools/system_info.py
```diff
@@ -350,156 +511,272 @@ async def get_memory_information(
...
+            pci_devices = pci_devices[:limit]
+        filtered["pci_devices"] = pci_devices
+
+    if should_include_usb and "usb_devices" in parsed_data:
+        usb_devices = parsed_data["usb_devices"]
+        if limit is not None:
+            usb_devices = usb_devices[:limit]
+        filtered["usb_devices"] = usb_devices
+
+    return filtered
```
samdoran: Again, it would be better to return an object unless that's really not feasible.

src/linux_mcp_server/data_pipeline.py
samdoran, line 1: `processing.py` or `data_processing.py` would probably be a better name. Or this could be a `data` package with separate modules for filter functions and the main processing class.

src/linux_mcp_server/data_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+import typing as t
+
+from mcp.server.fastmcp.exceptions import ToolError
+
+from linux_mcp_server.connection.ssh import execute_command
+from linux_mcp_server.utils.types import Host
+
+
+# Type aliases for the collect-parse-filter pattern
+CommandKey = tuple[str, ...]
```
samdoran: Rather than a tuple, this could be a hashable object.

src/linux_mcp_server/data_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+RawCommandOutput = str
+ParsedData = dict[str, t.Any]
```
samdoran, line 35: How tedious would it be to create typed objects with consistent properties from each filter?

src/linux_mcp_server/data_pipeline.py
```diff
@@ -0,0 +1,106 @@
...
+CommandList = list[list[str]]
```
samdoran, line 36: Is this a list that contains a list of strings or just a list of strings?

subpop: It's a list of a list of strings. So that multiple command strings can be collected. For example:

```python
[
    ["cat", "/proc/cpuinfo"],
    ["cat", "/etc/os-release"]
]
```

src/linux_mcp_server/data_pipeline.py
subpop: Considering each parse function is very specific, probably not too tedious. But it does mean those types would have to live closer to the parse functions that declare them, and we'd have to have some other type here for all those custom parser types to conform to.

src/linux_mcp_server/data_pipeline.py
subpop: This is why I had a hard time with the name. I'm fine calling it `DataParser` too.

src/linux_mcp_server/tools/system_info.py
subpop: Almost everything about collecting data for PCI and USB devices does not fit the collect-parse-filter interfaces. Each time I've iterated on the design, all the other tools conform nicely until I get to this one.

src/linux_mcp_server/tools/system_info.py
subpop: I'm actually thinking we should just use those tools or fail. It will simplify the code *a lot*. It's worth the extra dependencies on `pciutils` and `usbutils`.

src/linux_mcp_server/tools/system_info.py
subpop: I see. Python calls computed properties "properties" and stored properties "parameters"?

samdoran: Parameters are what I'm calling values passed into the function or method call. Arguments is probably more accurate.

A method decorated with `@property` makes the method behave like an attribute. It's a way of separating expensive operations from instantiation and changing the interface. The results could be cached if desired.

Without the `@property` decorator, the method needs to be called to get the value: `instance.usb_device_count()`. The end result is no different, only the interface.

samdoran: There's also [cached_property](https://docs.python.org/3.10/library/functools.html#functools.cached_property) that's useful if the value doesn't change.

subpop: I see. I think using computed properties for the lengths is fine. It's not going to be so big it'll need to be cached.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -264,81 +408,121 @@ async def get_cpu_information(  # noqa: C901
     annotations=ToolAnnotations(readOnlyHint=True),
 )
 @log_tool_call
-@disallow_local_execution_in_containers
```
r0x0d, line 267: was this line expected to be removed?

subpop: Good catch. I restored the decorator on all the tools in the file.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -1,138 +1,310 @@
...
+        architecture=filtered_data.get("architecture"),
+        uptime=filtered_data.get("uptime"),
+        boot_time=filtered_data.get("boot_time"),
+    )
+
+
+def _parse_cpuinfo(output: str) -> ParsedData:
+    """Parse /proc/cpuinfo output into structured CPU data."""
+    parsed: ParsedData = {}
+    lines = output.strip().split("\n")
```
r0x0d, line 255: Hmm, since we are doing text parsing here, should we convert all text to lower?

PR #43, opened 2025-11-20
[RSPEED-2177] Raise ToolError when running in a container

subpop:
Add a function decorator that raises a ToolError if it is detected to be running in a container.

Fixes: RSPEED-2177

### DIFF
--- a/tests/utils/test_decorators.py
+++ b/tests/utils/test_decorators.py
@@ -1,7 +1,5 @@
 """Tests for decorator utilities."""
 
-import os
-
 import pytest
 
 from mcp.server.fastmcp.exceptions import ToolError
@@ -51,19 +49,17 @@ class TestDisallowLocalExecutionInContainers:
         result = await basic_test_func(host="remote.example.com", username="user")
         assert result == "success"
 
-    async def test_allows_execution_when_not_in_container(self, basic_test_func, mocker):
+    async def test_allows_execution_when_not_in_container(self, basic_test_func, monkeypatch):
         """Test that local execution is allowed when not running in a container."""
         # Ensure we're not in container environment
-        mocker.patch.dict(os.environ, {}, clear=False)
-        if "container" in os.environ:
-            del os.environ["container"]
+        monkeypatch.delenv("container", raising=False)
         result = await basic_test_func(host=None, username="user")
         assert result == "success"
 
-    async def test_raises_error_when_local_in_container(self, basic_test_func, mocker):
+    async def test_raises_error_when_local_in_container(self, basic_test_func, monkeypatch):
         """Test that ToolError is raised when attempting local execution in a container."""
         # Simulate running in a container
-        mocker.patch.dict(os.environ, {"container": "docker"})
+        monkeypatch.setenv("container", "docker")
         with pytest.raises(ToolError) as exc_info:
             await basic_test_func(host=None, username="user")
 
@@ -95,39 +91,28 @@ class TestDisallowLocalExecutionInContainers:
         assert "cmd=ls" in result
         assert "host=remote.example.com" in result
 
-    async def test_preserves_function_metadata(self):
-        """Test that decorator preserves function name and docstring."""
-
-        @disallow_local_execution_in_containers
-        async def my_special_function(host=None):
-            """This is a test function."""
-            return "test"
-
-        assert my_special_function.__name__ == "my_special_function"
-        assert my_special_function.__doc__ == "This is a test function."
-
-    async def test_raises_with_none_host_in_container(self, basic_test_func, mocker):
+    async def test_raises_with_none_host_in_container(self, basic_test_func, monkeypatch):
         """Test that explicitly passing host=None raises error in container."""
         # Simulate running in a container with explicit host=None
-        mocker.patch.dict(os.environ, {"container": "podman"})
+        monkeypatch.setenv("container", "podman")
         with pytest.raises(ToolError) as exc_info:
             await basic_test_func(host=None)
-            assert "Local execution is not allowed" in str(exc_info.value)
+        assert "Local execution is not allowed" in str(exc_info.value)
 
     @pytest.mark.parametrize(
         "container_value",
         ["openvz", "lxc", "lxc-libvirt", "systemd-nspawn", "docker", "podman", "rkt", "wsl", "proot", "pouch"],
     )
-    async def test_works_with_supported_container_values(self, host_only_test_func, mocker, container_value):
+    async def test_works_with_supported_container_values(self, host_only_test_func, monkeypatch, container_value):
         """Test that any truthy value for 'container' env var triggers the check."""
-        mocker.patch.dict(os.environ, {"container": container_value})
+        monkeypatch.setenv("container", container_value)
         with pytest.raises(ToolError):
             await host_only_test_func(host=None)
 
-    async def test_works_with_empty_container_value(self, host_only_test_func, mocker):
+    async def test_works_with_empty_container_value(self, host_only_test_func, monkeypatch):
         """Test that empty string for 'container' env var works."""
         # Presence of 'container' env var with empty value raises error
-        mocker.patch.dict(os.environ, {"container": ""})
+        monkeypatch.setenv("container", "")
         result = await host_only_test_func(host=None)
         assert result == "success"
 
@@ -165,7 +150,7 @@ class TestDisallowLocalExecutionInContainers:
 
         assert "Original error" in str(exc_info.value)
 
-    async def test_with_default_host_value_in_container(self, mocker):
+    async def test_with_default_host_value_in_container(self, monkeypatch):
         """Test behavior when function has host with default value in container."""
 
         @disallow_local_execution_in_containers
@@ -173,7 +158,7 @@ class TestDisallowLocalExecutionInContainers:
             return f"cmd={command}, host={host}, timeout={timeout}"
 
         # Should raise in container with default host=None
-        mocker.patch.dict(os.environ, {"container": "docker"})
+        monkeypatch.setenv("container", "docker")
         with pytest.raises(ToolError):
             await test_func("ls")
 


subpop:
Tests added, and they even caught a slight logic problem in the decorator itself. 💪

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,204 @@
...
+
+from mcp.server.fastmcp.exceptions import ToolError
+
+from linux_mcp_server.utils.decorators import disallow_local_execution_in_containers
+
+
+class TestDisallowLocalExecutionInContainers:
+    """Test disallow_local_execution_in_containers decorator."""
+
+    @pytest.mark.asyncio
```
samdoran: These decorators aren't needed because it is handled by the pytest config.

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,204 @@
+"""Tests for decorator utilities."""
+
+import os
+
+from unittest.mock import patch
```
samdoran: Use the `mocker` fixture instead.

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,204 @@
...
+class TestDisallowLocalExecutionInContainers:
+    """Test disallow_local_execution_in_containers decorator."""
+
+    @pytest.mark.asyncio
+    async def test_allows_execution_when_host_provided(self):
+        """Test that execution is allowed when host parameter is provided."""
+
+        @disallow_local_execution_in_containers
+        async def test_func(host=None, username=None):
+            return "success"
```
samdoran: Instead of defining this in every test, make it a fixture.

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,204 @@
...
+from unittest.mock import patch
+
+import pytest
+
+from mcp.server.fastmcp.exceptions import ToolError
+
+from linux_mcp_server.utils.decorators import disallow_local_execution_in_containers
+
+
+class TestDisallowLocalExecutionInContainers:
```
samdoran, line 43: Do these tests need to be in a class? It's only for organizational purposes.

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,204 @@
...
+            return "success"
+
+        # Simulate running in a container
+        with patch.dict(os.environ, {"container": "docker"}):
+            with pytest.raises(ToolError) as exc_info:
+                await test_func(host=None, username="user")
+
+            assert "Local execution is not allowed" in str(exc_info.value)
+            assert "container" in str(exc_info.value)
+            assert "SSH" in str(exc_info.value)
```
samdoran: Here's an example of how these tests could be simplified by using fixtures.
```suggestion
    async def test_raises_error_when_local_in_container(self, test_func, set_container_env):
        """Test that ToolError is raised when attempting local execution in a container."""
        with pytest.raises(ToolError) as exc_info:
            await test_func(host=None, username="user")

        assert "Local execution is not allowed" in str(exc_info.value)
        assert "container" in str(exc_info.value)
        assert "SSH" in str(exc_info.value)
```

samdoran (changes requested):
I left a few examples of ways the tests can be simplified. I think it may even be possible for the tests to be combined into one or two tests with different parameters.

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,204 @@
...
+    @pytest.mark.asyncio
+    async def test_works_with_different_container_values(self):
+        """Test that any truthy value for 'container' env var triggers the check."""
+
+        @disallow_local_execution_in_containers
+        async def test_func(host=None):
+            return "success"
+
+        # Test with different container environment values
+        for container_value in ["docker", "podman", "true", "1", "yes"]:
```
samdoran: Something like this should be parameters to the test itself, not a loop inside the test. Using test parameters makes it far easier to identify which value caused the failure.

src/linux_mcp_server/utils/decorators.py
```diff
@@ -0,0 +1,50 @@
...
+        # Get the function signature to find the 'host' parameter
+        sig = inspect.signature(func)
+        bound_args = sig.bind_partial(*args, **kwargs)
+        bound_args.apply_defaults()
+
+        # Check if 'host' parameter exists and is None
+        host_value = bound_args.arguments.get("host")
+
+        # Check if running in a container and host is None (local execution)
+        if host_value is None and os.getenv("container") is not None:
```
kokesak: I'm not sure who sets this envar, but when the `container` is empty string:
```
container=""
```
then it would be interpreted as True in this condition, which I'm not sure is desired

subpop: My assumption is that the mere existence of the variable in the environment is sufficient to identify the process as being bound to a container. Is that not a safe assumption? I'll read more on this variable too; maybe my assumption is indeed too naïve.

tests/utils/test_decorators.py
subpop: I thought about this. We only have one decorator to test in `decorators.py` right now, but the future might look different, so I accepted the class to organize the tests into a suite focused on this particular decorator. As you mentioned, it's purely for organizational purposes, so I was okay just leaving it in.

src/linux_mcp_server/utils/decorators.py
subpop: [Here](https://github.com/systemd/systemd/blob/f295cfa1a758147226308f802c19a12fd1a95715/src/basic/virt.c#L704-L710) is a useful comment about the expected state of the `container` environment variable. I suppose the best thing to do is mimic that logic.

```
                /* If we are PID 1 we can just check our own environment variable, and that's authoritative.
                 * We distinguish three cases:
                 * - the variable is not defined → we jump to other checks
                 * - the variable is defined to an empty value → we are not in a container
                 * - anything else → some container, either one of the known ones or "container-other"
                 */
```

samdoran: That's probably a safe assumption but it might be better to check that it is both set and is a non-empty value. You can either do a truthiness check: 

```python
if host_value is None and os.getenv("container"):
```

Or check that there is a truthy value returned and that it evaluates to True.
```python
container = os.getenv("container") or None  # If unset or "", will be None
if host_value is None and container:
```

samdoran: @kokesak The env var is set [in the container file](https://github.com/rhel-lightspeed/linux-mcp-server/blob/2181a59c5e00f7fcd6a865b07a097b75bc11df83/Containerfile#L47).

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,182 @@
...
+        # Should not raise when host is provided
+        result = await basic_test_func(host="remote.example.com", username="user")
+        assert result == "success"
+
+    async def test_allows_execution_when_not_in_container(self, basic_test_func, mocker):
+        """Test that local execution is allowed when not running in a container."""
+        # Ensure we're not in container environment
+        mocker.patch.dict(os.environ, {}, clear=False)
+        if "container" in os.environ:
+            del os.environ["container"]
```
samdoran: The [monkeypatch fixture](https://docs.pytest.org/en/stable/reference/reference.html#pytest.MonkeyPatch.delenv) can handle this a bit more succinctly.
```suggestion
        monkeypatch.delenv("container", raising=False)
```

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,182 @@
...
+        result = await basic_test_func(host=None, username="user")
+        assert result == "success"
+
+    async def test_raises_error_when_local_in_container(self, basic_test_func, mocker):
+        """Test that ToolError is raised when attempting local execution in a container."""
+        # Simulate running in a container
+        mocker.patch.dict(os.environ, {"container": "docker"})
```
samdoran: Monkeypatch could also be used here. There's no real functional difference, though.

```suggestion
        monkeypatch.setenv("container", "docker")
```

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,182 @@
...
+    async def test_preserves_function_metadata(self):
+        """Test that decorator preserves function name and docstring."""
+
+        @disallow_local_execution_in_containers
+        async def my_special_function(host=None):
+            """This is a test function."""
+            return "test"
+
+        assert my_special_function.__name__ == "my_special_function"
+        assert my_special_function.__doc__ == "This is a test function."
```
samdoran: This seems like more of a test for `functors.wraps` than our code. This test can probably be removed unless it is actually adding meaningful coverage to our code.

tests/utils/test_decorators.py
```diff
@@ -0,0 +1,182 @@
...
+    async def test_works_with_supported_container_values(self, host_only_test_func, mocker, container_value):
+        """Test that any truthy value for 'container' env var triggers the check."""
+        mocker.patch.dict(os.environ, {"container": container_value})
+        with pytest.raises(ToolError):
+            await host_only_test_func(host=None)
+
+    async def test_works_with_empty_container_value(self, host_only_test_func, mocker):
+        """Test that empty string for 'container' env var works."""
+        # Presence of 'container' env var with empty value raises error
+        mocker.patch.dict(os.environ, {"container": ""})
```
samdoran: We may want to error in this case. I'm learning more towards "the environment variable should be set and it should be non-empty".

src/linux_mcp_server/utils/decorators.py
kokesak: > [Here](https://github.com/systemd/systemd/blob/f295cfa1a758147226308f802c19a12fd1a95715/src/basic/virt.c#L704-L710) is a useful comment about the expected state of the `container` environment variable. I suppose the best thing to do is mimic that logic.

I agree with this, we can do the same

tests/utils/test_decorators.py
subpop: It doesn't seem to reduce the coverage percentage at all, so I'll drop it.

tests/utils/test_decorators.py
subpop: According to the comment in systemd's source code, the presence of the environment variable with an empty value should be interpreted as not running in a container. I interpreted that as meaning it shouldn't error, but rather should pass and allow execution.

```
/* If we are PID 1 we can just check our own environment variable, and that's authoritative.
 * We distinguish three cases:
 * - the variable is not defined → we jump to other checks
 * - the variable is defined to an empty value → we are not in a container
 * - anything else → some container, either one of the known ones or "container-other"
 */
```

PR #44, opened 2025-11-20
[RSPEED-2107] Remove `username` as a tool parameter

samdoran:
Instead make it a configuration setting.

Added a basic settings class for the username setting.

### DIFF
--- a/tests/tools/test_logs.py
+++ b/tests/tools/test_logs.py
@@ -1,0 +1,29 @@
+import pytest
+
+from linux_mcp_server.tools import get_audit_logs
+from linux_mcp_server.tools import get_journal_logs
+from linux_mcp_server.tools import read_log_file
+
+
+async def test_get_journal_logs(mocker):
+    mocker.patch("linux_mcp_server.tools.logs.execute_command", return_value=(1, "", ""))
+
+    result = await get_journal_logs()
+
+    assert "Error reading journal logs" in result
+
+
+async def test_get_audit_logs(mocker):
+    mocker.patch("linux_mcp_server.tools.logs.validate_line_count", side_effect=ValueError("Raised intentionally"))
+
+    with pytest.raises(ValueError, match="Raised intentionally"):
+        await get_audit_logs()
+
+
+async def test_read_log_file(mocker):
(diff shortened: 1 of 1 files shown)


src/linux_mcp_server/connection/ssh.py
```diff
@@ -286,7 +287,7 @@ async def execute_command(
     # Route to remote execution if host is provided
     if host:
         if not username:
-            username = getpass.getuser()
+            username = os.getenv("LINUX_MCP_USER", getpass.getuser())
```
subpop: Is this a reasonable fallback? In the container, it's going to fall back to `mcp`, a completely non-sensical choice. Wouldn't it be more prudent to just fail unequivocally if `LINUX_MCP_USER` is undefined?

src/linux_mcp_server/config.py
```diff
@@ -0,0 +1,11 @@
+from pydantic_settings import BaseSettings
+from pydantic_settings import SettingsConfigDict
+
+
+class Config(BaseSettings):
+    model_config = SettingsConfigDict(env_prefix="LINUX_MCP_", env_ignore_empty=True)
+
+    user: str = ""
```
subpop: Does this somehow automagically get assigned the value of the environment variable (in pseudocode) `$env_prefix + user.to_upper()` as part of the Pydantic magickery?

src/linux_mcp_server/connection/ssh.py
samdoran: It's a reasonable fallback when running natively. It definitely makes no sense when running in a container.

The reason I don't want to fail when undefined is because using the current user provides a seamless experience when running natively and is inline with user expectations.

src/linux_mcp_server/config.py
samdoran: Other way around: this [attribute is assigned](https://docs.pydantic.dev/latest/concepts/pydantic_settings/#environment-variable-names) the value of `LINUX_MCP_USER` by doing a case-insensitive match.

src/linux_mcp_server/connection/ssh.py
samdoran: It's more complex, but maybe we should fail when that is undefined _and_ running in a container. Otherwise, just use the value from `getuser()`.

PR #47, opened 2025-11-24
[RSPEED-2122] Add remaining os.environ settings to Config class

r0x0d:
This patch refactors the remaining places where os.environ was being used to be now part of the Config class instance.

Closes: [RSPEED-2122](https://issues.redhat.com/browse/RSPEED-2122)


tests/test_logging_config.py
```diff
@@ -1,9 +1,12 @@
 """Tests for logging configuration."""
 
+import importlib
 import json
 import logging
 import os
 
+from unittest.mock import patch
```
subpop: I think we're trying to avoid using `unittest.mock`. Is there a pytest mock equivalent that can be used instead?

r0x0d: updated to use mocker

PR #48, opened 2025-11-24
[RSPEED-2181] Add list_files tool calling for storage

r0x0d:
This commit introduces a new tool calling function for listing files in the system. The behavior is exactly like the `list_directories` one, but using `find` instead of `du` for the size operation.

Closes: [RSPEED-2181](https://issues.redhat.com/browse/RSPEED-2181)

### DIFF
--- a/src/linux_mcp_server/tools/storage.py
+++ b/src/linux_mcp_server/tools/storage.py
@@ -21,7 +21,29 @@ from linux_mcp_server.utils.decorators import disallow_local_execution_in_contai
 from linux_mcp_server.utils.types import Host
 
 
-class DirectoryEntry(BaseModel):
+class ToolExecutionAttributes(BaseModel):
+    names: list[str]
+    command: list[str]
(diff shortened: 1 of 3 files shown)


src/linux_mcp_server/tools/storage.py
```diff
@@ -209,26 +299,14 @@ async def read_file(
     """
     # For local execution, validate path
     if not host:
-        try:
-            path_obj = Path(path).resolve(strict=True)
-        except (OSError, RuntimeError):
-            raise ToolError(f"Path does not exist or cannot be resolved: {path}")
+        path = _validate_path(path)
```
subpop, line 302: This bit of change is reverting from using a `Path` instance to a string path. Is that intentional?

r0x0d: Yeah. We do revert to str down the road, and it seems that Path was only used to call `resolve(strict=True)`, and then got converted back to str.

PR #56, opened 2025-11-26
Add more unit tests for tools/processes.py

r0x0d:
The tool had a good coverage, but there were some tests missing, this patch solves this.

## Summary by CodeRabbit

* **Tests**
  * Reworked tests to be platform-agnostic using fixture-based mocks; expanded coverage for process listing and per-PID info, including current/missing PIDs, invalid inputs, float PID handling, and field truncation.
  * Added parametrized scenarios for remote host interactions, varied command outputs, error paths, and formatting edge cases. No production API changes; test-only.

### DIFF
--- a/tests/tools/test_processes.py
+++ b/tests/tools/test_processes.py
@@ -1,156 +1,262 @@
 """Tests for process management tools."""
 
 import os
-import sys
 
 import pytest
 
 from linux_mcp_server.tools import processes
 
 
-@pytest.mark.skipif(sys.platform != "linux", reason="requires Linux ps command")
-class TestProcesses:
-    """Test process management tools."""
+@pytest.fixture
+def mock_execute_command(mock_execute_command_for):
+    """Fixture to mock execute_command for the processes module."""
+    return mock_execute_command_for("linux_mcp_server.tools.processes")
 
-    async def test_list_processes_returns_string(self):
-        """Test that list_processes returns a string."""
+
+class TestProcesses:
+    """Test process management tools using real system calls."""
+
+    async def test_list_processes_returns_string_with_process_info(self):
+        """Test that list_processes returns a string with process information."""
         result = await processes.list_processes()
         assert isinstance(result, str)
         assert len(result) > 0
-
-    async def test_list_processes_contains_process_info(self):
-        """Test that list_processes contains process information."""
-        result = await processes.list_processes()
-
         # Should contain process-related keywords
         assert "pid" in result.lower() or "process" in result.lower()
         # Should contain resource usage info
         assert "cpu" in result.lower() or "memory" in result.lower() or "mem" in result.lower()
+
+    async def test_list_processes_shows_total_processes(self):
+        """Test that list_processes shows total process count."""
+        result = await processes.list_processes()
+        assert "Total processes:" in result
+        assert "Running Processes" in result
+
+    @pytest.mark.parametrize(
+        "pid,expected_in_result",
+        [
+            pytest.param(1, ["1", "systemd", "init"], id="init_process"),
+        ],
+    )
+    async def test_get_process_info_with_known_pids(self, pid, expected_in_result):
+        """Test getting info about known processes."""
+        result = await processes.get_process_info(pid)
+        assert isinstance(result, str)
+        assert len(result) > 0
+        # At least one of the expected strings should be in result
+        assert any(exp.lower() in result.lower() for exp in expected_in_result)
 
     async def test_get_process_info_with_current_process(self):
         """Test getting info about the current process."""
         current_pid = os.getpid()
         result = await processes.get_process_info(current_pid)
         assert isinstance(result, str)
-        assert len(result) > 0
-        # Should contain the PID
         assert str(current_pid) in result
-
-    async def test_get_process_info_with_init_process(self):
-        """Test getting info about init process (PID 1)."""
-        result = await processes.get_process_info(1)
-        assert isinstance(result, str)
-        assert len(result) > 0
-        # Should contain process information
-        assert "1" in result or "systemd" in result.lower() or "init" in result.lower()
 
     async def test_get_process_info_with_nonexistent_process(self):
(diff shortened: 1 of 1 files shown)


tests/tools/test_processes.py
```diff
@@ -2,56 +2,547 @@
...
+        mock_proc = mocker.Mock()
+        mock_proc.info = {"pid": 1, "name": "test"}
+
+        def process_iter_side_effect(*args, **kwargs):
+            # Return one good process and one that raises AccessDenied
+            yield mock_proc
+            mock_bad = mocker.Mock()
+            mock_bad.info
+            type(mock_bad).info = mocker.PropertyMock(side_effect=psutil.AccessDenied())
+            yield mock_bad
```
samdoran: The coverage report indicates this isn't being called. I don't know why exactly but that's something that needs to be investigated.

It may be better to have separate functions, one that returns ok and another that returns AccessDenied and call them explicitly.

I'm also not sure about line 53. What is the point of `mock_bad.info`?

I'm also not really sure about line 54. Modifying the global state of the class seems to be asking for trouble even though this is the way the docs say to use `PropertyMock`.

It may be simpler to create a class with the desired behavior instead of using a Mock object.

```python
class Denier:
    def __init__(self, deny=False):
        self.deny = deny

    @property
    def info(self):
        if self.deny:
            raise psutil.AccessDenied

        return ["something", "useful"]
```

Then pass a list of those objects to `side_effect` and `return_value` isn't needed. The first time `pcocess_iter` is called a `Denier` object is returned that won't raise `AccessDenied` on `info` but the second and later calls to `process_into` will return `Denier` objects that raise when `info` is accessed.

```python
mocker.patch("psutil.process_iter", side_effect=[Denier(), Denier(True)]

```

Separate objects could be created instead of using a toggle, too.

tests/tools/test_processes.py
```diff
@@ -2,56 +2,547 @@
...
+    async def test_list_processes_exception_handling(self, mocker):
+        """Test that list_processes handles general exceptions."""
+        mocker.patch("psutil.process_iter", side_effect=Exception("Test error"))
+
+        result = await mcp.call_tool("list_processes", {})
+        text = _get_text_from_mcp_result(result)
+        assert "Error listing processes" in text
+        assert "Test error" in text
 
     async def test_get_process_info_with_current_process(self):
```
samdoran, line 49: It seems like many of these tests can be combined into a single test with different parameters.

tests/tools/test_processes.py
```diff
@@ -2,56 +2,547 @@
...
+        """Test that get_process_info accepts float PIDs (from LLMs)."""
+        current_pid = float(os.getpid())
+        result = await mcp.call_tool("get_process_info", {"pid": current_pid})
+        text = _get_text_from_mcp_result(result)
+        assert isinstance(text, str)
+        assert str(int(current_pid)) in text
+
+    async def test_get_process_info_access_denied(self, mocker):
+        """Test get_process_info when access is denied."""
+        mock_proc = mocker.Mock(spec=psutil.Process)
```
samdoran: Could this be a fixture instead of having to build it in the test?

tests/tools/test_processes.py
```diff
@@ -2,56 +2,547 @@
...
+        text = _get_text_from_mcp_result(result)
+        assert "Error: PID must be at least 1" in text
+
+    async def test_list_processes(self):
+        """Test calling list_processes through mcp.call_tool()."""
+        result = await mcp.call_tool("list_processes", {})
+
+        # mcp.call_tool returns a tuple of (list[TextContent], dict)
+        assert isinstance(result, tuple)
+        assert len(result) == 2
```
samdoran: I'm not sure these statements add much value. It would be better to do this once rather than in every test. The interface in FastMCP is totally different, which means all of these tests need to be rewritten. Reducing the number of times that needs to be done would be ideal.

tests/tools/test_processes.py
```diff
@@ -2,56 +2,547 @@
...
+        # mcp.call_tool returns a tuple of (list[TextContent], dict)
+        assert isinstance(result, tuple)
+        content_list, metadata = result
+        assert isinstance(content_list, list)
+        content = content_list[0]
+        assert "Error: PID must be at least 1" in content.text
+
+    async def test_get_process_info_nonexistent_pid(self):
+        """Test calling get_process_info via MCP with nonexistent PID."""
+        result = await mcp.call_tool("get_process_info", {"pid": 999999})
```
samdoran: Is there any way the `get_process_info` tests could be combined into one test with different parameters?

tests/tools/test_processes.py
```diff
@@ -1,156 +1,253 @@
...
-Pid:	1
-PPid:	0
-Threads:	1
-VmRSS:	    11892 kB"""
-
-        mock_execute.side_effect = [
-            (0, ps_output, ""),  # ps command
-            (0, proc_status, ""),  # /proc/PID/status
+        assert "Running Processes" in result
+        assert "Total processes: 102" in result
```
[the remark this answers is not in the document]
major: @r0x0d This is good stuff but you don't need to cover that in this commit.

tests/tools/test_processes.py
```diff
@@ -1,156 +1,253 @@
...
 
-    async def test_list_processes_handles_command_failure(self, mock_execute):
+    async def test_list_processes_handles_command_failure(self, mock_execute_command):
         """Test that list_processes handles command failure gracefully."""
-        mock_execute.return_value = (1, "", "Command not found")
+        mock_execute_command.return_value = (1, "", "Command not found")
 
         result = await processes.list_processes(host="remote.host")
 
         assert "Error" in result
```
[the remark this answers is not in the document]
major: This one looks easy to fix and deleting code feels nice.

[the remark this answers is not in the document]
major: Uh I had no idea coderabbit was watching

PR #59, opened 2025-12-02
Transparently vendor external packages

samdoran:
Modify `sys.path` by prepending all packages inside of `_vendor`.

Since this code comes from Ansible and is GPL 3.0 licensed, update the license for this project.

Using the code from https://github.com/ansible/ansible/pull/69850. Thanks @nitzmahone for the work and @sivel for point it out to me!

### Todo ###

- [x] Add tests


src/linux_mcp_server/_vendor/__init__.py
```diff
@@ -0,0 +1,50 @@
+# (c) 2020 Ansible Project
+# GNU General Public License v3.0+ (see COPYING or https://www.gnu.org/licenses/gpl-3.0.txt)
+
+import pathlib
+import pkgutil
+import sys
+import warnings
+
+
+"""
```
r0x0d, line 10: I would also add a note that the goal of this, in the long term, is to be optional and not a required path for the tooling. 

As we discussed in slack, the ideal path is that we consume dependencies as rpms instead of vendoring them in our own rpm. `fastmcp` is a particular case as it does not exist yet in any of the fedora repositories, unfortunately.

src/linux_mcp_server/_vendor/__init__.py
```diff
@@ -0,0 +1,50 @@
...
+A warning will be displayed in the event that a module from this package is
+already loaded. That should be a rare, if ever, occuerence due to the natue
+of this application.
+
+Packages in this directory are added to the beginning of sys.path. They will
+take precedent over any other packages.
+
+Install packages here during downstream packaging using a command such as:
+
+    pip install --upgrade --target [path to this directory] pydantic-settings fastmcp
```
r0x0d:
```suggestion
    pip install --upgrade --target [path to this directory] fastmcp
```

pydantic-settings exists in fedora repos, so I would not include it as an example to not be confusing

Containerfile
```diff
@@ -69,7 +69,7 @@ LABEL url="https://github.com/rhel-lightspeed/linux-mcp-server"
 LABEL vendor="Red Hat, Inc."
 LABEL version=${VERSION}
 
-ADD LICENSE /licenses/Apache-2.0.txt
+ADD licenses/ /licenses/
```
r0x0d, line 72: I'm wondering if it would make sense to keep our Apache-2.0 license where it is right now, and only add third-party licenses to the `licenses` folder... Maybe renaming it to `licenses-3rdparty`? 

Not any particular reason, just that if we maintain our license at the root level, it is easier for parsers to identify that this project is associated with Apache 2.0 rather than any of `licenses/*.txt`

src/linux_mcp_server/_vendor/__init__.py
samdoran: I think that might confuse same packagers. It's an optional feature available for downstream packagers. I don't want to give the impression this will be removed at some point in the future.

samdoran: I'll add a note that packages from the OS package manager should be preferred over vending when practical.

r0x0d: That works too

src/linux_mcp_server/_vendor/__init__.py
samdoran: I updated the example. Thanks for the feedback!

Containerfile
samdoran: I was thinking that GitHub supported multiple license files in a top level `licenses/` directory but that doesn't seem to be the case. Let me see what the proper way to do this is. Multiple licenses in a project does complicate things.

samdoran: I put the main license back at the repo root but added the GPL in the `/licenses` folder. For the container build, both are added to `/licenses` which is only for clerical purposes.

r0x0d (approved):

PR #64, opened 2025-12-03
resolve PermissionError for local logs

alexxa:
Add a troubleshooting step to address PermissionError (Errno 13) when starting the server due to incorrect local log file ownership.


r0x0d:
looks good, thanks @alexxa!

PR #69, opened 2025-12-04
[RSPEED-2205] Switch to using FastMCP

samdoran:
The version of FastMCP included in the Python SDK from Anthropic is quite old. We would like to use newer features.

Changed a lot of the tests the work with the newer API as well as eliminate a lot of cruft.

Closes #117.

### DIFF
--- a/src/linux_mcp_server/tools/storage.py
+++ b/src/linux_mcp_server/tools/storage.py
@@ -148,8 +148,6 @@ async def list_directories(
         # Format the output
         return format_directory_listing(entries, path, order_by, reverse=sort == SortBy.DESCENDING)
 
-    except ToolError:
-        raise
     except Exception as e:
         raise ToolError(f"Error listing directories: {str(e)}") from e
 
@@ -216,8 +214,6 @@ async def list_files(
         # Format the output
         return format_file_listing(entries, path, order_by, reverse=sort == SortBy.DESCENDING)
 
-    except ToolError:
-        raise
     except Exception as e:
         raise ToolError(f"Error listing files: {str(e)}") from e
 


tests/tools/test_storage.py
```diff
@@ -139,81 +135,65 @@ async def test_list_block_devices_lsblk_success(
...
-        result = await mcp.call_tool("list_block_devices", {})
-        assert isinstance(result, tuple)
-        assert len(result) == 2
-        assert isinstance(result[0], list)
-        output = result[0][0].text
+        result = await mcp_client.call_tool("list_block_devices", {})
+        result_text = result.content[0].text
 
-        assert "Error" in output or "not found" in output
+        assert "Error" in result_text or "not found" in result_text
```
r0x0d: We could probably merge them together and use a fixture to return the side_effect in one case, and the return_value on the other

tests/tools/test_storage.py
```diff
@@ -139,81 +135,65 @@ async def test_list_block_devices_lsblk_success(
...
-        call_kwargs = mock_storage_execute_command.call_args[1]
-        assert call_kwargs["host"] == "remote.host.com"
+        assert "=== Block Devices ===" in result_text
+        assert "sda" in result_text
+        assert "=== Disk I/O Statistics" not in result_text
+        assert mock_storage_execute_command.call_count == 1
+        assert mock_storage_execute_command.call_args.kwargs["host"] == "remote.host.com"
 
-    async def test_list_block_devices_exception_handling(self, mock_storage_execute_command):
+    async def test_list_block_devices_exception_handling(self, mock_storage_execute_command, mcp_client):
```
r0x0d: Can probably be grouped  with https://github.com/rhel-lightspeed/linux-mcp-server/pull/69/changes#r2624204574

tests/tools/test_storage.py
```diff
@@ -139,81 +135,65 @@ async def test_list_block_devices_lsblk_success(
...
-        assert "Error" in output
+        result = await mcp_client.call_tool("list_block_devices", {})
+
+        assert "Error" in result.content[0].text
 
 
 @pytest.mark.skipif(sys.platform != "linux", reason="requires GNU version of coreutils/findutils")
 class TestListDirectories:
-    async def test_list_directories_returns_string_output(self, setup_test_directory):
+    async def test_list_directories_returns_string_output(self, setup_test_directory, mcp_client):
```
r0x0d: I'm wondering if this test is really needed. It looks like it does the same thing as `test_list_directories_by_name`, but only check for a different string. The code path should be the same

tests/tools/test_storage.py
```diff
@@ -238,20 +217,21 @@ async def test_list_directories_by_name(self, setup_test_directory):
...
+        assert "beta" in result_text
+        assert "gamma" in result_text
+
         # Verify sorted order (alpha should appear before beta, beta before gamma)
-        alpha_pos = result.find("alpha")
-        beta_pos = result.find("beta")
-        gamma_pos = result.find("gamma")
+        alpha_pos = result_text.find("alpha")
+        beta_pos = result_text.find("beta")
+        gamma_pos = result_text.find("gamma")
```
r0x0d: I guess this is not needed given that we search the text position below. Maybe we could just cut this part and rely on the `result_text.find()`

```python
        # means the text was not found in the `result_text` variableV
        assert all(pos > -1 for pos in [alpha_pos, beta_pos, gamma_pos])

```

tests/tools/test_storage.py
samdoran: Yup, there is a lot of room for consolidation. I'll fix this one up.

tests/tools/test_storage.py
samdoran: Done.

tests/tools/test_storage.py
samdoran: I'll combine these.

tests/tools/test_storage.py
samdoran: Agreed.

subpop (approved):
This looks great to me. I'd prefer to get this merged if it will allow us to adjust #122.

PR #71, opened 2025-12-05
Add a helper function to test whether an executable exists on a remote system

samdoran:
Add two functions for looking up the absolute path of an executable: one remote and one local. If the executable isn't found, raise a `ToolError` which the LLM client will pass to the model in order to provide an accurate failure message to the user.

### DIFF
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -335,7 +335,7 @@ async def execute_command(
 
     Args:
         command: Command and arguments to execute. If the command is not an absolute path
-                 it will be resolved to the full path before exeuction.
+                 it will be resolved to the full path before execution.
         host: Optional remote host address
         **kwargs: Additional arguments (reserved for future use)
 


src/linux_mcp_server/tools/logs.py
```diff
@@ -170,6 +173,7 @@ async def read_log_file(  # noqa: C901
 
         # For local execution, validate path
         if not host:
+            bin_path = shutil.which("tail")
```
r0x0d: why not using get_bin_path here too? probably could just move the `await get_bin_path` to the top, and that would execute regardless if it is local or remote

samdoran: Most of the tool functions have code paths based on whether or not a remote host parameter is provided. I'm hoping to eventually remove all those separate branches and I didn't want to complicate `get_bin_path` with the same local or remote execution branches.

We could have `get_remote_bin_path` and `get_bin_path`, but I don't want to make any more kitchen sink functions.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -291,6 +292,22 @@ async def execute_command(
...
+
+    Raises VauleError if not found and required is True.
+    """
+    bin_path = None
+    rc, out, err = await execute_command(["command", "-v", command], hostname)
+    if rc == 0:
+        return out.strip()
+
+    if required:
+        raise ValueError(f"Unable to find '{command}' on {hostname}")
```
r0x0d: Hmm, to me, it feels like if this check should belong to the tool, as we can return a `ToolError()`  (even though, it's not a tool error per se, but a missing binary error)

I see that you are using this same function in both modes (required=true, required=false) in different tools, and in both, you have a duplication of this check (kinda). 

This feels a bit confusing to me, so, either;

* let's remove the `required` parameter from this function, and handle it at the tool level
* let's rely on the error being raised here, avoiding checking on multiple places the same thing

src/linux_mcp_server/tools/logs.py
r0x0d: Exactly, if `host` is `None`, it seems to me that `execute_command` will run this locally, and not on remote. I think the path is already there... Or maybe I'm missing on something?

r0x0d: > We could have `get_remote_bin_path` and `get_bin_path`, but I don't want to make any more kitchen sink functions.

I understand where you're going with this one, but, it feels to me that the entrypoint of deciding if a tool will run on remote or local, is the `host`. If that's the case, `host` is being set at the tool parameter level and being passed down until we reach the `execute_command`. Seems pretty solid to me to avoid bugs, and I don't necessarily know if there is any real difference that we need to care about "I want to check `tail` locally` to "I want to check `tail` remotely" 🤔

src/linux_mcp_server/connection/ssh.py
samdoran: I was thinking the same thing but wanted to get some feedback. I'll remove it.

src/linux_mcp_server/tools/logs.py
samdoran: The nature of a remote call versus a local call is quite different. Encapsulating them in separate functions simplifies the code and makes testing and maintenance easier.

r0x0d: Ah, okay! I think I got it now.

samdoran:
I moved it into `execute_command` and I think it's much better. I did some local testing and it's doing a good job of looking up the full path with `command -v` on the remote as well as fails with a nice error message when the tool isn't found on the remote.

```
2025-12-05 17:15:35 | INFO | linux_mcp_server.audit | REMOTE_EXEC: command -v top | host=rally | exit_code=0 | command=command -v top | host=rally | exit_code=0
2025-12-05 17:15:35 | INFO | linux_mcp_server.audit | SSH_CONNECT: rally | host=rally | status=success
2025-12-05 17:15:35 | INFO | asyncssh | [conn=0, chan=25] Requesting new SSH session
2025-12-05 17:15:35 | INFO | asyncssh | [conn=0, chan=25]   Command: /usr/bin/top -bn1
2025-12-05 17:15:35 | INFO | asyncssh | [conn=0, chan=25] Received exit status 0
2025-12-05 17:15:35 | INFO | asyncssh | [conn=0, chan=25] Received channel close
2025-12-05 17:15:35 | INFO | asyncssh | [conn=0, chan=25] Channel closed
2025-12-05 17:15:35 | INFO | linux_mcp_server.audit | REMOTE_EXEC: /usr/bin/top -bn1 | host=rally | exit_code=0 | command=/usr/bin/top -bn1 | host=rally | exit_code=0
2025-12-05 17:15:35 | INFO | linux-mcp-server | TOOL_COMPLETE: get_cpu_information | tool=get_cpu_information | status=success | duration=0:00:00.425357s
```


```
( O)> What process are running on rally?

─── list_processes | linux-tools ──────────────────────────
host: rally

I was unable to retrieve the running processes on `rally`. It seems there was an error executing the `list_processes` tool. The error message indicates that the command 'pizza' was not found on the host 'rally'. This is unexpected, as `list_processes` should not be trying to execute a command named 'pizza'.

⏱️  Elapsed time: 1.96s
```

src/linux_mcp_server/connection/ssh.py
```diff
@@ -129,19 +136,21 @@ async def get_connection(self, host: str, username: str) -> asyncssh.SSHClientCo
...
 
+            if CONFIG.user:
+                connect_kwargs["username"] = CONFIG.user
+
             conn = await asyncssh.connect(**connect_kwargs)
             self._connections[key] = conn
 
             # Log successful connection using audit function
-            log_ssh_connect(host, username, status=Status.success, reused=False, key_path=self._ssh_key)
+            log_ssh_connect(host, username=conn._username, status=Status.success, reused=False, key_path=self._ssh_key)
```
r0x0d, line 158: why not using `connect_kwargs["username"]` instead of accessing a private member of `conn`?

r0x0d: or either, `username = CONFIG.user or connect_kwargs["username"]`

src/linux_mcp_server/connection/ssh.py
```diff
@@ -242,10 +250,40 @@ async def close_all(self):
 _connection_manager = SSHConnectionManager()
 
 
+def get_bin_path(command: str) -> str:
+    """Get the full path to an executable.
+
+    Raises VauleError if not found.
+    """
+    sbin_paths = ("/sbin", "/usr/sbin", "/usr/local/sbin")
+    paths = os.getenv("PATH", "").split(os.pathsep)
```
r0x0d: I'm wondering if we should raise some error in case `$PATH` is not defined. I don't see a clear picture where `$PATH` would be undefined, but I guess it would make sense for defensive programming?

src/linux_mcp_server/connection/ssh.py
```diff
@@ -242,10 +250,40 @@ async def close_all(self):
...
+    for path in sbin_paths:
+        if path not in paths:
+            paths.append(path)
+
+    path = os.pathsep.join(paths)
```
r0x0d: Not a necessary change, but for me, it makes the code clear if we don't override the variables with new data in loops like this


```suggestion
    env_paths = os.getenv("PATH", "").split(os.pathsep)
    paths = []
    for path in sbin_paths:
        if path not in env_paths:
            paths.append(path)

    path = os.pathsep.join(paths)
```

src/linux_mcp_server/connection/ssh.py
```diff
@@ -242,10 +250,40 @@ async def close_all(self):
...
+
+    return bin_path
+
+
+async def get_remote_bin_path(command: str, hostname: Host, connection: SSHConnectionManager) -> str:
+    """Get the full path to an executable on a remote system.
+
+    Raises VauleError if not found.
+    """
+    rc, out, err = await connection.execute_remote(["command", "-v", command], hostname)
```
r0x0d: Can we use full path for `command`? 
```suggestion
    rc, out, err = await connection.execute_remote(["/usr/bin/command", "-v", command], hostname)
```

@narmaku raised a good point that we can avoid aliases and other stuff by using full path

src/linux_mcp_server/connection/ssh.py
```diff
@@ -242,10 +250,40 @@ async def close_all(self):
...
+    Raises VauleError if not found.
+    """
+    sbin_paths = ("/sbin", "/usr/sbin", "/usr/local/sbin")
+    paths = os.getenv("PATH", "").split(os.pathsep)
+    for path in sbin_paths:
+        if path not in paths:
+            paths.append(path)
+
+    path = os.pathsep.join(paths)
+    bin_path = shutil.which(command, path=path)
```
subpop, line 292: Could we always hardcode the value of `PATH` in all our subprocesses? I know that systemd does this with units.


```suggestion
    bin_path = shutil.which(command, path=["/usr/bin:/usr/sbin:/bin:/sbin"])
```

src/linux_mcp_server/connection/ssh.py
subpop: I think that's a good defensive programming strategy.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -242,10 +250,40 @@ async def close_all(self):
...
+    if rc == 0:
+        return out.strip()
+
+    raise ValueError(f"Unable to find command '{command}' on host {hostname}")
+
+
 async def execute_command(
     command: list[str],
     host: str | None = None,
-    username: str = CONFIG.user,
```
subpop, line 275: So with the total removal of the `username` argument, does asyncssh fall back to the `~/.ssh/config` settings?

src/linux_mcp_server/connection/ssh.py
```diff
@@ -280,11 +318,20 @@ async def execute_command(
         ... )
     """
     cmd_str = " ".join(command)
+    bin = command[0]
+    if not Path(bin).is_absolute():
+        try:
+            if host:
+                command[0] = await get_remote_bin_path(bin, host, _connection_manager)
+            else:
+                command[0] = get_bin_path(bin)
```
subpop: Every time I see this logic, it makes me think that local execution and remote execution are totally different things and shouldn't be together in the same toolset... are we certain this is the right decision (to sprinkle "if host:else:" throughout the codebase?

src/linux_mcp_server/connection/ssh.py
samdoran: Using the attribute on the connection will get the correct value even if it is coming from `ssh_config`. `CONFIG.user` and `connect_kwargs["username"]` may be empty.

src/linux_mcp_server/connection/ssh.py
samdoran: In the case where PATH doesn't exist, only the `sbin_paths` will be searched instead of raising an error. I think that's ok, but that's my opinion. A warning would probably be better than an error.

src/linux_mcp_server/connection/ssh.py
samdoran: That's not a bad idea. We would then need to prepend `env_paths` to the final result, the idea being only to append paths from `sbin_paths` if they are not already in `PATH`. Something like `path = env_paths + os.pathsep.join(paths)`.

src/linux_mcp_server/connection/ssh.py
samdoran: In this case the shell builtin `command` is preferable over `/usr/bin/command`, which is why it is not an absolute path. `/usr/bin/command`, at least in RHEL, is a shell script that runs `builtin` with `/usr/bin/bash`. If the login shell is not `bash` that may result in unexpected behavior, say for example if you use using `zsh` and have `PATH` configured in `zprofile`.

src/linux_mcp_server/connection/ssh.py
samdoran: Yup, we sure could. The way to do it locally versus remote would be very different but it's possible.

src/linux_mcp_server/connection/ssh.py
samdoran: Yes! The goal is to better align with user expectations.

The current behavior is to default to the name of the current user unless `LINUX_MCP_USER` is defined. The user setting in `~/.ssh/config` is completely ignored. It's not a great experience and seems broken.

src/linux_mcp_server/connection/ssh.py
samdoran: I don't like doing this at all. It was the least worst I could come up with, doing it once here instead of in every single tool.

The entire architecture needs rethinking. The same code should run locally and remote.

The alternative was to do the `if host` check in `get_bin_path` and make it a kitchen sink function like the current tool functions. I'm trying to at least start drawing the line and making targeted functions that do specific things without a toggle that changes behavior.

Ideally only one these functions should exist in the future and it will be cleaner to just delete the function rather than having to do surgery to remove parts of a more complex `get_bin_path` function.

major:
Now that #77 and #79 went in, could you give this fella a rebase? _(and sorry about that)_

samdoran:
@subpop As far as I can tell, raising `ToolError` adds `isError: true` to the result JSON response. https://modelcontextprotocol.io/specification/2025-06-18/server/tools#error-handling

src/linux_mcp_server/connection/ssh.py
samdoran: What about using a generator? This might be a bit too dense.
```python
    path = os.getenv("PATH", "").split(os.pathsep)
    path.extend(new_path for new_path in sbin_paths if new_path not in path)
    path = os.pathsep.join(path)
```

src/linux_mcp_server/connection/ssh.py
samdoran: Let me try pushing this down a layer. I'll have to use the connection directly to avoid a recursion error.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -269,10 +279,40 @@ async def close_all(self):
...
+    if bin_path is None:
+        raise ValueError(f"Unable to find '{command}'")
+
+    return bin_path
+
+
+async def get_remote_bin_path(command: str, hostname: Host, connection: SSHConnectionManager) -> str:
+    """Get the full path to an executable on a remote system.
+
+    Raises VauleError if not found.
```
major: typo: `ValueError`

src/linux_mcp_server/connection/ssh.py
```diff
@@ -307,11 +347,20 @@ async def execute_command(
...
     cmd_str = " ".join(command)
+    bin = command[0]
+    if not Path(bin).is_absolute():
+        try:
+            if host:
+                command[0] = await get_remote_bin_path(bin, host, _connection_manager)
+            else:
+                command[0] = get_bin_path(bin)
+        except ValueError as ve:
+            raise ToolError(ve.args)
```
major: Did you mean `raise ToolError(str(ve))`?

src/linux_mcp_server/connection/ssh.py
```diff
@@ -106,7 +111,7 @@ async def get_connection(self, host: str, username: str) -> asyncssh.SSHClientCo
         Raises:
             ConnectionError: If connection fails
         """
-        key = f"{username}@{host}"
+        key = f"{host}"
```
major, line 112: Wouldn't we want to keep the `username` in the key for connection pooling?

src/linux_mcp_server/tools/logs.py
```diff
@@ -84,6 +85,8 @@ async def get_journal_logs(
         result.append(stdout)
 
         return "\n".join(result)
+    except ToolError:
```
major: Do we need the `except/raise`? Couldn't we just let the `ToolError` raise on its own? Or are you trying to catch this one explicitly before another exception hits?

major (commented):
Just a few questions. Nothing huge except a typo and a question around the `ToolError` argument.

src/linux_mcp_server/connection/ssh.py
samdoran: I pushed this down into each function. I think this is cleaner though I did have to repeat a lot of the connection error handling in `get_remote_bin_path`. It's worth it, though.

src/linux_mcp_server/connection/ssh.py
samdoran: I actually meant to unpack that. `str(ve)` is a convenience mechanism to do that.

src/linux_mcp_server/connection/ssh.py
samdoran: Yes, but we have a chicken and egg problem. If we want the username to be able to come from the SSH config, then we don't know what it is at the time we ask `asyncssh` to create the connection. We could store it in the cache after the fact, but we wouldn't be able to look it up.

I don't know if there is a way to query `asncyssh` to determine what username it _would_ use before requesting the connection.

src/linux_mcp_server/tools/logs.py
samdoran: This is part of the very bad pattern that needs to be fixed in these tools: they all catch `Exception` and return a string, which doesn't correctly indicate to the LLM client a failure in the tool occurred.

If we removed all the `except Exception` code, we would not need to catch and reraise the `ToolError`.

samdoran: I'm trying to make changes without getting entirely bogged down in everything that needs fixing. It's quite difficult.

src/linux_mcp_server/connection/ssh.py
r0x0d: The generator looks way better, indeed. I like this even more!

src/linux_mcp_server/tools/logs.py
r0x0d: > I'm trying to make changes without getting entirely bogged down in everything that needs fixing. It's quite difficult.

I think this PR is fixing what was proposed initially, but also, diving into fixing a lot more things as well (that are related to the initial bug report). 

The pattern I see here is that, like #36, we keep finding to fix/improve and won't really move forward :sweat_smile: 

I'm always in favor of merging what is working and keep improving on it. Maybe it's even easier to think about what to do next if you are not trying to keep the correlation of what is working x what you want to achieve in such PRs

samdoran: After looking through the FastMCP code a bit, any exception raised by a tool is raised as a `ToolError`. That's very good news. That will make it very easy to remove the erroneous Exception handling.

major:
If I'm reading this code right, does it connect twice for each remote command? Once to check if the command is there, and then the second time to run it? On slower connections, this could take a while.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -283,9 +334,9 @@ async def execute_command(
     whether host/username parameters are provided.
 
     Args:
-        command: Command and arguments to execute
+        command: Command and arguments to execute. If the command is not an absolute path
+                 it will be resolved to the full path before exeuction.
```
major: typo: execution

major (approved):
Other than a typo, I think this is fine.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -269,10 +280,50 @@ async def close_all(self):
 _connection_manager = SSHConnectionManager()
 
 
+def get_bin_path(command: str) -> str:
```
major, line 283: Would `shutil.which()` save you some time here?

samdoran: This function is using `shutil.which()` with a few extra paths in the mix and changes the interface slightly from `shutil.which()`: this function raises instead of returning `None` if the file isn't found.

PR #75, opened 2025-12-09
fix: filter out ipv6 link-local addresses

### LINKED ISSUE #74: psutil returns IPv6 link local addresses
Link local addresses are always assigned to interfaces as part of RFC, but they're terrible for reliable communication in an enterprise environment. We should filter these out throughout the project.

major:
IPv6 link-local addresses are automatically assigned but they are poor options for reliable connectivity in an IPv6 network. Filter out these addresses carefully.

Fixes #74

### DIFF
--- a/src/linux_mcp_server/utils/format.py
+++ b/src/linux_mcp_server/utils/format.py
@@ -1,3 +1,38 @@
+import ipaddress
+
+
+def is_ipv6_link_local(address: str) -> bool:
+    """Check if an IPv6 address is link-local (fe80::/10).
+
+    Link-local addresses are only valid on a single network segment and
+    require a scope identifier to route. They're not useful for enterprise
+    cross-machine communication.
+
+    Args:
+        address: IPv6 address string, possibly with scope identifier (e.g., "fe80::1%eth0")
+
+    Returns:
+        True if the address is link-local, False otherwise
+
+    Examples:
+        >>> is_ipv6_link_local("fe80::1")
+        True
+        >>> is_ipv6_link_local("fe80::1%eth0")
+        True
+        >>> is_ipv6_link_local("2001:db8::1")
+        False
+        >>> is_ipv6_link_local("192.168.1.1")
+        False
+    """
+    try:
+        # Parse and check if it's link-local
+        addr = ipaddress.IPv6Address(address)
+        return addr.is_link_local
+    except (ValueError, ipaddress.AddressValueError):
+        # Not a valid IPv6 address (could be IPv4, empty, or malformed)
+        return False
+
+
 def format_bytes(bytes_value: int | float) -> str:
     """
     Format bytes into human-readable format.
--- a/tests/tools/test_processes.py
+++ b/tests/tools/test_processes.py
@@ -2,7 +2,18 @@
 
 import os
 
+from unittest.mock import MagicMock
+
+import pytest
+
 from linux_mcp_server.tools import processes
+from tests.conftest import GLOBAL_IPV6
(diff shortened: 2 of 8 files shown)


subpop (commented):
This looks good to me, but I think it might make codecov happier (and thereby Sam happier 😄) if we added a test to ensure those missing lines are covered.

src/linux_mcp_server/utils/format.py
```diff
@@ -1,3 +1,41 @@
...
+        >>> is_ipv6_link_local("fe80::1%eth0")
+        True
+        >>> is_ipv6_link_local("2001:db8::1")
+        False
+        >>> is_ipv6_link_local("192.168.1.1")
+        False
+    """
+    try:
+        # Strip scope identifier (e.g., %eth0) if present
+        addr_str = address.split("%")[0]
```
samdoran: It's not necessary to remove the scope zone ID. It's a [valid part](https://docs.python.org/3/library/ipaddress.html#ipaddress.IPv6Address) of the address.

major: Ah good catch. I thought the ipaddress module would explode with that, but it doesn't!

tests/tools/test_processes.py
```diff
@@ -55,3 +66,130 @@ async def test_list_processes_with_host(self, mocker):
         result = await processes.list_processes(host="starship.command")
 
         assert "Running Processes" in result
+
+
+def _create_mock_process(connections):
```
samdoran: Any reason this isn't a fixture?

major: Ah good point, I did re-use it a lot of times.

tests/tools/test_processes.py
```diff
@@ -55,3 +66,130 @@ async def test_list_processes_with_host(self, mocker):
...
+        mocker.patch.object(processes.psutil, "Process", return_value=mock_proc)
+
+        result = await processes.get_process_info(12345)
+
+        # should_be_filtered means connection should NOT be shown
+        if should_be_filtered:
+            assert "Network Connections" not in result
+        else:
+            assert "Network Connections (1)" in result
+            assert f"{laddr_ip}:8080" in result
```
samdoran: Instead of a toggle, list out the conditions in the test cases.
```python
LINK_LOCAL_FILTER_CASES = [
    pytest.param(LINK_LOCAL_IPV6, IPV4_ADDR, '"Network Connections" not in result', id="laddr-link-local"),
    pytest.param(IPV4_ADDR, LINK_LOCAL_IPV6, '"Network Connections" not in result', id="raddr-link-local"),
    pytest.param(LINK_LOCAL_IPV6, LINK_LOCAL_IPV6, '"Network Connections" not in result', id="both-link-local"),
    pytest.param(IPV4_ADDR, GLOBAL_IPV6, '"Network Connections (1)" in result', id="no-link-local"),
    pytest.param(IPV4_ADDR, GLOBAL_IPV6, 'f"{laddr_ip}:8080" in result', id="no-link-local"),
    pytest.param(GLOBAL_IPV6, IPV4_ADDR, '"Network Connections (1)" in result', id="global-ipv6"),
    pytest.param(GLOBAL_IPV6, IPV4_ADDR, 'f"{laddr_ip}:8080" in result', id="global-ipv6"),
]
```

Change `should_be_filtered` to `expectation` or something else that makes sense.

```suggestion
       assert eval(expectation)
```

major: Cases where stuff isn't filtered gets weird, but I'll see what I can do.

PR #77, opened 2025-12-09
feat: Add ssh timeouts

### LINKED ISSUE #76: Missing timeout for ssh connections
When the MCP server makes an ssh connection, there should be some sort of timeout set. It could be configured to something reasonable, such as 30 seconds, to avoid situations where the network destination is unreachable and a firewall drops the packets.

major:
Add a global ssh timeout in the main configuration file, but also allow the project to override timeouts for tools that sometimes take a long time to run. The default is now 30 seconds.

This avoids situations where the connection hangs indefinitely due to firewall rules or misconfigured networks.

Fixes #76

### DIFF
--- a/src/linux_mcp_server/config.py
+++ b/src/linux_mcp_server/config.py
@@ -30,5 +30,8 @@ class Config(BaseSettings):
     key_passphrase: str | None = None
     search_for_ssh_key: bool = False
 
+    # Command execution timeout (applies to remote SSH commands)
+    command_timeout: int = 30  # Timeout in seconds; prevents hung SSH operations
+
 
 CONFIG = Config()
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -164,20 +164,26 @@ class SSHConnectionManager:
         command: list[str],
         host: str,
         username: str,
+        timeout: int = CONFIG.command_timeout,
     ) -> tuple[int, str, str]:
         """
         Execute a command on a remote host via SSH.
+
+        Commands are subject to a timeout to prevent indefinite hangs. The timeout
+        can be specified per-call or defaults to CONFIG.command_timeout (30s).
 
         Args:
             command: Command and arguments to execute
             host: Remote host address
             username: SSH username
+            timeout: Command timeout in seconds. Defaults to CONFIG.command_timeout.
+                Use for commands that need longer execution time.
 
         Returns:
             Tuple of (return_code, stdout, stderr)
 
         Raises:
-            ConnectionError: If SSH connection fails
+            ConnectionError: If SSH connection fails or command times out
         """
         conn = await self.get_connection(host, username)
 
@@ -189,7 +195,21 @@ class SSHConnectionManager:
         start_time = time.time()
 
         try:
-            result = await conn.run(cmd_str, check=False)
+            try:
+                result = await conn.run(cmd_str, check=False, timeout=timeout)
+            except asyncssh.TimeoutError:
+                duration = time.time() - start_time
+                logger.error(
+                    f"Command timed out after {timeout}s",
+                    extra={
+                        "event": Event.REMOTE_EXEC_ERROR,
+                        "command": cmd_str,
+                        "host": host,
+                        "duration": f"{duration:.3f}s",
+                        "error": "timeout",
+                    },
+                )
+                raise ConnectionError(f"Command timed out after {timeout}s on {username}@{host}: {cmd_str}") from None
 
             return_code = result.exit_status if result.exit_status is not None else 0
 
--- a/tests/connection/test_ssh.py
+++ b/tests/connection/test_ssh.py
@@ -1,11 +1,14 @@
 """Tests for SSH executor module."""
 
+from contextlib import nullcontext
 from unittest.mock import AsyncMock
 from unittest.mock import MagicMock
 from unittest.mock import Mock
 
 import asyncssh
 import pytest
+
+from asyncssh import SSHClientConnection
 
 from linux_mcp_server.connection.ssh import discover_ssh_key
 from linux_mcp_server.connection.ssh import execute_command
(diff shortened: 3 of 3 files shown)


src/linux_mcp_server/connection/ssh.py
```diff
@@ -164,32 +164,60 @@ async def execute_remote(
         command: list[str],
         host: str,
         username: str,
+        timeout: int | None = None,
```
samdoran: This could default to using `CONFIG.timeout` instead of having to later check the value.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -164,32 +164,60 @@ async def execute_remote(
...
+        effective_timeout = timeout if timeout is not None else CONFIG.command_timeout
+
         # Start timing for command execution
         start_time = time.time()
 
         try:
-            result = await conn.run(cmd_str, check=False)
+            try:
+                result = await asyncio.wait_for(
+                    conn.run(cmd_str, check=False),
```
samdoran: The timeout value should be passed to [SSHClientConnection.run](https://asyncssh.readthedocs.io/en/latest/api.html#asyncssh.SSHClientConnection.run) rather than wrapping the call. Catch `asyncssh.TimeoutError`.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -164,32 +164,60 @@ async def execute_remote(
...
+                    f"Command timed out after {effective_timeout}s",
+                    extra={
+                        "event": Event.REMOTE_EXEC_ERROR,
+                        "command": cmd_str,
+                        "host": host,
+                        "duration": f"{duration:.3f}s",
+                        "error": "timeout",
+                    },
+                )
+                raise ConnectionError(
```
samdoran: We need to make sure we're catching and handling this correctly so that the LLM gets the memo in a way it can understand.

src/linux_mcp_server/config.py
```diff
@@ -30,5 +30,8 @@ class Config(BaseSettings):
     key_passphrase: str | None = None
     search_for_ssh_key: bool = False
 
+    # Command execution timeout (applies to remote SSH commands)
+    command_timeout: int = 30  # Timeout in seconds; prevents hung SSH operations
```
samdoran, line 34: Looking at the `asyncssh` docs, I don't think it honors timeout settings in `.ssh/config` so it's probably ok to set a default value here. Otherwise I was going to suggest setting it to `None`.

tests/connection/test_ssh.py
```diff
@@ -333,3 +335,99 @@ async def async_connect(*args, **kwargs):
...
+        mock_conn.is_closed = Mock(return_value=False)
+
+        async def async_connect(*args, **kwargs):
+            return mock_conn
+
+        mock_connect = MagicMock()
+        mock_connect.side_effect = async_connect
+        mocker.patch("asyncssh.connect", mock_connect)
+
+        return mock_conn
```
samdoran: It'd be good to add specs to these mocks to make sure they have the correct attributes. This can also be done more concisely.

```suggestion
        mock_connect = AsyncMock(asyncssh.connect)
        mock_connect.return_value = AsyncMock(SSHClientConnection)
        mock_connect.return_value.is_closed.return_value = False
        
        mocker.patch("asyncssh.connect", mock_connect)

        return mock_conn
```

tests/connection/test_ssh.py
```diff
@@ -333,3 +335,99 @@ async def async_connect(*args, **kwargs):
...
+        mocker.patch("linux_mcp_server.connection.ssh.CONFIG.command_timeout", global_timeout)
+        setup_delayed_command(cmd_delay)
+
+        if should_timeout:
+            with pytest.raises(ConnectionError, match="Command timed out"):
+                await ssh_manager.execute_remote(["cmd"], "host", "user", timeout=per_cmd_timeout)
+        else:
+            returncode, stdout, _ = await ssh_manager.execute_remote(["cmd"], "host", "user", timeout=per_cmd_timeout)
+            assert returncode == 0
+            assert stdout == "ok"
```
samdoran: Instead of using a toggle, [specify the failure state](https://docs.pytest.org/en/stable/example/parametrize.html#parametrizing-conditional-raising) in the parameters.
```python
from contextlib import nullcontext
```

```suggestion
    ("global_timeout", "cmd_delay", "per_cmd_timeout", "expectation"),
    [
        # Global timeout tests (per_cmd_timeout=None means use global)
        (0.1, 10, None, pytest.raises(ConnectionError, match="Command timed out")),  # Short global timeout, long command -> timeout
        (30, 0, None, nullcontext()),  # Long global timeout, instant command -> success
        (0.2, 0.5, None, pytest.raises(ConnectionError, match="Command timed out")),  # Global timeout shorter than command -> timeout
        (1.0, 0.1, None, nullcontext()),  # Global timeout longer than command -> success
        # Per-command timeout override tests
        (60, 1, 0.1, pytest.raises(ConnectionError, match="Command timed out")),  # Long global, short override -> timeout
        (0.05, 0.2, 1, nullcontext()),  # Short global, long override -> success
    ],
    ids=[
        "global_slow_cmd_times_out",
        "global_fast_cmd_succeeds",
        "global_tight_timeout",
        "global_ample_timeout",
        "override_short_times_out",
        "override_long_succeeds",
    ],
)
async def test_timeout_behavior(
    self,
    mocker,
    ssh_manager,
    setup_delayed_command,
    global_timeout,
    cmd_delay,
    per_cmd_timeout,
    expectation,
):
    """Test timeout behavior with global and per-command timeout combinations."""
    mocker.patch("linux_mcp_server.connection.ssh.CONFIG.command_timeout", global_timeout)
    setup_delayed_command(cmd_delay)

    with expectation as e:
        returncode, stdout, _ = await ssh_manager.execute_remote(["cmd"], "host", "user", timeout=per_cmd_timeout)
        
        assert returncode == 0
        assert stdout == "ok"
```

tests/connection/test_ssh.py
```diff
@@ -333,3 +335,99 @@ async def async_connect(*args, **kwargs):
...
+
+    async def test_timeout_error_contains_context(self, mocker, ssh_manager, setup_delayed_command):
```
samdoran: It may be possible to combine this with the previous test.

src/linux_mcp_server/connection/ssh.py
major: Can do. I just figured it made sense to be super explicit here in the function.

src/linux_mcp_server/config.py
major: I figured 30 would be a pretty safe default. Perhaps if they're using MCP to talk to satellites in orbit... 🤔

src/linux_mcp_server/connection/ssh.py
major: Done

tests/connection/test_ssh.py
major: I LEARNED SOMETHING NEW TODAY!

tests/connection/test_ssh.py
major: I tried, but it started to get super confusing.

tests/connection/test_ssh.py
```diff
@@ -333,3 +336,110 @@ async def async_connect(*args, **kwargs):
...
+        global_timeout,
+        per_cmd_timeout,
+        expected_timeout,
+        raises_timeout,
+        expectation,
+    ):
+        """Test timeout parameter passing and error handling."""
+        mocker.patch("linux_mcp_server.connection.ssh.CONFIG.command_timeout", global_timeout)
+
+        if raises_timeout:
```
samdoran: You can probably just use the value of `expectation` instead of having to specify an additional parameter.
```suggestion
        if expectation:
```

samdoran: Or you could provide the mock as a parameter.
```python
mock_ssh_connection.run = mock_run
```

major: I hope my fix meets your _expectation_

PR #79, opened 2025-12-09
feat: Handle ssh key verification

### LINKED ISSUE #78: SSH key verification missing
This project should have a configuration for handling ssh key verification. It might be worthwhile to disable it for now until it's closer to the time for it to go into production.

major:
Make ssh key verification configurable, but leave it off for now until this project is ready for production.

Fixes #78 

This one depends on #77.

### DIFF
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -127,10 +127,17 @@ class SSHConnectionManager:
         logger.debug(f"{Event.SSH_CONNECTING}: {key} | key={self._ssh_key or 'none'}")
 
         try:
+            # Determine host key verification settings
+            if CONFIG.verify_host_keys:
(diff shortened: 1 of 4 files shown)


src/linux_mcp_server/connection/ssh.py
```diff
@@ -127,10 +127,21 @@ async def get_connection(self, host: str, username: str) -> asyncssh.SSHClientCo
         logger.debug(f"{Event.SSH_CONNECTING}: {key} | key={self._ssh_key or 'none'}")
 
         try:
+            # Determine host key verification settings
+            if CONFIG.verify_host_keys:
+                known_hosts = (
+                    str(CONFIG.known_hosts_path)
+                    if CONFIG.known_hosts_path
+                    else str(Path.home() / ".ssh" / "known_hosts")
```
samdoran: This could be handled in the CONFIG class. That way it's only necessary to get the property here.

PR #80, opened 2025-12-09
ci: Use makefile for quick local tests

major:
Add a makefile for quick local testing that includes most of what's being run in github actions. `make verify` runs everything.

### DIFF
--- a/Makefile
+++ b/Makefile
@@ -1,0 +1,46 @@
+.PHONY: help sync lint format types test ci verify fix clean
+
+# Default target
+help:
+	@echo "🔧 CI Targets:"
+	@echo "  make ci       - Run ALL CI checks (lint + format + types + test)"
+	@echo "  make verify   - Sync dependencies + run all CI checks"
+	@echo "  make lint     - Run ruff linter"
+	@echo "  make format   - Check code formatting"
+	@echo "  make types    - Run pyright type checker"
+	@echo "  make test     - Run pytest with coverage"
+	@echo ""
+	@echo "🛠️  Development Targets:"
+	@echo "  make sync     - Install/sync all dependencies"
+	@echo "  make fix      - Auto-fix lint and format issues"
+	@echo "  make clean    - Remove build artifacts and caches"
+
+sync:
+	uv sync --locked
+
+lint:
+	uv run --locked ruff check --diff
+
+format:
+	uv run --locked ruff format --diff
+
+types:
+	uv run --locked pyright
+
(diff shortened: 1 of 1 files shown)


Makefile
```diff
@@ -0,0 +1,69 @@
+# Makefile for running CI checks locally
```
samdoran: Most of these comments are not adding much value and can be removed.

Makefile
```diff
@@ -0,0 +1,69 @@
...
+format:
+	uv run --locked ruff format --diff
+
+types:
+	uv run --locked pyright
+
+# === CI Tests ===
+
+test:
+	uv run --locked pytest --cov-report=xml
```
samdoran: Locally we don't want or need the XML report. The HTML report is much more useful for local development.

Makefile
```diff
@@ -0,0 +1,69 @@
...
+	@echo "  make test     - Run pytest with coverage"
+	@echo ""
+	@echo "🛠️  Development Targets:"
+	@echo "  make sync     - Install/sync all dependencies"
+	@echo "  make fix      - Auto-fix lint and format issues"
+	@echo "  make clean    - Remove build artifacts and caches"
+
+# Install dependencies (combines lint + test groups like CI does)
+sync:
+	uv sync --locked --group dev
```
samdoran: `--group dev` is the default and can be omitted.

Makefile
```diff
@@ -0,0 +1,69 @@
...
+.PHONY: help sync lint format types test ci verify fix clean
+
+# CI environment variables (from .github/workflows/ci.yml)
+export PIP_DISABLE_PIP_VERSION_CHECK := 1
+export COVERAGE_IGOR_VERBOSE := 1
+export FORCE_COLOR := 1
+
+# Default target
+help:
+	@echo "🔧 CI Targets (mirror GitHub Actions):"
```
samdoran:
```suggestion
	@echo "🔧 CI Targets:"
```

Makefile
major: Ah, I use this with code gutters in vscode, but I can pull it

samdoran: Oh, that's fine then. I know it's only in CI to upload to Codecov.

samdoran: As long as it has a purpose.

samdoran: Though maybe we could just add that to the default config instead of having a parameter in a makefile. That way you get consistent results regardless of how `pytest` is called.

PR #85, opened 2025-12-10
feat(utils): add command-line sanitization utility

major:
Add sanitize_cmdline() function to redact potential secrets from command-line arguments before display. This prevents accidental exposure of passwords, API keys, and tokens in process listings.

Handles:
- Flag-style secrets (-p, --password, --token, etc.)
- Key=value patterns (password=, api_key=, etc.)
- Both separate values and --flag=value syntax
- Case-insensitive matching

### DIFF
--- a/src/linux_mcp_server/utils/__init__.py
+++ b/src/linux_mcp_server/utils/__init__.py
@@ -1,3 +1,5 @@
 from linux_mcp_server.utils.enum import StrEnum as StrEnum
 from linux_mcp_server.utils.format import format_bytes as format_bytes
 from linux_mcp_server.utils.format import is_ipv6_link_local as is_ipv6_link_local
+from linux_mcp_server.utils.sanitize import REDACTED as REDACTED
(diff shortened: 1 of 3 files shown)


src/linux_mcp_server/utils/__init__.py
```diff
@@ -1,3 +1,5 @@
 from linux_mcp_server.utils.enum import StrEnum as StrEnum
 from linux_mcp_server.utils.format import format_bytes as format_bytes
 from linux_mcp_server.utils.format import is_ipv6_link_local as is_ipv6_link_local
+from linux_mcp_server.utils.sanitize import REDACTED as REDACTED
+from linux_mcp_server.utils.sanitize import sanitize_cmdline as sanitize_cmdline
```
r0x0d, line 5:
```suggestion
from linux_mcp_server.utils.sanitize import REDACTED
from linux_mcp_server.utils.sanitize import sanitize_cmdline
```

major:
My idea was to try to potentially hide some of the sensitive information from the screen, but I'm not sure the best way to do it. :|

PR #87, opened 2025-12-10
fix(processes): eliminate TOCTOU race condition in get_process_info

major:
Add _get_process_safe() helper function that safely retrieves a psutil.Process instance without the Time-of-Check to Time-of-Use (TOCTOU) race condition that existed in the previous implementation.

Previously, the code checked pid_exists() before constructing the Process object, allowing a window where the process could exit between the check and use. Now we directly attempt construction and handle NoSuchProcess gracefully.

### DIFF
--- a/src/linux_mcp_server/tools/processes.py
+++ b/src/linux_mcp_server/tools/processes.py
@@ -169,11 +169,13 @@ async def get_process_info(  # noqa: C901
                 return f"Process with PID {validated_pid} does not exist on remote host."
         else:
             # Local execution - use psutil
-            # Check if process exists
-            if not psutil.pid_exists(validated_pid):
+            # Get process safely and avoid a TOCTOU race condition where the state of a
+            # resource changes between checking it and using it.
+            try:
+                proc = psutil.Process(validated_pid)
+            except psutil.NoSuchProcess:
                 return f"Process with PID {validated_pid} does not exist."
 
-            proc = psutil.Process(validated_pid)
             info = []
 
             info.append(f"=== Process Information for PID {validated_pid} ===\n")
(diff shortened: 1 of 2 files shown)


src/linux_mcp_server/tools/processes.py
```diff
@@ -169,11 +193,10 @@ async def get_process_info(  # noqa: C901
                 return f"Process with PID {validated_pid} does not exist on remote host."
         else:
             # Local execution - use psutil
-            # Check if process exists
-            if not psutil.pid_exists(validated_pid):
+            # Get process safely (avoids TOCTOU race)
+            proc = _get_process_safe(validated_pid)
```
subpop: Is there an advantage to making this a function call that helps eliminating the race condition? The separate function body seems so simple. It could just be inlined into the calling function.

major: My goal there was to make it testable as an independent thing, but I can push it back into the existing function if you like that better.

subpop: Yea. I can see how that makes it testable. Are we able to integrate the condition into an existing test?

subpop (commented):
The implementation looks good. The test is puzzling me though; I can't see how that test is testing the condition.

major:
@subpop Ah, I forgot that `test_get_process_info_with_nonexistent_process` was sitting there and basically testing the same thing. I'll remove that extra test.

src/linux_mcp_server/tools/processes.py
samdoran: I had this same thought: I don't think the function did enough to justify its existence.

major:
I'll just see if I can fix this as part of the tests rework that I'm doing.

PR #88, opened 2025-12-11
Refactor storage tests to mock results

subpop:
This way they can run on any platform and aren't dependent on live system calls.


subpop:
Major is doing a much better job cleaning up the tests than this PR does, so I'll close it.

PR #91, opened 2025-12-11
refactor(tests): simplify test_logs.py with fixtures and parameterization

major:
- Add pytest fixtures with proper specs (mock_execute_command, mock_path_exists, mock_allowed_log_paths) replacing inline mocking
- Use pytest.mark.parametrize to consolidate duplicate test cases (e.g., 6 filter tests → 1 parameterized test)
- Add mock specs (spec=execute_command) for type safety verification
- Extract shared assertion logic into assert_tool_result_structure()

### DIFF
--- a/tests/tools/test_logs.py
+++ b/tests/tools/test_logs.py
@@ -242,245 +242,149 @@ class TestGetAuditLogs:
 
 
 class TestReadLogFile:
-    async def test_read_log_file_success(self, mocker, tmp_path):
-        """Test read_log_file with successful read."""
-        mock_execute_command = mocker.patch("linux_mcp_server.tools.logs.execute_command")
-        # Create a test log file
-        log_file = tmp_path / "test.log"
-        log_content = "Log line 1\nLog line 2\nLog line 3"
-        log_file.write_text(log_content)
+    """Tests for read_log_file tool."""
 
-        # Set allowed paths
-        mocker.patch("linux_mcp_server.tools.logs.CONFIG.allowed_log_paths", str(log_file))
+    @pytest.fixture
+    def setup_log_file(self, tmp_path, mock_allowed_log_paths):
+        """Create a test log file and configure allowed paths."""
 
-        mock_execute_command.return_value = (0, log_content, "")
+        def _setup(content="Test log content\nLine 2\nLine 3"):
+            log_file = tmp_path / "test.log"
+            log_file.write_text(content)
+            mock_allowed_log_paths(str(log_file))
+            return log_file
 
-        result = await mcp.call_tool("read_log_file", {"log_path": str(log_file)})
+        return _setup
 
-        assert isinstance(result, tuple)
-        assert len(result) == 2
-        assert isinstance(result[0], list)
-        output = result[0][0].text
+    @pytest.mark.parametrize(
+        "lines,expected_line_count",
+        [
+            (100, "100"),  # Default
+            (50, "50"),  # Custom
+        ],
+    )
+    async def test_read_log_file_success(self, mock_execute_command, setup_log_file, lines, expected_line_count):
+        """Test read_log_file with various line counts."""
+        log_file = setup_log_file()
+        mock_execute_command.return_value = (0, "Test log content\nLine 2", "")
 
-        assert f"=== Log File: {log_file} (last 100 lines) ===" in output
-        assert "Log line 1" in output
+        params = {"log_path": str(log_file)}
+        if lines != 100:
+            params["lines"] = lines
 
-    async def test_read_log_file_custom_lines(self, mocker, tmp_path):
-        """Test read_log_file with custom line count."""
-        mock_execute_command = mocker.patch("linux_mcp_server.tools.logs.execute_command")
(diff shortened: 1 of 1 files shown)


tests/tools/test_logs.py
```diff
@@ -1,641 +1,463 @@
...
-        output = result[0][0].text
+@pytest.fixture
+def mock_execute_command(mocker):
+    """Mock execute_command with proper spec."""
+    mock = mocker.patch(
+        "linux_mcp_server.tools.logs.execute_command",
+        spec=execute_command,
+        new_callable=AsyncMock,
+    )
+    return mock
```
r0x0d: Can probably be moved to conftest, since there is a lot of files that use this

major: Good point. I'll adjust that.

tests/tools/test_logs.py
```diff
@@ -1,641 +1,450 @@
...
-        )
+from linux_mcp_server.server import mcp
 
-        result = await mcp.call_tool("get_journal_logs", {"priority": "err"})
 
-        assert isinstance(result, tuple)
-        assert len(result) == 2
-        assert isinstance(result[0], list)
-        output = result[0][0].text
+async def assert_tool_result_structure(result):
```
r0x0d:
```suggestion
def assert_tool_result_structure(result):
```

no need for it to be async, right?

tests/tools/test_logs.py
```diff
@@ -1,641 +1,450 @@
...
 
-        assert "priority=err" in output
-        assert "Error occurred" in output
 
-        # Verify priority filter was passed
-        args = mock_execute_command.call_args[0][0]
-        assert "-p" in args
-        assert "err" in args
+@pytest.fixture
+def mock_path_exists(mocker):
```
r0x0d: Sorry, I don't think that this fixture provides much value to the test module...

tests/tools/test_logs.py
r0x0d: Since this is a common used assertion for all tools, I think in a follow-up PR we could move this to a different module and re-use that

tests/tools/test_logs.py
```diff
@@ -1,641 +1,450 @@
...
-        assert "No journal entries found matching the criteria" in output
-
-    async def test_get_journal_logs_command_error(self, mocker):
-        """Test get_journal_logs handles command errors."""
-        mock_execute_command = mocker.patch("linux_mcp_server.tools.logs.execute_command")
-        mock_execute_command.return_value = (1, "", "journalctl: Failed to access journal")
-
-        result = await mcp.call_tool("get_journal_logs", {})
+        result = await mcp.call_tool("get_journal_logs", params)
+        output = await assert_tool_result_structure(result)
```
r0x0d:
```suggestion
        output =  assert_tool_result_structure(result)
```

we don't need to await on this assert because we have already awaited the result above

tests/tools/test_logs.py
```diff
@@ -1,641 +1,450 @@
...
+    )
+    async def test_get_journal_logs_edge_cases(
+        self, mock_execute_command, returncode, stdout, stderr, expected_error, expectation
+    ):
+        """Test get_journal_logs error handling and edge cases."""
+        mock_execute_command.return_value = (returncode, stdout, stderr)
+
+        with expectation:
+            result = await mcp.call_tool("get_journal_logs", {})
+            output = await assert_tool_result_structure(result)
```
r0x0d: same here

tests/conftest.py
```diff
@@ -211,3 +211,18 @@ async def list_services(*args, **kwargs):
...
+
+@pytest.fixture
+def mock_execute_command(mocker):
+    """Mock execute_command with proper spec."""
+    from unittest.mock import AsyncMock
+
+    from linux_mcp_server.connection.ssh import execute_command
+
+    mock = mocker.patch(
+        "linux_mcp_server.tools.logs.execute_command",
```
r0x0d, line 227: If this will be here available to all others, then we can't mock only the `logs.execute_command`, but `connection.sh.exececute_command`

tests/conftest.py
```diff
@@ -211,3 +211,18 @@ async def list_services(*args, **kwargs):
         raise ValueError("Raised intentionally")
 
     return list_services
+
+
+@pytest.fixture
+def mock_execute_command(mocker):
+    """Mock execute_command with proper spec."""
+    from unittest.mock import AsyncMock
```
r0x0d: This can be moved to the top. No reason to have it in the function

tests/conftest.py
```diff
@@ -211,3 +211,18 @@ async def list_services(*args, **kwargs):
...
+
+    from linux_mcp_server.connection.ssh import execute_command
```
r0x0d: Add a comment on why we are importing in here

tests/tools/test_logs.py
major: Yeah, it's pretty basic. I'll move it back to inline

tests/tools/test_logs.py
major: Oh good point. I missed that.

PR #92, opened 2025-12-11
Add devcontainers for VSCode

r0x0d:


.devcontainer/devcontainer.json
```diff
@@ -0,0 +1,47 @@
+{
+    "$schema": "https://raw.githubusercontent.com/devcontainers/spec/refs/heads/main/schemas/devContainer.schema.json",
+    "name": "linux-mcp-server",
+    "build": {
+        "dockerfile": "Containerfile",
+        "context": "."
+    },
+    // Podman-specific: keep UID/GID mapping for rootless containers
+    "runArgs": [
+        "--userns=keep-id",
```
major, line 10: This is podman-specific, so docker users might get surprised. I think it's okay for now.

PR #94, opened 2025-12-12
refactor(tests): consolidate TestListBlockDevices with fixtures

major:
- Add mock_disk_io_stats and mock_partition fixtures
- Add verify_result_structure helper for common assertions
- Combine lsblk success tests into parameterized test
- Combine psutil fallback tests into parameterized test
- Reduce ~165 lines to ~90 lines (~45% reduction)
- Maintain 100% test coverage with all 7 test scenarios

### DIFF
--- a/AGENTS.md
+++ b/AGENTS.md
@@ -7,3 +7,5 @@
 - Use fixtures to deduplicate setup code across tests.
 - If a fixture could be used in multiple test modules, place it in `conftest.py`.
 - Use mocks sparingly and try to pass objects to the code under test instead.
+- Use `autospec=True` when patching to verify arguments match the real function signature.
+- Use `spec=<object>` with MagicMock to restrict attributes to those of the real object.


tests/tools/test_storage.py
```diff
@@ -110,162 +110,184 @@ def restricted_path(tmp_path):
...
-        )
 
-        # Create mock disk I/O stats
-        mock_stats = MagicMock()
-        mock_stats.read_bytes = 1024 * 1024 * 1024  # 1 GB
-        mock_stats.write_bytes = 512 * 1024 * 1024  # 512 MB
-        mock_stats.read_count = 1000
-        mock_stats.write_count = 500
+@pytest.fixture
+def verify_result_structure():
```
r0x0d: What about extract this to a more general, top-level function? This is the same assert we use in all tools

major: Are you thinking something in `conftest.py`?

tests/tools/test_storage.py
```diff
@@ -110,162 +110,184 @@ def restricted_path(tmp_path):
...
 
-    async def test_list_block_devices_lsblk_fallback(self, mocker):
-        """Test list_block_devices falls back to psutil when lsblk fails."""
-        # lsblk returns non-zero
-        mocker.patch(
-            "linux_mcp_server.tools.storage.execute_command", AsyncMock(return_value=(1, "", "command failed"))
+        mock_execute_command = mocker.patch(
+            "linux_mcp_server.tools.storage.execute_command",
+            AsyncMock(return_value=(0, lsblk_output, "")),
         )
```
r0x0d: we have the `mock_execute_command` now! we can use it everywhere

tests/tools/test_storage.py
```diff
@@ -110,162 +110,184 @@ def restricted_path(tmp_path):
...
-        mock_execute_command = AsyncMock(
-            return_value=(
-                0,
-                "NAME   SIZE TYPE\nsda    1TB  disk",
-                "",
-            )
+        mock_execute_command = mocker.patch(
+            "linux_mcp_server.tools.storage.execute_command",
+            AsyncMock(return_value=(0, "NAME   SIZE TYPE\nsda    1TB  disk", "")),
         )
```
r0x0d: same here

major:
@r0x0d I turned that mock fixture into `mock_execute_command_for` so that multiple modules can use it.

tests/__init__.py
```diff
@@ -1 +1,19 @@
...
+    Args:
+        result: The result tuple from an MCP tool call
+
+    Returns:
+        The text content from the first item in the result list
+    """
+    assert isinstance(result, tuple)
+    assert len(result) == 2
+    assert isinstance(result[0], list)
+    return result[0][0].text
```
samdoran, line 19: I'm glad to see this removed from every single test. It goes away completely with the move to FastMCP since the return value is completely different. It was arduous fixing this all over the place.

tests/tools/test_storage.py
```diff
@@ -110,169 +111,172 @@ def restricted_path(tmp_path):
     restricted_path.chmod(0o755)
 
 
+@pytest.fixture
+def mock_disk_io_stats():
+    """Fixture providing mock disk I/O statistics."""
+    mock_stats = MagicMock()
```
samdoran: I'm not sure the magic methods are needed but it's a good idea to at last prevent typo errors with a spec.
```suggestion
    mock_stats = MagicMock(spec=True)
```

tests/tools/test_storage.py
```diff
@@ -110,169 +111,172 @@ def restricted_path(tmp_path):
...
+    mock_stats.read_bytes = 1024 * 1024 * 1024  # 1 GB
+    mock_stats.write_bytes = 512 * 1024 * 1024  # 512 MB
+    mock_stats.read_count = 1000
+    mock_stats.write_count = 500
+    return {"sda": mock_stats}
+
+
+@pytest.fixture
+def mock_execute_command(mock_execute_command_for):
```
samdoran: Maybe give this a more descriptive name so it's more obvious what exactly it is patching.

tests/tools/test_storage.py
```diff
@@ -110,169 +111,172 @@ def restricted_path(tmp_path):
...
+        partition_mountpoint,
+        partition_fstype,
+        expected_header,
+    ):
+        """Test list_block_devices falls back to psutil when lsblk is unavailable or fails."""
+        # Mock execute_command to fail in different ways
+        mocker.patch("linux_mcp_server.tools.storage.execute_command", side_effect=execute_side_effect)
+
+        # Create mock partition with test-specific values
         mock_partition = MagicMock()
```
samdoran:
```python
mock_partition = MagicMock(spec=True)
```

tests/conftest.py
```diff
@@ -217,14 +217,28 @@ async def list_services(*args, **kwargs):
...
-    return mocker.patch(
-        "linux_mcp_server.tools.logs.execute_command",
-        spec=execute_command,
-        new_callable=AsyncMock,
-    )
+
+    def _mock(module: str):
+        return mocker.patch(
+            f"{module}.execute_command",
+            spec=execute_command,
```
samdoran: When mocking functions, it's better to use `autospec=True` since it will verify the arguments are correct. Mock objects accept arbitrary args and kwargs by default, so incorrect calls will silently pass tests.
```suggestion
            autospec=True,
```

samdoran (commented):
This is a welcome refactor. I was trying to figure out how to remove the redundancy and could never quite get it right.

PR #97, opened 2025-12-12
Refactor local/remote execution logic

subpop:
Currently tools currently have divergent implementations:

- Remote: Shell commands via SSH (`ss`, `ps`, `free`, etc.)
- Local: Python libraries (`psutil`, `platform`)

This PR reimplements the tool calling logic by introducing a command registry layer that keeps a list of commands needed for each tool. It extracts the parameter substitution into a separate function by supporting placeholders in the command spec. It also extracts the command stdout parsing and final tool response formatting into two separate modules: `parsers.py` and `formatters.py`.

```mermaid
flowchart TB
    subgraph tools [Tool Functions]
        tool[Tool Function]
    end
    
    subgraph registry [Command Layer]
        commands[Command Registry]
        parsers[Output Parsers]
        formatters[Output Formatters]
    end
    
    subgraph execution [Execution Layer]
        executor[execute_command]
        local[Local Subprocess]
        remote[SSH Remote]
    end
    
    tool --> commands
    tool --> parsers
    tool --> formatters
    commands --> executor
    executor -->|"host=None"| local
    executor -->|"host=server"| remote
```

### Command Spec Registry

```python

COMMANDS: dict[str, CommandEntry] = {
    # === Single-command tools ===
    # Network (single commands)
    "network_connections": CommandSpec(
        args=["ss", "-tunap"], 
        fallback=["netstat", "-tunap"]
    ),
    
    # === Multi-command tools (CommandGroup) ===
    
    "network_interfaces": CommandGroup(commands={
        "brief": CommandSpec(args=["ip", "-brief", "address"]),
        "detail": CommandSpec(args=["ip", "address"]),
        "stats": CommandSpec(args=["cat", "/proc/net/dev"]),
    }),
 }
```

### Command Execution Pattern

For single commands:

```python
cmd = COMMANDS["network_connections"]
returncode, stdout, _ = await execute_with_fallback(cmd, host=host)
```

For command groups:

```python
group = COMMANDS["system_info"]
results = {}
for name, cmd in group.commands.items():
    returncode, stdout, _ = await execute_with_fallback(cmd, host=host)
    if returncode == 0:
        results[name] = stdout.strip()
return format_system_info(results)
```

For commands with placeholders (e.g., `{pid}`):

```python
group = COMMANDS["process_info"]
cmd = group.commands["ps_detail"]
args = [arg.format(pid=validated_pid) for arg in cmd.args]
returncode, stdout, _ = await execute_command(args, host=host)
```

## Summary by CodeRabbit

* **New Features**
  * Centralized command registry with argument substitution and fallback support
  * Standardized parsers and formatters to convert command output into human-readable reports

* **Refactoring**
  * Tools (network, processes, services, storage, system-info, logs) moved to registry-driven execution and unified parsing/formatting
  * Removed many local collection fallbacks in favor of centralized command flows

* **Documentation**
  * Development guideline: prefer Pydantic models

* **Tests**
  * Added extensive parser/formatter tests and updated tool tests for remote execution scenarios

### DIFF
--- a/tests/tools/test_storage.py
+++ b/tests/tools/test_storage.py
@@ -6,9 +6,7 @@ import sys
 from collections.abc import Callable
 from pathlib import Path
 from unittest.mock import AsyncMock
-from unittest.mock import MagicMock
 
-import psutil
 import pytest
 
 from mcp.server.fastmcp.exceptions import ToolError
@@ -113,17 +111,6 @@ def restricted_path(tmp_path):
 
 
 @pytest.fixture
-def mock_disk_io_stats():
-    """Fixture providing mock disk I/O statistics."""
-    mock_stats = MagicMock(spec=psutil._pslinux.sdiskio)
-    mock_stats.read_bytes = 1024 * 1024 * 1024  # 1 GB
-    mock_stats.write_bytes = 512 * 1024 * 1024  # 512 MB
-    mock_stats.read_count = 1000
-    mock_stats.write_count = 500
-    return {"sda": mock_stats}
-
-
-@pytest.fixture
 def mock_storage_execute_command(mock_execute_command_for):
     """Storage-specific execute_command mock using the shared factory."""
     return mock_execute_command_for("linux_mcp_server.tools.storage")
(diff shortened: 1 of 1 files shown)


src/linux_mcp_server/utils/types.py
```diff
@@ -1,8 +1,130 @@
...
+    tx_errors: int = 0
+    rx_dropped: int = 0
+    tx_dropped: int = 0
+
+
+# Process types
+class ProcessInfo(BaseModel):
+    """Parsed process information from ps output."""
+
+    pid: str
```
major, line 57: Would it make sense for some of these numeric ones to be int/float?

subpop: Yes, probably. This sets us up for MCP structured output. So a follow-up PR that snugs up the typing of these data structures is probably a good idea.

src/linux_mcp_server/commands.py
```diff
@@ -0,0 +1,196 @@
+"""Command registry for unified local and remote execution.
+
+This module provides a centralized registry of commands used by tools,
+enabling consistent execution across local and remote systems.
+"""
+
+from dataclasses import dataclass
+
+
+@dataclass
```
samdoran: Since this project uses Pydantic, use that in stead of `dataclasses`.

subpop: I thought I caught all those! Thanks for the catch.

subpop: Oh I just changed the data types in `types.py`, not the command specs. 🤦‍♂️

src/linux_mcp_server/commands.py
```diff
@@ -0,0 +1,196 @@
...
+    Args:
+        name: The command name in the registry.
+
+    Returns:
+        The CommandSpec or CommandGroup for the given name.
+
+    Raises:
+        KeyError: If the command name is not found in the registry.
+    """
+    return COMMANDS[name]
```
samdoran, line 148: Why does this function need to exist? `COMMANDS` is already a dict and so there is a `.get` method as well as keyword lookup.

src/linux_mcp_server/commands.py
```diff
@@ -0,0 +1,196 @@
...
+        **kwargs: Key-value pairs to substitute into placeholders.
+
+    Returns:
+        List of command arguments with placeholders replaced.
+
+    Example:
+        >>> substitute_command_args(["ps", "-p", "{pid}"], pid=1234)
+        ["ps", "-p", "1234"]
+    """
+    return [arg.format(**kwargs) for arg in args]
```
samdoran, line 165: Even if this is done in lots of places, I think it would be more obvious to just do the comprehension directly instead of calling a function. I could be wrong, but this doesn't seem worth it to make this call.

src/linux_mcp_server/commands.py
subpop: I questioned this too. I think I'll remove it. I don't think it provides any advantage.

src/linux_mcp_server/tools/storage.py
```diff
@@ -218,26 +118,52 @@ async def list_directories(
...
+    if not host:
+        path = _validate_path(path)
+
+    # Get the appropriate command for the order_by field
+    cmd_name = DIRECTORY_COMMANDS[order_by]
+    cmd = COMMANDS[cmd_name]
+    assert isinstance(cmd, CommandSpec)
+
+    # Substitute path into command args
+    args = substitute_command_args(cmd.args, path=path)
```
samdoran, line 133: Maybe handling the arg formatting could be handled by a `run()` method on the command object. The `run()` method could format the args then call `execute_command`.

src/linux_mcp_server/commands.py
subpop: Oh Cursor...


> Done! I removed the get_command function from commands.py. Since the function was never used anywhere in the codebase (only defined), no replacements were needed — callers should use COMMANDS["command_name"] directly. All tests, linters, and type checkers pass.

🫠

src/linux_mcp_server/commands.py
r0x0d: I always followed the principle that if you have just one line wrapped in a function, that's probably gonna cause more issues than help you. 

I agree with Sam on this one.

subpop: I have a vision for cleaning this up more, but I'd like to make it a followup PR.

src/linux_mcp_server/utils/types.py
subpop: #107

PR #103, opened 2025-12-12
refactor(tests): consolidate TestReadFile with fixtures

major:
- Add CONTENT_TEST_CASES for parameterized content tests (4 → 1)
- Add ERROR_TEST_CASES for parameterized error tests (2 → 1)
- Add REMOTE_CONTENT_CASES for parameterized remote tests (3 → 1)
- Use shared mock_storage_execute_command fixture
- Use verify_result_structure() helper for common assertions
- Split remote execution into focused tests (content, args, failure)
- Reduce 15 test methods to 11 (~27% reduction)
- Reduce ~50 lines while maintaining full test coverage

☝️ Depends on #101 to merge first

## Summary by CodeRabbit

* **Tests**
  * Expanded test coverage for storage tools, including parameterized test cases for sorting, filtering, and remote execution scenarios.
  * Added tests for edge cases such as empty directories, invalid parameters, and permission errors.
  * Enhanced test utilities to validate result structures across various content types and path resolution scenarios.

### DIFF
--- a/tests/tools/test_storage.py
+++ b/tests/tools/test_storage.py
@@ -821,48 +821,71 @@ class TestListFiles:
 
 
 class TestReadFile:
-    async def test_read_file_success(self, tmp_path):
-        """Test read_file with a valid file."""
-        # Create a test file
-        test_file = tmp_path / "test.txt"
-        test_content = "Hello, World!\nThis is a test file.\nLine 3."
-        test_file.write_text(test_content)
+    """Test suite for read_file tool."""
+
+    # Test data for content tests - (filename, content)
+    CONTENT_TEST_CASES = [
+        pytest.param(
+            "basic.txt",
+            "Hello, World!\nThis is a test file.\nLine 3.",
+            id="basic_multiline",
+        ),
+        pytest.param(
+            "empty.txt",
+            "",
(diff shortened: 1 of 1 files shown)


tests/tools/test_storage.py
```diff
@@ -288,208 +289,201 @@ async def test_list_block_devices_exception_handling(self, mock_storage_execute_
 
 @pytest.mark.skipif(sys.platform != "linux", reason="requires GNU version of coreutils/findutils")
 class TestListDirectories:
+    """Test suite for list_directories tool."""
+
+    # Test data for sorting tests - (order_by, sort, dir_specs, expected_entries)
+    SORT_TEST_CASES = [
```
r0x0d, line 295: I guess this is duplicated.. Can we move this to be outside the class and only add the cases that are related to the test itself?

major: That makes sense. This one is stuck behind two other PRs which is also waiting on #97 to get in, so I'll revisit this once the chain starts moving...

r0x0d: that was the only thing I spotted in this PR that I would like to have changed. Other than that, looks fine for me

PR #105, opened 2025-12-15
Add install tests for pip, uvx, and uv tool run

alexxa:
[RSPEED-2191](https://issues.redhat.com/browse/RSPEED-2191)

Verify pip and uvx package sanity, uv tool run and successful installation.

## Summary by CodeRabbit

* **Tests**
  * Added an automated "Installation Validation" workflow that runs installation checks across multiple Python versions (3.10–3.14) and install methods.
  * Introduced a QA verification script that exercises different install flows, ensures the server starts correctly, captures logs, and fails the job if startup is not detected.


scripts/verify_installation.sh
```diff
@@ -0,0 +1,61 @@
...
+    ;;
+
+  "uvx_run")
+    verify_server_startup "uvx linux-mcp-server" "uvx run"
+    ;;
+
+  "uv_tool_run")
+    echo "Starting uv tool run installation QA..."
+    uv tool install linux-mcp-server
+    verify_server_startup "uv tool run linux-mcp-server" "uv tool run"
```
samdoran, line 54: `uv tool run` and `uvx` are [equivalent](https://docs.astral.sh/uv/guides/tools/#running-tools). They both run the program in an ephemeral virtual environment.

This test should run `uv tool install linux-mcp-server` then verify the application runs successfully from `~/.local/bin`.

This test should be renamed since "uv tool run" means "run without installing" in `uv` parlance. Maybe `uv_tool_install` would work.

samdoran: I had that incorrect in the documentation. I only really understood the distinction after rereading the `uv` documentation more thoroughly. It's actually very confusing that `uv tool run` doesn't actually run the installed version of the tool.

samdoran (approved):

PR #110, opened 2025-12-15
Lower coverage targets in .codecov.yml

major:
Reduce patch/app test coverage requirements and we don't need to test our tests. 😆

## Summary by CodeRabbit

* **Chores**
  * Updated code coverage configuration: lowered patch and app coverage targets from 100% to 95%.
  * Made a minor formatting adjustment in the coverage configuration file.

### DIFF
--- a/.codecov.yml
+++ b/.codecov.yml
@@ -4,7 +4,7 @@ codecov:
     wait_for_ci: false
 
   require_ci_to_pass: false
-  token: 74bc381b-9c5f-44c1-9efa-af57d5f3fe50  # notsecret  # repo-scoped, upload-only, stability in fork PRs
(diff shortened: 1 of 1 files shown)


.codecov.yml
```diff
@@ -18,21 +18,13 @@ coverage:
         threshold: 10%
         paths:
           - src/
-      tests:
-        target: 100%
-        paths:
-          - tests/
```
samdoran, line 24: Test coverage should remain at 100%. If we have tests that aren't being run, they should be removed.

subpop: This seems like a reasonable usage of 100% expectation.

samdoran: It's basically an indicator of dead test code. It's far less onerous than requiring coverage on code under test.

major: Fixed.

PR #116, opened 2025-12-15
refactor(commands): add type-safe registry lookup helpers

major:
Add get_command_spec() and get_command_group() functions to centralize type validation when retrieving commands from the COMMANDS registry. This eliminates repeated 2-3 line isinstance checks across all tool modules.

- Add get_command_spec() for CommandSpec lookups
- Add get_command_group() for CommandGroup lookups
- Update all tool modules to use the new helpers
- Remove direct COMMANDS access and manual isinstance checks

### DIFF
--- a/src/linux_mcp_server/commands.py
+++ b/src/linux_mcp_server/commands.py
@@ -148,6 +148,42 @@ def get_command(name: str) -> CommandEntry:
     return COMMANDS[name]
 
 
+def get_command_spec(key: str) -> CommandSpec:
+    """Get and validate a CommandSpec from the COMMANDS registry.
+
+    Args:
+        key: The command key to look up.
+
+    Returns:
+        The validated CommandSpec.
+
+    Raises:
+        TypeError: If the retrieved command is not a CommandSpec.
+    """
+    cmd = COMMANDS[key]
+    if not isinstance(cmd, CommandSpec):
+        raise TypeError(f"Expected CommandSpec for '{key}', got {type(cmd).__name__}")
+    return cmd
+
+
+def get_command_group(key: str) -> CommandGroup:
+    """Get and validate a CommandGroup from the COMMANDS registry.
+
+    Args:
+        key: The command group key to look up.
+
+    Returns:
+        The validated CommandGroup.
+
+    Raises:
+        TypeError: If the retrieved command is not a CommandGroup.
+    """
+    group = COMMANDS[key]
+    if not isinstance(group, CommandGroup):
+        raise TypeError(f"Expected CommandGroup for '{key}', got {type(group).__name__}")
+    return group
+
+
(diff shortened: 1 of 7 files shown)


src/linux_mcp_server/commands.py
```diff
@@ -148,6 +148,42 @@ def get_command(name: str) -> CommandEntry:
     return COMMANDS[name]
 
 
+def get_command_spec(key: str) -> CommandSpec:
```
samdoran, line 151: This functionality can be removed and all the calls to it replaced by the using the [validate_call decorator](https://docs.pydantic.dev/latest/concepts/validation_decorator/). This lets Pydantic handle input and optionally return validation and coercion.

src/linux_mcp_server/commands.py
```diff
@@ -148,6 +148,42 @@ def get_command(name: str) -> CommandEntry:
...
+    Raises:
+        TypeError: If the retrieved command is not a CommandSpec.
+    """
+    cmd = COMMANDS[key]
+    if not isinstance(cmd, CommandSpec):
+        raise TypeError(f"Expected CommandSpec for '{key}', got {type(cmd).__name__}")
+    return cmd
+
+
+def get_command_group(key: str) -> CommandGroup:
```
samdoran, line 169: Same thing here: let's leverage Pydantic.

src/linux_mcp_server/commands.py
major: Oh shoot, I didn't know that decorator existed.

src/linux_mcp_server/tools/logs.py
samdoran, line 87: Here is an example of how this would look.
```suggestion
@validate_call
async def get_audit_logs(
lines: t.Annotated[int, Field(description="Number of log lines to retrieve.")] = 100,
host: Host | None = None,
command: CommandSpec = COMMANDS["audit_logs"]
) -> str:
```

This also makes the code easier to test because we can pass in a `CommandSpec` object with the desired attributes.

src/linux_mcp_server/commands.py
samdoran: I didn't either. I just went looking for a way to leverage Pydantic's validation machinery on plain old functions and I found that in docs. It's really cool!

src/linux_mcp_server/tools/logs.py
major: Some of the functions dynamically choose commands, so this could get dicey, but I see your point.

samdoran: It looks like `COMMANDS` contains multiple types. We could use the `CommandGroup` type alias, make separate dicts for each type, or as @subpop suggested just make all of them a `CommandGroup` but some have only one item. That last suggestion is probably the best idea since it makes the objects more consistent.

PR #118, opened 2025-12-16
Use get_command instead of accessing global dictionary

r0x0d:
All tools were accessing the global dictionary of `COMMANDS` directly, and some of them had condition handling in case the specified command was not in the proper state, but many of them didn't. This patch moves the conditional handling to be inside `get_command` and refactor the tools to use it.

## Summary by CodeRabbit

* **Refactor**
  * Unified command resolution across logging, network, processes, services, storage, and system-info tools to use a centralized registry accessor with consistent validation and error handling; no public APIs or function signatures changed.
* **Tests**
  * Added tests validating the centralized lookup behavior and its TypeError/error conditions.

### DIFF
--- a/src/linux_mcp_server/tools/system_info.py
+++ b/src/linux_mcp_server/tools/system_info.py
@@ -1,11 +1,13 @@
 """System information tools."""
 
+import typing as t
+
 from mcp.types import ToolAnnotations
 
 from linux_mcp_server.audit import log_tool_call
 from linux_mcp_server.commands import CommandGroup
-from linux_mcp_server.commands import COMMANDS
 from linux_mcp_server.commands import CommandSpec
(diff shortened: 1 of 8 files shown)


src/linux_mcp_server/tools/system_info.py
```diff
@@ -36,10 +38,7 @@ async def get_system_information(
     """
     try:
         # Get command group from registry
-        group = COMMANDS["system_info"]
-        if not isinstance(group, CommandGroup):
-            raise TypeError(f"Expected CommandGroup for 'system_info', got {type(group).__name__}")
-
+        group = t.cast(CommandGroup, get_command("system_info"))
```
subpop, line 41: Does `t.cast` raise an exception if the typecast fails?

PR #119, opened 2025-12-16
refactor(commands): unify command registry to use CommandGroup consistently

major:
- Wrap all commands in CommandGroup for consistency, eliminating mixed CommandSpec/CommandGroup types and isinstance() checks in consuming code
- Add get_command_group() helper for tools that iterate over subcommands
- Improve get_command() error messages to list available commands/subcommands
- Update system_info.py to use get_command_group() instead of direct access
- Add tests for get_command() and get_command_group() error cases

## Summary by CodeRabbit

* **Refactor**
  * Converted command registry to a group-and-subcommand model with centralized retrieval; tools now use the unified accessors and legacy registry checks were removed.
  * Improved error messages when groups or subcommands are missing.

* **Bug Fixes**
  * Ensured remote/host context is propagated for network-related command execution.

* **Tests**
  * Added coverage for group and subcommand retrieval (success and error paths).
  * Adjusted storage tests to always create test content files, stabilizing test scaffolding.

### DIFF
--- a/tests/tools/test_storage.py
+++ b/tests/tools/test_storage.py
@@ -44,9 +44,8 @@ def setup_test_directory(tmp_path) -> Callable[[list[tuple[str, int, float]]], t
             dir_path.mkdir()
 
             # Create a file inside the directory to give it size
-            if size > 0:
-                content_file = dir_path / "content.txt"
-                content_file.write_text("x" * size)
+            content_file = dir_path / "content.txt"
+            content_file.write_text("x" * size)
 
             # Set modification time on the directory itself
             os.utime(dir_path, (modified_time, modified_time))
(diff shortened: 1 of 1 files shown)


src/linux_mcp_server/tools/network.py
```diff
@@ -33,16 +31,11 @@ async def get_network_interfaces(
...
-        group = COMMANDS["network_interfaces"]
-        if not isinstance(group, CommandGroup):
-            raise TypeError(f"Expected CommandGroup for 'network_interfaces', got {type(group).__name__}")
-
         interfaces = {}
         stats = {}
 
         # Get brief interface info
-        brief_cmd = group.commands["brief"]
+        brief_cmd = get_command("network_interfaces", "brief")
```
samdoran, line 38: The `get_command` and `get_command_groups` are not adding much benefit and can be removed.

Tool functions would be simpler and easier to test if the command is provided as a default parameter value. Any errors in the default values would show up immediately on application launch instead of later when tools are called.

```python
@mcp.tool()
@log_tool_call
@disallow_local_execution_in_containers
async def get_network_interfaces(
host: Host | None = None,
cmd: CommandSpec = COMMAND["network_interfaces"].brief
) -> str:
```

With FastMCP 2.14, we can use dependency injection to exclude the parameter from the tool schema so the LLM won't be tempted to provide any arguments.

PR #120, opened 2025-12-16
build(deps): add pytest-randomly to test dependencies

major:
Randomizes test execution order to help detect hidden test interdependencies that could cause flaky tests.

## Summary by CodeRabbit

* **Bug Fixes**
  * Fixed an issue where provided command lists could be unexpectedly modified during execution, ensuring callers’ inputs remain unchanged.

* **Chores**
  * Enabled randomized test ordering to help surface flaky tests and improve overall test robustness; no runtime behavior changes.

### DIFF
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -201,6 +201,8 @@ class SSHConnectionManager:
         conn = await self.get_connection(host)
         bin = command[0]
         if not Path(bin).is_absolute():
+            # NOTE(major): Copy to avoid mutating the caller's list (e.g., CommandSpec.args)
+            command = command.copy()
             command[0] = await get_remote_bin_path(bin, host, conn)
 
         # Build command string with proper shell escaping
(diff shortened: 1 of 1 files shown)


r0x0d (approved):
great addition!

src/linux_mcp_server/connection/ssh.py
```diff
@@ -201,6 +201,8 @@ async def execute_remote(
         conn = await self.get_connection(host)
         bin = command[0]
```
subpop, line 202: So are assignments like this in Python effectively pointer references? This creates a variable `bin` that is a string pointing to the same address space as `command[0]`? And are the parameters also references? So passing `command` in as a function parameter is like passing in a mutable reference pointer?

src/linux_mcp_server/connection/ssh.py
```diff
@@ -201,6 +201,8 @@ async def execute_remote(
...
         if not Path(bin).is_absolute():
+            # NOTE(major): Copy to avoid mutating the caller's list (e.g., CommandSpec.args)
+            command = command.copy()
```
subpop, line 205: Presumably this creates a copy of the object `command`, but given Python's a-function-is-a-block-scope behavior, does this assign to a *new* variable named `command`, shadowing the function argument? Or does it assign a copy of `command` to `command`, changing the address to which `command` points? Does that change the global reference of `command` (if Python arguments are all reference pointers)?

src/linux_mcp_server/connection/ssh.py
major: This has confused me a bit, too, in the past, but the top few comments here explain it pretty well: https://www.reddit.com/r/learnpython/comments/t2kp6j/how_is_python_pass_by_reference_different_from/

PR #127, opened 2025-12-17
Make COMMANDS less easily mutable

subpop:
Add some belts and suspenders to make COMMANDS less mutable.

- Use frozen Pydantic models
- Use tuples instead of lists
- Use Sequence where applicable
- Use Mapping to provide a public read-only interface to COMMANDS
- 

## Summary by CodeRabbit

* **Refactor**
  * Command system made immutable and more strongly typed to improve stability and prevent accidental mutation
  * Public command registry now exposed as a read-only mapping
  * Interfaces broadened to accept generic sequence types while preserving behavior

* **Tests**
  * Test expectations updated to reflect immutable sequence types used by the command APIs


src/linux_mcp_server/commands.py
```diff
@@ -26,183 +33,187 @@ class CommandGroup(BaseModel):
...
-            "lspci": CommandSpec(args=["lspci"]),
-            "lsusb": CommandSpec(args=["lsusb"]),
+            "lscpu": CommandSpec(args=("lscpu",)),
+            "lspci": CommandSpec(args=("lspci",)),
+            "lsusb": CommandSpec(args=("lsusb",)),
         }
     ),
 }
 
+COMMANDS: Mapping[str, CommandGroup] = MappingProxyType(_COMMANDS)
```
samdoran: This could be done to the value in `_COMMANDS` directly removing the need for this extra variable.
```python
COMMANDS: Mapping[str, CommandGroup] = MappingProxyType({
    # all the things
})
```

tests/test_commands.py
```diff
@@ -14,33 +14,33 @@ class TestSubstituteCommandArgs:
         ("args", "kwargs", "expected"),
         [
             pytest.param(
-                ["ps", "-p", "{pid}"],
+                ("ps", "-p", "{pid}"),
```
samdoran: It might be good to leave some inputs as lists to verify that a sequence type other than tuple does work.

samdoran (commented):
This is a nice improvement though there are still some holes as there always will be in Python. Objects can only be made faux-immutable at best.

Here is one of several examples where data can still be mutated.

```
In [23]: COMMANDS['system_info'].commands['hostname'] = 'wat'

In [24]: COMMANDS['system_info'].commands
Out[24]:
{'hostname': 'wat',
 'os_release': CommandSpec(args=('cat', '/etc/os-release'), fallback=None),
 'kernel': CommandSpec(args=('uname', '-r'), fallback=None),
 'arch': CommandSpec(args=('uname', '-m'), fallback=None),
 'uptime': CommandSpec(args=('uptime', '-p'), fallback=None),
 'boot_time': CommandSpec(args=('uptime', '-s'), fallback=None)}
```

The original goal was to prevent elements of `CommandSpec.args` from being mutated and this change accomplishes that.

src/linux_mcp_server/commands.py
subpop: Good point!

samdoran (approved):

PR #128, opened 2025-12-17
[RSPEED-2278] Extract common logic into _list_resources helper

major:
Extract duplicated code from list_directories() and list_files() into a shared _list_resources() helper function. This consolidates path validation, command execution, parsing, sorting, and formatting into a single well-documented function.

Also corrects the top_n field descriptions which incorrectly stated the parameter was "only used with size ordering" when it actually works with all sort orders.

## Summary by CodeRabbit

* **Refactor**
  * Centralized file and directory listing logic to unify validation, command execution, parsing, sorting, limiting and error handling.

* **Documentation**
  * Clarified parameter guidance for listing operations, including updated behavior for top‑N and ordering options.

* **Tests**
  * Tightened log-output test filtering to exclude header/marker lines and strengthened the expected line-count assertion.

### DIFF
--- a/tests/tools/test_services.py
+++ b/tests/tools/test_services.py
@@ -50,9 +50,14 @@ class TestServices:
 
     async def test_get_service_logs(self, mcp_client):
         result = await mcp_client.call_tool("get_service_logs", arguments={"service_name": "sshd.service", "lines": 5})
-        result_lines = [line for line in result.content[0].text.split("\n") if line and not line.startswith("=")]
+        # Filter out empty lines, header lines (=), and journalctl boot markers (--)
+        result_lines = [
+            line
+            for line in result.content[0].text.split("\n")
+            if line and not line.startswith("=") and not line.startswith("--")
+        ]
 
-        assert len(result_lines) < 6, "Got more lines than expected"
+        assert len(result_lines) <= 5, "Got more lines than expected"
 
     async def test_get_service_logs_with_nonexistent_service(self, mcp_client):
         result = await mcp_client.call_tool(


src/linux_mcp_server/tools/storage.py
```diff
@@ -60,6 +61,65 @@ def _validate_path(path: str) -> str:
...
+        entries = parser(stdout, order_by)
+
+        if top_n:
+            if order_by == OrderBy.SIZE:
+                entries = sorted(entries, key=lambda e: e.size, reverse=sort == SortBy.DESCENDING)
+            elif order_by == OrderBy.MODIFIED:
+                entries = sorted(entries, key=lambda e: e.modified, reverse=sort == SortBy.DESCENDING)
+            else:
+                entries = sorted(entries, key=lambda e: e.name.lower(), reverse=sort == SortBy.DESCENDING)
+            entries = entries[:top_n]
```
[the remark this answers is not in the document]
major: Oh nice catch. Missed that! Good bot. 👏

src/linux_mcp_server/tools/storage.py
```diff
@@ -60,6 +61,68 @@ def _validate_path(path: str) -> str:
     return Path(path).as_posix()
 
 
+async def _list_resources(
+    path: str,
+    order_by: OrderBy,
+    sort: SortBy,
+    top_n: int | None,
+    host: Host | None,
+    command_map: dict[OrderBy, str],
```
subpop: Should this be a `Mapping` or some other immutable-y type?

major: Good idea!

src/linux_mcp_server/tools/storage.py
```diff
@@ -104,47 +169,39 @@ async def list_directories(
...
+        path: Absolute path to the directory to analyze.
+        order_by: Sort criterion (size, name, or modified).
+        sort: Sort direction (ascending or descending).
+        top_n: Optional limit on number of directories to return.
+        host: Optional remote host to execute command on.
 
-    try:
-        returncode, stdout, stderr = await cmd.run(host=host, path=path)
+    Returns:
+        Formatted string listing of directories with metadata.
```
subpop: I thought we were dropping the docstring annotations in favor of the Pydantic annotations. Is there a reason you've added docstring annotations back here?

major: My intent here was to just be verbose. I can remove these if they're bothersome.

src/linux_mcp_server/tools/storage.py
```diff
@@ -49,6 +51,69 @@ class SortBy(StrEnum):
...
+    """
+    cmd_name = command_map[order_by]
+    cmd = get_command(cmd_name)
+
+    try:
+        returncode, stdout, stderr = await cmd.run(host=host, path=path)
+
+        # The du command will exit with code 1 even if it gets some valid results.
+        # Only error in the case where we got non-zero exit code and no data in stdout.
+        if returncode != 0 and not stdout:
```
subpop, line 92: We have a utility function that does this check now: `is_successful_output`. It's a one-liner: `return returncode == 0 and not is_empty_output(stdout)`. I'm not sure if we should move towards or away from tiny utility functions like that.

src/linux_mcp_server/tools/storage.py
```diff
@@ -49,6 +51,69 @@ class SortBy(StrEnum):
...
+            entries = sorted(entries, key=lambda e: e.name.lower(), reverse=sort == SortBy.DESCENDING)
+
+        # Apply limit if specified
+        if top_n:
+            entries = entries[:top_n]
+
+        return formatter(entries, path, order_by, sort == SortBy.DESCENDING)
+
+    except ToolError:
+        raise
```
subpop: I think I recall from my adventures in converting tools to structured output, this pattern ends up being unnecessary. I don't remember why though.

subpop (commented):
I noticed a couple things that stood out to me from my recent work in the tool code. Whether it's worth addressing is a matter of opinion though; the tool will work either way.

src/linux_mcp_server/tools/storage.py
samdoran: In this case, I couldn't use that helper function which is why I put the comment in there. It was using the helper function before, which caused the tool to not function.

src/linux_mcp_server/tools/storage.py
samdoran: The framework catches `Exception` and raises as `ToolError`.

major: Makes sense, I'll pull it.

src/linux_mcp_server/tools/storage.py
major: I'll likely just leave it there since this is a funky special case.

PR #130, opened 2025-12-18
Add container health verification and MCP protocol audit tasks

alexxa:
to complete [RSPEED-2191](https://issues.redhat.com/browse/RSPEED-2191)

### DIFF
--- a/.tekton/pipeline-build-multiarch.yaml
+++ b/.tekton/pipeline-build-multiarch.yaml
@@ -226,6 +226,28 @@ spec:
       taskRef:
         name: get-version
 
+    - name: build-qa-toolbox
+      params:
+        - name: IMAGE
+          # We use the same output-image param but append a suffix
+          value: $(params.output-image)-test-suite
+        - name: DOCKERFILE
+          value: Containerfile.qa
+        - name: CONTEXT
+          value: .
+        - name: SOURCE_ARTIFACT
+          value: $(tasks.clone-repository.results.SOURCE_ARTIFACT)
+      runAfter:
+        - clone-repository
+      taskRef:
+        resolver: bundles
+        params:
+          - name: name
+            value: buildah-remote-oci-ta
+          - name: bundle
+            # Using the same bundle version you use for build-images 
+            value: quay.io/konflux-ci/tekton-catalog/task-buildah-remote-oci-ta:0.7@sha256:ee5e01eb59a3f70bb1012950fbc4081bac96d3f3517e6d204314484cd2e0059b
+
     - name: build-images
       matrix:
         params:
--- a/.tekton/task-mcp-protocol-audit.yaml
+++ b/.tekton/task-mcp-protocol-audit.yaml
@@ -7,16 +7,28 @@ spec:
   params:
     - name: image_url
       type: string
-      description: The URL of the image to audit.
+      [cite_start]description: The URL of the image to audit. [cite: 4]
+    - name: qa_image_url
+      default: "quay.io/redhat-user-workloads/rhel-lightspeed-tenant/linux-mcp-server-test-suite:latest"
+
+  sidecars:
+    - name: mcp-server
+      [cite_start]image: $(params.image_url) [cite: 1]
+      command: ["/bin/sh", "-c"]
+      args:
+        - |
+          microdnf install -y socat
+          socat TCP-LISTEN:8080,reuseaddr,fork EXEC:linux-mcp-server,stderr
+
   steps:
     - name: validate-tool-discovery
-      image: $(params.image_url)
-      workingDir: $(workspaces.source.path)
+      image: $(params.qa_image_url)
       script: |
-        ./scripts/mcp_protocol_audit.sh
+        #!/usr/bin/env bash
+        /usrl/local/bin/scripts/mcp_protocol_audit.sh [cite: 4]
 
     - name: verify-stdout-hygiene
-      image: $(params.image_url)
-      workingDir: $(workspaces.source.path)
+      image: $(params.qa_image_url)
       script: |
-        ./scripts/verify_stdout_hygiene.sh
+        #!/usr/bin/env bash
+        /usr/local/bin/scripts/verify_stdout_hygiene.sh [cite: 4]
--- a/.tekton/task-verify-container.yaml
+++ b/.tekton/task-verify-container.yaml
@@ -2,27 +2,32 @@ apiVersion: tekton.dev/v1
 kind: Task
 metadata:
   name: verify-container-lifecycle
-  namespace: rhel-lightspeed-tenant
 spec:
-  description: Verifies that the MCP server container starts, has correct user permissions, and responds to JSON-RPC.
+  [cite_start]description: Verifies MCP server startup, permissions, and handshake via sidecar. [cite: 4]
   params:
     - name: image_url
-      description: The URL of the image to verify.
       type: string
+    - name: qa_image_url
+      default: "quay.io/redhat-user-workloads/rhel-lightspeed-tenant/linux-mcp-server-test-suite:latest"
 
-    - name: startup_timeout
-      description: Time to wait for server statup
-      type: string
-      default: "30s"
+  sidecars:
+    - name: server
+      image: $(params.image_url)
+      command: ["/bin/sh", "-c"]
+      args:
+        - |
+          socat TCP-LISTEN:8080,reuseaddr,fork EXEC:linux-mcp-server,stderr 
+
   steps:
     - name: verify-user-and-startup
-      image: $(params.image_url)
-      workingDir: $(workspaces.source.path)
+      image: $(params.qa_image_url)
       script: |
-        ./scripts/verify_container_lifecycle.sh
+        #!/usr/bin/env bash
+        # The QA container probes the sidecar's network port
+        /usr/local/bin/scripts/verify_container_lifecycle.sh
 
     - name: verify-mcp-handshake
-      image: $(params.image_url)
-      workingDir: $(workspaces.source.path)
+      image: $(params.qa_image_url)
       script: |
-        ./scripts/verify_mcp_handshake.sh
+        #!/usr/bin/env bash
+        /usr/local/bin/scripts/verify_mcp_handshake.sh
--- a/Containerfile.qa
+++ b/Containerfile.qa
@@ -1,0 +1,11 @@
+FROM registry.access.redhat.com/ubi10/ubi-minimal:latest
+
+RUN microdnf -y install nc jq bash socat openssh && microdnf clean all
+
+RUN mkdir -p /usr/local/bin/scripts
+
+# Copy all scripts into the test image
+COPY scripts/ /usr/local/bin/scripts/
+RUN chmod +x /usr/local/bin/scripts/*.sh
+
+ENTRYPOINT ["/bin/bash"]
--- a/.tekton/integration-lifecycle.yaml
+++ b/.tekton/integration-lifecycle.yaml
@@ -13,3 +13,6 @@ spec:
         value: main
       - name: pathInRepo
         value: .tekton/task-verify-container.yaml
+  params:
+    - name: qa_image_url
+      value: "quay.io/redhat-user-workloads/rhel-lightspeed-tenant/linux-mcp-server-test-suite:latest"
--- a/.tekton/integration-protocol.yaml
+++ b/.tekton/integration-protocol.yaml
@@ -13,3 +13,6 @@ spec:
         value: main
       - name: pathInRepo
         value: .tekton/task-mcp-protocol-audit.yaml
+  params:
+    - name: qa_image_url
+      value: "quay.io/redhat-user-workloads/rhel-lightspeed-tenant/linux-mcp-server-test-suite:latest"
--- a/scripts/mcp_protocol_audit.sh
+++ b/scripts/mcp_protocol_audit.sh
@@ -3,7 +3,8 @@ set -euo pipefail
 
 echo "--- Verification: Tool Discovery & Schema ---"
 LIST_REQ='{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}'
-RESPONSE=$(echo "$LIST_REQ" | linux-mcp-server 2>/dev/null)
+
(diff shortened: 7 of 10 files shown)


scripts/verify_container_integration.sh
```diff
@@ -0,0 +1,46 @@
...
+
+  echo "Verifying Container Wrapper server startup..."
+  rm -f "$log_file"
+
+  # Start the wrapper script
+  eval "$start_command" &> "$log_file" &
+  SERVER_PID=$!
+
+  # Wait for the specific MCP startup message
+  if timeout 15s grep -q "Running Linux MCP Server" <(tail -f "$log_file"); then
```
narmaku: We could make this timeout configurable via a variable in the beginning of the script.
Also I would add 30s as default just to be safe (unless we have a clear performance requirement of spawning the MCP server in less than x seconds).

scripts/verify_installation.sh
```diff
@@ -20,18 +20,17 @@ verify_server_startup() {
   SERVER_PID=$!
 
   # Wait for log message
-  if timeout 10s grep -q "Running Linux MCP Server" <(tail -f "$log_file"); then
+  # Note: The server logs "Running Linux MCP Server" upon successful init
+  if timeout 15s grep -q "Running Linux MCP Server" <(tail -f "$log_file"); then
     echo "$method_name server successfully initialized and logged startup message."
-    kill $SERVER_PID
-    exit 0
+    kill $SERVER_PID 2>/dev/null || true
```
narmaku, line 26: Could we redirect the errors to a temporary file, and then output its content in case something failed?
Same in line :33.

.tekton/task-mcp-protocol-audit.yaml
```diff
@@ -0,0 +1,55 @@
...
+          echo "❌ Hygiene: Stdout is polluted with non-JSON text: $FIRST_LINE"
+          exit 1
+        fi
+        
+        # 2. Ensure it's a valid JSON-RPC response object
+        if [[ ! "$FIRST_LINE" == *"jsonrpc"* ]]; then
+          echo "❌ Hygiene: Valid JSON found, but it's not a JSON-RPC 2.0 message."
+          exit 1
+        fi
+        echo "✅ Hygiene: Stdout is pure JSON-RPC."
```
narmaku: Not sure if this check is deep enough to verify that the JSON retrieves is "purely" in JSON-RPC.
We may need to add a tool that parses the whole JSON and verifies its structure. But I believe that can be added later. This check is more than enough for our dev-preview!

.tekton/task-mcp-protocol-audit.yaml
```diff
@@ -0,0 +1,55 @@
...
+      image: $(params.image_url)
+      script: |
+        #!/usr/bin/env bash
+        set -euo pipefail
+
+        # 1. Simulate the MCP 'tools/list' request
+        LIST_REQ='{"jsonrpc":"2.0","id":1,"method":"tools/list","params":{}}'
+        
+        # 2. Capture response, ignoring logs on stderr
+        RESPONSE=$(echo "$LIST_REQ" | linux-mcp-server 2>/dev/null)
```
narmaku: If would be great to make sure that some of the existing tools are actually listed there.

.tekton/task-verify-container.yaml
```diff
@@ -0,0 +1,57 @@
...
+          exit 1
+        fi
+
+        echo "--- Verification Step 2: Server Startup Log ---"
+        # Start server in background, capture logs
+        # Uses the CMD defined in Containerfile
+        linux-mcp-server > startup.log 2>&1 &
+        SERVER_PID=$!
+
+        timeout 15s bash -c 'until grep -q "Running Linux MCP Server" startup.log; do sleep 1; done' || {
```
narmaku: Can we add the timeout as an env var? So we can easily change it for debugging, etc.

.tekton/task-verify-container.yaml
```diff
@@ -0,0 +1,57 @@
...
+    - name: verify-user-and-startup
+      image: $(params.image_url)
+      script: |
+        #!/usr/bin/env bash
+        set -euo pipefail
+
+        echo "--- Verification Step 1: User Identity ---"
+        CURRENT_UID=$(id -u)
+        echo "Running as UID: $CURRENT_UID"
+        if [ "$CURRENT_UID" -ne 1001 ]; then
```
narmaku: Not sure if we are only compatible with this UID. I believe the end user can use whatever UID they want, as long as it exists in the local system where the container will be used. Maybe @samdoran can clarify this to us.

scripts/linux-mcp-container.sh
```diff
@@ -0,0 +1,52 @@
...
+SSH_KEY="${LINUX_MCP_SSH_KEY_PATH:-$HOME/.ssh/id_rsa}"
+IMAGE="${MCP_CONTAINER_IMAGE:-quay.io/redhat-services-prod/rhel-lightspeed-tenant/linux-mcp-server:latest}"
+LOGS_DIR="$HOME/.local/share/linux-mcp-server/logs"
+
+# Ensure logs directory exists
+mkdir -p "$LOGS_DIR"
+
+# Build podman command
+PODMAN_ARGS=(
+    run --rm -i
```
narmaku, line 39: Let's add `--pull=always` just in case we use this locally. Martin had issues with this as it was using an old container and had to manually pull first.

alexxa: this file is pulled in as your commit in https://github.com/rhel-lightspeed/linux-mcp-server/pull/104/files
if `--pull-always` is desired, I would add it there, and then I can repull your commit. 

However, it may be a double-edged sword as latest security patches and features of the MCP server are trade off for network latency to every single execution of the MCP server. AFAIK, some host agents like, e.g. Claude desktop start/stop MCP servers, which can lead in 3-5 seconds delay every time the tool is initialized. Correct me if I'm wrong. 

what about adding `--pull=newer` instead, then speed and container freshness are balanced?

.tekton/task-verify-container.yaml
alexxa: our docs and container file say it's 1001

see https://github.com/search?q=repo%3Arhel-lightspeed%2Flinux-mcp-server%201001&type=code

scripts/verify_installation.sh
alexxa: the script already redirects the server output to server_output.log, line # 20. In line # 26 - success, we just cleanup, in line # 33 - failure, we already print the log file cat "$log_file" on line # 30

scripts/verify_container_integration.sh
alexxa: updated to variable and 30s

.tekton/task-mcp-protocol-audit.yaml
alexxa: good point, replaced grep with jq

.tekton/task-mcp-protocol-audit.yaml
alexxa: certainly, if one shares the expected list... what about adding it later?

.tekton/task-verify-container.yaml
alexxa: updated to a variable and 30s

scripts/linux-mcp-container.sh
narmaku: Awesome, yes! Let's use `newer`.

.tekton/task-mcp-protocol-audit.yaml
narmaku: Agree. Let's add this later.

.tekton/task-verify-container.yaml
```diff
@@ -0,0 +1,62 @@
...
+        SERVER_PID=$!
+
+        timeout $(params.startup_timeout) bash -c 'until grep -q "Running Linux MCP Server" startup.log; do sleep 1; done' || {
+          echo "::error::Server failed to log 'Running Linux MCP Server' within $(params.startup_timeout)"
+          cat startup.log
+          kill $SERVER_PID || true
+          exit 1
+        }
+        echo "Success: Server initialized correctly."
+        kill $SERVER_PID
```
subpop: I don't like this idea of embedding an executable script inside a YAML file. It hides execution functionality inside a very undiscoverable location. If we have to have all these various scripts that get run as part of the tekton build process, it's probably best we keep them outside the YAML definitions and keep them in `scripts`. Or if they're highly specialized to the tekton infrastructure, then can we store them along side the YAML in this `.tekton` directory?

.github/workflows/ci_installation.yaml
```diff
@@ -62,3 +64,13 @@ jobs:
...
           ./scripts/verify_installation.sh
+
+  container_integration_qa:
+    name: Container Integration - Linux
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6.0.1
+      - name: Run Container Integration Test
+        run: |
+          chmod +x scripts/verify_container_integration.sh
```
samdoran: Can the script be made executable in version control in order to avoid the need to set the execute bit every time?

.tekton/pipeline-build-multiarch.yaml
```diff
@@ -346,6 +346,32 @@ spec:
           values:
             - "true"
 
+    - name: verify-container
+      params:
+        - name: image_url
+          value: $(tasks.build-image-index.results.IMAGE_URL)
```
samdoran: This should not be part of the build process but should be an integration test in Konflux.

.tekton/pipeline-build-multiarch.yaml
```diff
@@ -346,6 +346,32 @@ spec:
...
+      taskRef:
+        name: verify-container-lifecycle
+
+      when:
+        - input: $(tasks.init.results.build)
+          operator: in
+          values:
+            - "true"
+
+    - name: mcp-protocol-audit
```
samdoran: Same here. This is a test, not really part of the build process.

.tekton/task-mcp-protocol-audit.yaml
```diff
@@ -0,0 +1,63 @@
...
+spec:
+  description: Audits the MCP server for protocol compliance, tool discovery, and stdout hygiene.
+  params:
+    - name: image_url
+      type: string
+      description: The URL of the image to audit.
+  steps:
+    - name: validate-tool-discovery
+      image: $(params.image_url)
+      script: |
```
samdoran, line 26: Rather than having large inline scripts inside of YAML that are hard to read and maintain, these scripts should be in a CI container image that this task references.

.tekton/task-mcp-protocol-audit.yaml
```diff
@@ -0,0 +1,63 @@
...
+          echo "::error::Protocol Audit Failed: Response within $TIMEOUT was structurally invalid or missing tools."
+          echo "Response received: $RESPONSE"
+          exit 1
+        }
+
+        echo "✅ Tool Discovery: 'inspect_linux' found with valid inputSchema."
+
+    - name: verify-stdout-hygiene
+      image: $(params.image_url)
+      script: |
```
samdoran, line 32: Same thing here. This script should be in a container image that this task references.

.tekton/task-verify-container.yaml
samdoran: I agree wholeheartedly with this. These scripts should be in a container image that this task references. [Here is an example](https://github.com/project-koku/koku-test-container).

.tekton/pipeline-build-multiarch.yaml
```diff
@@ -226,6 +226,28 @@ spec:
       taskRef:
         name: get-version
 
+    - name: build-qa-toolbox
```
samdoran, line 229: This modifies the build pipeline for this project to build two container images. This image needs to be built outside of this repository so they can be maintained and updated independently.

.tekton/task-mcp-protocol-audit.yaml
```diff
@@ -0,0 +1,34 @@
...
+        - |
+          microdnf install -y socat
+          socat TCP-LISTEN:8080,reuseaddr,fork EXEC:linux-mcp-server,stderr
+
+  steps:
+    - name: validate-tool-discovery
+      image: $(params.qa_image_url)
+      script: |
+        #!/usr/bin/env bash
+        /usrl/local/bin/scripts/mcp_protocol_audit.sh [cite: 4]
```
samdoran, line 28:
```suggestion
        /usr/local/bin/scripts/mcp_protocol_audit.sh [cite: 4]
```

.tekton/task-mcp-protocol-audit.yaml
```diff
@@ -0,0 +1,34 @@
+apiVersion: tekton.dev/v1
+kind: Task
```
samdoran, line 2: These tasks should be in the integration test container repo and referenced by a [custom integration test](https://konflux.pages.redhat.com/docs/users/testing/integration/creating.html) that needs to be created in Konflux.

Containerfile.qa
samdoran, line 1: This container file should be in the repo for building the CI image

.github/workflows/ci_installation.yaml
```diff
@@ -60,5 +62,13 @@ jobs:
...
+          bash ./scripts/verify_installation.sh
+
+  container_integration_qa:
+    name: Container Integration - Linux
+    runs-on: ubuntu-latest
+    steps:
+      - uses: actions/checkout@v6.0.1
+      - name: Run Container Integration Test
+        run: |
+          bash ./scripts/verify_container_integration.sh
```
samdoran, line 74: Looking at what this does more closely, it will mostly be testing an incorrect container image since the test script always pulls `quay.io/redhat-services-prod/rhel-lightspeed-tenant/linux-mcp-server:latest`, I don't see where `MCP_CONTAINER_IMAGE` is being overridden in the task parameters, and opening PRs to this repo does not trigger container image builds.

I don't think it makes sense to have any CI tests related to the container image here since the container image is built by Konflux.

Container image related tests need to be [Konflux integration tests](https://konflux.pages.redhat.com/docs/users/testing/integration/index.html) since those run after the image is built.

samdoran:
I think most of the work related to this needs to happen outside of this repo.
- Create a new git repo to build a container image that contains the test scripts, pipelines, and tasks
- Create a Konflux integration test in the `konflux-release-data` git repo that uses the above image

PR #132, opened 2025-12-22
Add functional testing to CI/CD

kokesak:
Add functional tests for MCP server to the CI workflow.

It is scheduled to run every day at 03:00 UTC


samdoran (commented):
This looks good. We just need to add a minimum set of permissions for the job.

PR #152, opened 2026-01-06
docs(tools): expand docstrings for all tool modules

major:
## Summary

- Update the `Host` type in `types.py` with a more descriptive annotation ("Remote host to connect to via SSH")
- Expand docstrings in all tool functions with meaningful descriptions

## Why

This enables mkdocstrings to display better parameter documentation in the generated API docs.

### system_info.py

| Function | Docstring Description |
|----------|----------------------|
| `get_system_information()` | Hostname, OS name/version, kernel version, architecture, uptime, boot time |
| `get_cpu_information()` | CPU model, core counts, frequency, load averages |
| `get_memory_information()` | RAM and swap usage (total, used, free, shared, buffers, cached, available) |
| `get_disk_usage()` | Filesystem usage for mounted volumes (size, used/available, utilization, mount points) |
| `get_hardware_information()` | Hardware inventory (CPU specs, PCI/USB devices, DMI/SMBIOS data) |

### processes.py

| Function | Docstring Description |
|----------|----------------------|
| `list_processes()` | Retrieves a snapshot of all running processes with details including PID, user, CPU/memory usage, process state, start time, and command line |
| `get_process_info()` | Retrieves comprehensive process details including CPU/memory usage, process state, virtual/resident memory size, controlling terminal, and additional metadata from /proc/<pid>/status |

### network.py

| Function | Docstring Description |
|----------|----------------------|
| `get_network_interfaces()` | Retrieves all network interfaces with their operational state, IP addresses, and traffic statistics (bytes/packets sent/received, errors, dropped packets) |
| `get_network_connections()` | Retrieves all established and pending network connections including protocol, state, local/remote addresses and ports, and associated process information |
| `get_listening_ports()` | Retrieves all ports with services actively listening for connections, including protocol (TCP/UDP), bind address, port number, and process name |

### services.py

| Function | Docstring Description |
|----------|----------------------|
| `list_services()` | Retrieves all systemd service units with their load state, active state, sub-state, and description. Also includes a count of currently running services |
| `get_service_status()` | Retrieves detailed service information including active/enabled state, main PID, memory usage, CPU time, and recent log entries from the journal |
| `get_service_logs()` | Retrieves journal entries for the specified service unit, including timestamps, priority levels, and log messages |

### storage.py

| Function | Docstring Description |
|----------|----------------------|
| `list_block_devices()` | Retrieves all block devices (disks, partitions, LVM volumes) with their name, size, type, mount point, and filesystem information |
| `list_directories()` | Retrieves subdirectories with their size (when ordered by size) or modification time, supporting flexible sorting and result limiting |
| `list_files()` | Retrieves files with their size or modification time, supporting flexible sorting and result limiting. Useful for finding large or recently modified files |
| `read_file()` | Retrieves the full contents of a text file. The path must be absolute and the file must exist |

### logs.py

| Function | Docstring Description |
|----------|----------------------|
| `get_journal_logs()` | Retrieves entries from the systemd journal with optional filtering by unit, priority level, and time range. Returns timestamped log messages |
| `get_audit_logs()` | Retrieves entries from /var/log/audit/audit.log containing security-relevant events such as authentication, authorization, and system call auditing. Requires root privileges |
| `read_log_file()` | Retrieves the last N lines from a log file. The file path must be in the allowed list configured via LINUX_MCP_ALLOWED_LOG_PATHS environment variable |

### DIFF
--- a/src/linux_mcp_server/tools/logs.py
+++ b/src/linux_mcp_server/tools/logs.py
@@ -46,8 +46,10 @@ async def get_journal_logs(
     lines: t.Annotated[int, Field(description="Number of log lines to retrieve. Default: 100")] = 100,
     host: Host | None = None,
 ) -> str:
-    """
-    Get systemd journal logs.
+    """Get systemd journal logs.
+
+    Retrieves entries from the systemd journal with optional filtering by unit,
+    priority level, and time range. Returns timestamped log messages.
     """
     try:
         # Validate lines parameter (accepts floats from LLMs)
@@ -81,8 +83,11 @@ async def get_audit_logs(
     lines: t.Annotated[int, Field(description="Number of log lines to retrieve.")] = 100,
     host: Host | None = None,
 ) -> str:
-    """
-    Get audit logs.
+    """Get Linux audit logs.
+
+    Retrieves entries from /var/log/audit/audit.log containing security-relevant
+    events such as authentication, authorization, and system call auditing.
+    Requires root privileges to read.
     """
     # Validate lines parameter (accepts floats from LLMs)
     lines, _ = validate_line_count(lines, default=100)
@@ -124,8 +129,10 @@ async def read_log_file(  # noqa: C901
     lines: t.Annotated[int, Field(description="Number of lines to retrieve from the end.")] = 100,
     host: Host | None = None,
 ) -> str:
-    """
-    Read a specific log file.
+    """Read a specific log file.
+
+    Retrieves the last N lines from a log file. The file path must be in the
+    allowed list configured via LINUX_MCP_ALLOWED_LOG_PATHS environment variable.
     """
     try:
         # Validate lines parameter (accepts floats from LLMs)


src/linux_mcp_server/tools/system_info.py
```diff
@@ -26,10 +29,14 @@
 @log_tool_call
 @disallow_local_execution_in_containers
 async def get_system_information(
-    host: Host | None = None,
+    host: t.Annotated[
+        Host | None, Field(description="Remote host to connect to via SSH. If None, runs locally.")
+    ] = None,
```
subpop: The `Host` type is already `t.Annotated` type. 

```python
Host = t.Annotated[str, Field(description="Remote host to connect to")]
```
https://github.com/rhel-lightspeed/linux-mcp-server/blob/main/src/linux_mcp_server/utils/types.py#L11

Shouldn't we just be able to update that type?

major: You're right. I've adjusted the PR to just update the docstrings a little. There's an issue where mkdocs isn't finding the annotation for some reason but I'll work on that separately.

src/linux_mcp_server/tools/logs.py
```diff
@@ -81,8 +83,11 @@ async def get_audit_logs(
     lines: t.Annotated[int, Field(description="Number of log lines to retrieve.")] = 100,
     host: Host | None = None,
 ) -> str:
-    """
-    Get audit logs.
+    """Get Linux audit logs.
+
+    Retrieves entries from /var/log/audit/audit.log containing security-relevant
+    events such as authentication, authorization, and system call auditing.
+    Requires root privileges to read.
```
subpop, line 90: This is true, but our command does not use sudo.

https://github.com/rhel-lightspeed/linux-mcp-server/blob/9cbe6a6328505a527cd1626fb7f8ddad2dd11021/src/linux_mcp_server/commands.py#L128-L132

Does this tool work?

src/linux_mcp_server/tools/processes.py
```diff
@@ -51,6 +56,12 @@ async def get_process_info(
     pid: t.Annotated[int, Field(description="Process ID")],
     host: Host | None = None,
 ) -> str:
+    """Get detailed information about a specific process.
+
+    Retrieves comprehensive process details including CPU/memory usage, process
```
subpop, line 61: "comprehensive" smells like LLM exaggeration. Let's remove it.

src/linux_mcp_server/tools/storage.py
```diff
@@ -111,8 +113,10 @@ async def list_directories(
     ] = None,
     host: Host | None = None,
 ) -> str:
-    """
-    List directories under a specified path.
+    """List directories under a specified path.
+
+    Retrieves subdirectories with their size (when ordered by size) or
+    modification time, supporting flexible sorting and result limiting.
```
subpop, line 119:
```suggestion
    Retrieves subdirectories sorted by size, modification time, or name. The sort order can be reversed. The resulting list can be limited to a fixed number.
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -172,8 +176,10 @@ async def list_files(
     ] = None,
     host: Host | None = None,
 ) -> str:
-    """
-    List files under a specified path.
+    """List files under a specified path.
+
+    Retrieves files with their size or modification time, supporting flexible
+    sorting and result limiting. Useful for finding large or recently modified files.
```
subpop, line 182:
```suggestion
    Retrieves files sorted by size, modification time, or name. The sort order can be reversed. The resulting list can be limited to a fixed number.
```

src/linux_mcp_server/tools/storage.py
```diff
@@ -221,8 +227,10 @@ async def read_file(
     path: t.Annotated[str, Field(description="The file path to read")],
     host: Host | None = None,
 ) -> str:
-    """
-    Read the contents of a file using cat.
+    """Read the contents of a file.
+
+    Retrieves the full contents of a text file. The path must be absolute
+    and the file must exist. Binary files may not display correctly.
```
subpop, line 233: This is a good observation. Should we only return the contents of files containing plain text? If we're retrieving file contents using cat, you won't get value bytes from catting a binary file through a terminal session, will you?

src/linux_mcp_server/tools/system_info.py
```diff
@@ -139,8 +147,11 @@ async def get_disk_usage(
 async def get_hardware_information(
     host: Host | None = None,
 ) -> str:
-    """
-    Get hardware information.
+    """Get hardware information.
+
+    Retrieves detailed hardware inventory including CPU specifications,
+    PCI devices, USB devices, and DMI/SMBIOS data (system manufacturer,
+    model, BIOS version, etc.). Some information may require root privileges.
```
subpop, line 154: Will it? We dropped the use of `dmidecode` and only run `lscpu`, `lspci` and `lsusb`. Do those require root?

PR #157, opened 2026-01-06
docs(tools): expand docstrings for logs tools

major:
## Summary

- Expand docstrings in `logs.py` tool functions with meaningful descriptions

## Why

This enables mkdocstrings to display better documentation in the generated API docs.

## Functions Updated

| Function | Docstring Description |
|----------|----------------------|
| `get_journal_logs()` | Retrieves entries from the systemd journal with optional filtering by unit, priority level, and time range. Returns timestamped log messages |
| `get_audit_logs()` | Retrieves entries from /var/log/audit/audit.log containing security-relevant events such as authentication, authorization, and system call auditing. Requires root privileges |
| `read_log_file()` | Retrieves the last N lines from a log file. The file path must be in the allowed list configured via LINUX_MCP_ALLOWED_LOG_PATHS environment variable |

### DIFF
--- a/src/linux_mcp_server/tools/logs.py
+++ b/src/linux_mcp_server/tools/logs.py
@@ -46,8 +46,10 @@ async def get_journal_logs(
     lines: t.Annotated[int, Field(description="Number of log lines to retrieve. Default: 100")] = 100,
     host: Host | None = None,
 ) -> str:
-    """
-    Get systemd journal logs.
+    """Get systemd journal logs.
+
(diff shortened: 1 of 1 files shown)


samdoran:
Can we combine these PRs into one since they are all making similar changes?

Also, if we only need to provide a description, we can use the [simple string descriptions](https://gofastmcp.com/servers/tools#simple-string-descriptions) now that we are using FastMCP>=2.11.0.

major:
Combined into #152

PR #160, opened 2026-01-06
read_file rejects reading binary files

subpop:
Add encoding parameter to remote execution functions to return either bytes or str. CommandSpec gains `run_bytes` method for raw output. The `read_file` tool now uses this and rejects binary files.

## Summary by CodeRabbit

* **New Features**
  * Added support for retrieving raw binary output from command execution.
  * Extended command execution with flexible encoding options (UTF-8 by default).

* **Bug Fixes**
  * Improved file reading with consistent encoding handling and enhanced path validation.

### DIFF
--- a/src/linux_mcp_server/tools/storage.py
+++ b/src/linux_mcp_server/tools/storage.py
@@ -232,21 +232,23 @@ async def read_file(
     Retrieves the full contents of a text file. The path must be absolute
     and the file must exist. Binary files may not display correctly.
     """
-    # For local execution, validate path
-    if not host:
-        path = _validate_path(path)
 
+    # Validate path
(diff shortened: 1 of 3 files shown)


src/linux_mcp_server/tools/storage.py
```diff
@@ -58,6 +58,37 @@ def _validate_path(path: str) -> str:
     return Path(path).as_posix()
 
 
+def _is_binary_file_local(path: str) -> bool:
```
r0x0d: Why not using the same `file` command locally?

subpop: I dropped the use of `file` entirely.

PR #175, opened 2026-01-07
Make substitute_command_args private

subpop:
Add some tests to ensure covrage for both `run` and `run_bytes`.

### DIFF
--- a/src/linux_mcp_server/commands.py
+++ b/src/linux_mcp_server/commands.py
@@ -38,11 +38,11 @@ class CommandSpec(BaseModel):
             host: Optional remote host address.
             **kwargs: Additional arguments passed to substitute_command_args.
         """
-        args = list(substitute_command_args(self.args, **kwargs))
+        args = list(_substitute_command_args(self.args, **kwargs))
         if self.optional_flags:
             for param_name, flag_args in self.optional_flags.items():
                 if kwargs.get(param_name):
-                    args.extend(substitute_command_args(flag_args, **kwargs))
+                    args.extend(_substitute_command_args(flag_args, **kwargs))
 
         returncode, stdout, stderr = await execute_with_fallback(tuple(args), fallback=self.fallback, host=host)
         stdout = stdout if isinstance(stdout, str) else stdout.decode("utf-8", errors="replace")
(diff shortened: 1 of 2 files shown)


tests/test_commands.py
```diff
@@ -118,3 +55,160 @@ def test_invalid_subcommand_raises_with_available(self):
         """Test that invalid subcommand raises KeyError listing available subcommands."""
         with pytest.raises(KeyError, match=r"Subcommand 'invalid' not found for 'system_info'.*Available:.*hostname"):
             get_command("system_info", "invalid")
+
+
+class TestCommandSpecRun:
+    """Tests for CommandSpec.run and CommandSpec.run_bytes methods."""
+
+    @pytest.mark.asyncio
```
samdoran: Decorating async test functions is not necessary since it is set in the main pytest configuration.

uv.lock
samdoran, line 1: I don't believe the dependency updates are relevant to this change and should be handled in a separate PR.

src/linux_mcp_server/commands.py
```diff
@@ -69,6 +69,36 @@ async def run_bytes(self, host: str | None = None, **kwargs: object) -> tuple[in
         stderr = stderr if isinstance(stderr, bytes) else stderr.encode("utf-8")
         return returncode, stdout, stderr
 
+    def _substitute_command_args(self, args: Sequence[str], **kwargs: object) -> tuple[str, ...]:
```
samdoran: What is the main purpose of changing this to a method? It's not referencing any state on the object and it makes it more difficult to test.

samdoran: If you want to mark it as private, you could just rename the function.

subpop: The object-oriented-programmer in me wants to make it private because it's never called or used outside the `CommandSpec` class, so moving it private, and attaching it to that class seems like the responsible thing to do. I guess it's easier to just leave it as a standalone function.

subpop: I dropped the commit that moved it to a method and just prefixed it with an underscore.

uv.lock
subpop: I left them as a separate commit, since they're atomically unrelated to the other changes in this PR. This PR is turning more into "fix a couple broken things from previous PRs".

tests/test_commands.py
subpop: Oh yea. I forgot about that again.

uv.lock
subpop: Looks like separating the commits isn't a good approach, since the only merge strategy available here is squash-and-merge. I'll drop those two extra commits from this PR and push them as two separate commit/PR pairs.

subpop: Now they're all separate PRs. Except #178 passes without running any CI jobs (because #177 fixes them), and #177 with running jobs fails jobs because the uv.lock file is fixed (in #178). 🙃

samdoran: Merged. Thanks for breaking those out!

src/linux_mcp_server/commands.py
samdoran: If you wanted to attach it to the class, it could be a `classmethod` since it doesn't need state. That would solve both problems since it would be attached to the class and still easy to test.

samdoran:
The journey is the reward. 😉

I think it would be fine to hang it off the class as a static method if it's only ever used by that class. Making it a method just complicated testing more than necessary, especially since it didn't need access to any state to do its job.

PR #180, opened 2026-01-07
Adding note for GSSAPI AysncSSH

aheath1992:
Adding a note in the install documentation that tells users to install the asyncssh[gssapi] pip package so that the linux-mcp-server will attempt to use gssapi-with-mic along side key based authentication

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -38,6 +38,11 @@ classifiers = [
     "Topic :: Utilities",
 ]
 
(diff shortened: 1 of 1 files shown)


r0x0d:
Maybe we should add this to `pyproject.toml` as well? Maybe we could add this as an extras option to the linux-mcp-server, where if they install `linux-mcp-server[kerberos]`, this dependency comes together.

PR #184, opened 2026-01-09
feat(ssh): add subprocess SSH backend with native config support

major:
## Summary

This PR introduces a subprocess-based SSH backend as an alternative to asyncssh, enabling full support for native SSH configurations on Linux/Unix systems.

## Why Raw SSH for Linux?

The asyncssh library is excellent for cross-platform SSH, but it has limitations when users have complex SSH configurations:

| Feature | asyncssh | Native SSH (this PR) |
|---------|----------|---------------------|
| **ProxyJump / ProxyCommand** | ❌ Manual config | ✅ Automatic via ~/.ssh/config |
| **Kerberos/GSSAPI auth** | ⚠️ Requires gssapi extras | ✅ Works out of the box |
| **Smartcard/PIV auth** | ❌ Not supported | ✅ Via ssh-agent |
| **ControlMaster multiplexing** | ❌ Not available | ✅ Built-in connection reuse |
| **~/.ssh/config inheritance** | ⚠️ Partial | ✅ Full support |
| **Corporate SSH configs** | ⚠️ Often broken | ✅ Just works |

### Key Benefits

1. **Zero configuration for most users** - If SSH works in your terminal, it works with linux-mcp-server
2. **Connection multiplexing** - ControlMaster reuses connections, reducing auth overhead
3. **FIPS compliance** - Uses system OpenSSH, respects system crypto policies

## Architecture

```
ssh.py (facade)
    ├── subprocess_ssh.py (default on Linux/macOS)
    │   └── Uses native `ssh` command with ControlMaster
    └── asyncssh_backend.py (default on Windows)
        └── Uses asyncssh library
```

Backend selection (automatic):
- **Linux/macOS**: subprocess backend (native ssh command)
- **Windows**: asyncssh backend (asyncssh library)

Override via environment variable:
```bash
LINUX_MCP_SSH_BACKEND=asyncssh    # Force asyncssh on any platform
LINUX_MCP_SSH_BACKEND=subprocess  # Force subprocess (requires ssh binary)
```

## Changes

- Add `subprocess_ssh.py` with `SSHSubprocessConnection` and `SSHSubprocessManager`
- Add `base.py` with protocol definitions for backend abstraction
- Refactor `ssh.py` as a facade that delegates to the appropriate backend
- Add lazy loading for asyncssh backend imports (avoids import errors when asyncssh not installed)
- Add `ssh_connect_timeout` config option (separate from command timeout) - works with both backends
- Log SSH backend selection at startup and in command logs
- Add comprehensive tests for the new backend
- Document SSH options with inline comments explaining each setting

## Testing

All existing tests pass. New tests cover:
- ControlMaster socket path generation (stays under 108-char limit)
- Connection pooling and reuse
- Timeout handling
- SSH connection error detection (exit code 255)

## Future Considerations

- [ ] Consider integration test that verifies both backends produce identical results

### DIFF
--- a/src/linux_mcp_server/connection/asyncssh_backend.py
+++ b/src/linux_mcp_server/connection/asyncssh_backend.py
@@ -220,7 +220,7 @@ class SSHAsyncSSHManager:
             stderr = result.stderr if result.stderr else b"" if encoding is None else ""
             duration = time.time() - start_time
 
-            log_ssh_command(cmd_str, host, exit_code=return_code, duration=duration)
+            log_ssh_command(cmd_str, host, exit_code=return_code, duration=duration, backend="asyncssh")
 
             return return_code, stdout, stderr
 
--- a/src/linux_mcp_server/connection/subprocess_ssh.py
+++ b/src/linux_mcp_server/connection/subprocess_ssh.py
@@ -178,7 +178,7 @@ class SSHSubprocessConnection:
                 self._connected = True
                 log_ssh_connect(self._host, username=CONFIG.user, status=Status.success, reused=False)
             duration = time.time() - start_time
-            log_ssh_command(remote_cmd, self._host, exit_code=return_code, duration=duration)
+            log_ssh_command(remote_cmd, self._host, exit_code=return_code, duration=duration, backend="subprocess")
 
             if encoding is None:
                 return return_code, stdout_bytes, stderr_bytes
--- a/src/linux_mcp_server/audit.py
+++ b/src/linux_mcp_server/audit.py
@@ -298,6 +298,7 @@ def log_ssh_command(
     host: str,
     exit_code: int,
     duration: float | None = None,
+    backend: str | None = None,
 ):
     """
     Log SSH command execution.
@@ -311,6 +312,7 @@ def log_ssh_command(
         host: Remote host
         exit_code: Command exit code
         duration: Optional execution duration in seconds (shown at DEBUG level)
+        backend: SSH backend used ("subprocess" or "asyncssh")
     """
     logger = logging.getLogger(__name__)
 
@@ -321,6 +323,10 @@ def log_ssh_command(
     }
 
     message = f"{Event.REMOTE_EXEC}: {command} | host={host} | exit_code={exit_code}"
+
+    if backend:
+        extra["backend"] = backend
+        message += f" | backend={backend}"
 
     # At DEBUG level, include duration
     if duration is not None and logger.isEnabledFor(logging.DEBUG):
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -4,6 +4,8 @@ import logging
 
 from fastmcp import FastMCP
 
+from linux_mcp_server.config import CONFIG
+
 
 logger = logging.getLogger("linux-mcp-server")
 
@@ -40,4 +42,5 @@ from linux_mcp_server.tools import *  # noqa: E402, F403
 
 
 def main():
+    logger.info(f"SSH backend: {CONFIG.effective_ssh_backend}")
     mcp.run(show_banner=False)


major:
@coderabbit Please review this draft PR.

major:
@coderabbit One last review please, I promise. I owe you a whole cake now! 🎂 What's your favorite flavor?

major:
I got this cleaned up enough that it's ready for a review. :wave:

src/linux_mcp_server/connection/asyncssh_backend.py
r0x0d, line 1: If this is meant to be a windows thing, we should also update the pyproject.toml to only install asyncssh if the target is windows, otherwise, don't even try to use this one. 

Unless, there is any use case where asyncssh would do good on linux/macos?

src/linux_mcp_server/connection/asyncssh_backend.py
```diff
@@ -0,0 +1,259 @@
...
+
+        for key_path in default_keys:
+            if key_path.exists() and key_path.is_file():
+                logger.info(f"Using SSH key: {key_path}")
+                return str(key_path)
+
+        logger.warning("No SSH private key found in default locations")
+
+    logger.debug("Not providing an SSH key")
+    return None
```
r0x0d, line 68:
```suggestion
```

src/linux_mcp_server/connection/base.py
```diff
@@ -0,0 +1,114 @@
...
+
+    Defines the interface for a single SSH connection to a remote host.
+    Implementations handle command execution and connection lifecycle.
+    """
+
+    async def run(
+        self,
+        command: str | Sequence[str],
+        timeout: int | None = None,
+        encoding: str | None = "utf-8",
```
r0x0d, line 54: I think it's safer to allow always an str to be passed, with the default of being `utf-8`

```suggestion
        encoding: str = "utf-8",
```

src/linux_mcp_server/connection/ssh.py
```diff
@@ -3,336 +3,117 @@
...
+_discover_ssh_key = None
+_get_remote_bin_path = None
 
-        logger.debug(f"Checking default SSH key locations: {[str(k) for k in default_keys]}")
 
-        for key_path in default_keys:
-            if key_path.exists() and key_path.is_file():
-                logger.info(f"Using SSH key: {key_path}")
-                return str(key_path)
+def _load_asyncssh_backend():
```
r0x0d, line 43: This is really confusing... In some places, we state that asyncssh is the default on Windows, but in others, we still use it if the user wants async connections via ssh. I think the docstrings/direction of it is a bit confusing

src/linux_mcp_server/connection/subprocess_ssh.py
```diff
@@ -0,0 +1,317 @@
...
+from linux_mcp_server.connection.base import parse_remote_bin_path
+
+
+logger = logging.getLogger("linux-mcp-server")
+
+# SSH exit code indicating connection-level failure (vs command failure)
+SSH_CONNECTION_ERROR_CODE = 255
+
+
+def _create_control_dir() -> Path:
```
r0x0d, line 36: We probably could use https://docs.python.org/3/library/tempfile.html#tempfile.mkdtemp instead to avoid race conditions on this

r0x0d: or even https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryDirectory if we want to re-create every time the temp dir

src/linux_mcp_server/connection/subprocess_ssh.py
```diff
@@ -0,0 +1,317 @@
...
+    on multi-user systems while keeping paths short for the 108-char
+    Unix socket path limit.
+    """
+    user_hash = hashlib.sha256(getpass.getuser().encode()).hexdigest()[:8]
+    control_dir = Path(tempfile.gettempdir()) / f"linux_mcp_{user_hash}"
+    control_dir.mkdir(mode=0o700, exist_ok=True)
+    return control_dir
+
+
+def _create_control_path(host: str) -> Path:
```
r0x0d, line 49: could we also try to use https://docs.python.org/3/library/tempfile.html#tempfile.TemporaryFile? I don't think we want a socket file lying around after the process finishes, right?

src/linux_mcp_server/connection/subprocess_ssh.py
```diff
@@ -0,0 +1,317 @@
...
+            "ControlMaster=auto",
+            "-o",
+            f"ControlPath={self._control_path}",
+            "-o",
+            f"ControlPersist={self._control_persist}",
+            "-o",
+            f"StrictHostKeyChecking={'yes' if CONFIG.verify_host_keys else 'no'}",
+            "-o",
+            f"UserKnownHostsFile={CONFIG.effective_known_hosts_path}",
+        ]
```
r0x0d, line 118: I know this may not be necessary, but what do you think about placing a comment above this saying why each option is being used?

src/linux_mcp_server/connection/subprocess_ssh.py
```diff
@@ -0,0 +1,317 @@
...
+        host: str,
+        control_persist: int = 300,
+    ):
+        self._host = host
+        self._control_path = _create_control_path(host)
+        self._control_persist = control_persist
+        self._ssh_bin = _find_ssh_binary()
+        self._connected = False
+
+    def _build_ssh_command(self, remote_command: str, timeout: int | None = None) -> list[str]:
```
r0x0d, line 85: Also, would be nice to comment what is the order of precedence here... I'm not sure if the definition over CLI takes precedence over the ones configured in ~/.ssh/config

src/linux_mcp_server/connection/subprocess_ssh.py
```diff
@@ -0,0 +1,317 @@
...
+        timeout: int,
+    ) -> str:
+        """Get the full path to an executable on the remote system."""
+        logger.debug(f"Getting path for {command} on {host}")
+        returncode, stdout, _ = await conn.run(
+            ["command", "-v", command],
+            timeout=timeout,
+            encoding="utf-8",
+        )
+        return parse_remote_bin_path(command, host, returncode, stdout, username=CONFIG.user)
```
r0x0d, line 315: should some version of this be in the base ssh.py class instead? to share similarity with asyncssh

src/linux_mcp_server/connection/subprocess_ssh.py
major: Ah, good point. Will do.

src/linux_mcp_server/connection/asyncssh_backend.py
major: Well the asyncssh stuff gets tested on Linux, so I've gotta leave it in there. It also gives us the chance to use either on linux if we feel like we need to.

src/linux_mcp_server/connection/base.py
major: So it's there on purpose. I had to go research this, but returning none is a signal to return raw bytes. Otherwise you might end up in a UTF-8 encoding situation that breaks the return data.

src/linux_mcp_server/connection/ssh.py
major: I'll add some comments

src/linux_mcp_server/connection/subprocess_ssh.py
major: This would disrupt the command multiplexing if we used tempfile. Also tempfile doesn't work for sockets last i checked.

src/linux_mcp_server/connection/subprocess_ssh.py
major: This would disrupt the command multiplexing if we used tempfile. Also tempfile doesn't work for sockets last i checked.

src/linux_mcp_server/connection/base.py
r0x0d: But that would need to be asked by the caller, do we have any case in the code that remove the encoding? Otherwise, this will always be `utf-8` and never None. 

But it looks like to me that we don't want this to be passed as `None`?

subpop: I think I wrote this into the code. Setting encoding to `None` to return raw bytes is a derived behavior from AsyncSSH. If you trace the calls down far enough into AsyncSSH, you can see the `encoding` parameter. It is documented that if you set the value to `None`, you will get a byte slice back. If you want encoded strings, you can set the value to a valid encoding scheme. I mirrored that pattern when I added encoding to our `run` methods.

src/linux_mcp_server/connection/subprocess_ssh.py
```diff
@@ -0,0 +1,330 @@
...
+        self._control_persist = control_persist
+        self._ssh_bin = _find_ssh_binary()
+        self._connected = False
+
+    def _build_ssh_command(self, remote_command: str, timeout: int | None = None) -> list[str]:
+        """Build the ssh command with all required options.
+
+        SSH options set here take precedence over ~/.ssh/config settings.
+        This is intentional to ensure consistent behavior regardless of
+        user configuration.
```
subpop, line 90: This is the way the `ssh` program itself behaves, IIRC.

PR #190, opened 2026-01-12
[RSPEED-2269] Convert system information commands to structured output

subpop:
Refactor system information commands to return structured output instead of raw strings. Introduce a new parser for disk usage output and update related functions to enhance error handling and data representation.

Instead of trying to parse output from `lspci`, `lscpu` and `lsusb` into fully typed data structures, I left them as partially parsed. For the USB and PCI devices, I split the output by line into a list. For `lscpu`, I left it unmodified; it's a big output of custom string, and I really didn't see much point in converting it into a data structure (even though it supports JSON formatted output).


src/linux_mcp_server/parsers.py
```diff
@@ -548,3 +549,43 @@ def parse_file_listing(
...
+    lines = stdout.strip().split("\n")[1:]  # Skip header line
+
+    for line in lines:
+        if not line.strip():
+            continue
+        parts = line.split()
+        if len(parts) < 6:
+            continue
+        filesystem = parts[0]
+        size_gb = float(parts[1]) / (1024 * 1024 * 1024)
```
major: Hmm, isn't this one too many divisions by 1024? The data coming from `df` is already in kb.

subpop: Yea, I think I had my scales thrown off by using KB. I updated it to run `df` with `--block-size=1M`, so the raw data is reported as megabytes, and then the parser scales it to GB. I also updated the tests to use fixed expected values instead of computing them. It makes the test easier to read what's going on.

src/linux_mcp_server/utils/types.py
```diff
@@ -121,6 +121,17 @@ class CpuInfo(BaseModel):
...
 
 
+class DiskUsage(BaseModel):
+    """Parsed disk usage information."""
+
+    filesystem: str = ""
+    size_gb: float = 0.0
+    used_gb: float = 0.0
+    available_gb: float = 0.0
+    use_percent: float = 0.0
```
samdoran: If necessary, we could enforce this to be non-negative values.

```python
    size_gb: t.Annotated[float, Field(ge=0.0) = 0.0
```

tests/tools/test_system_info.py
```diff
@@ -10,6 +10,27 @@ def mock_execute(mock_execute_with_fallback_for):
...
+
+    command_responses: dict of command_name -> output_string or Exception
+    """
+
+    def mock_execute_side_effect(*args, **_kwargs):
+        cmd = args[0]
+        match cmd[0]:
+            case cmd_name if cmd_name in command_responses:
+                response = command_responses[cmd_name]
+                if isinstance(response, Exception):
```
samdoran: I'm confused by this. Does `command_response[cmd_name]` _return_ an `Exception` object instead of raising? That seems like an odd interface.

tests/tools/test_system_info.py
```diff
@@ -64,10 +85,13 @@ async def test_system_info_tools(tool, expected, mcp_client):
 async def test_system_info_tools_unsuccessful(tool, mcp_client, mock_execute):
     mock_execute.return_value = (0, "", "")
 
-    result = await mcp_client.call_tool(tool)
-    result_text = result.content[0].text.casefold()
+    # These tools now raise ToolError when output is unsuccessful
+    # The MCP client's raise_on_error=True (default) causes it to raise an exception
+    with pytest.raises(Exception) as exc_info:
```
samdoran: This may hide legitimate failures since the test will pass when _any_ exception is raised. It would be better to catch a specific exception and match on the error message.

tests/tools/test_system_info.py
```diff
@@ -64,10 +85,13 @@ async def test_system_info_tools(tool, expected, mcp_client):
...
+        await mcp_client.call_tool(tool)
 
-    assert "error" in result_text
+    # Verify the error message contains expected text
+    assert "error" in str(exc_info.value).casefold()
```
samdoran: Use the [`match` parameter](https://docs.pytest.org/en/stable/reference/reference.html#pytest-raises) instead.
```python
with pytest.raises(ValueError, match="error"):
    await mcp_client.call_tool(tool)
```

tests/tools/test_system_info.py
```diff
@@ -81,6 +105,177 @@ async def test_system_info_tools_unsuccessful_empty(tool, mcp_client, mock_execu
...
+    ],
+)
+async def test_system_info_tools_exception(tool, failing_command, mcp_client, mock_execute):
+    command_responses = {failing_command: Exception("Command failed")}
+    mock_execute.side_effect = create_mock_execute_side_effect(command_responses)
+
+    with pytest.raises(Exception) as exc_info:
+        await mcp_client.call_tool(tool)
+
+    assert "error gathering" in str(exc_info.value).casefold()
```
samdoran: Same thing here: be more specific in what exception is caught and use `match`.
```suggestion
    with pytest.raises(ToolError, match="error gathering"):
        await mcp_client.call_tool(tool)
```

tests/tools/test_system_info.py
```diff
@@ -81,6 +105,177 @@ async def test_system_info_tools_unsuccessful_empty(tool, mcp_client, mock_execu
...
+
+async def test_get_disk_usage_parse_error(mcp_client, mock_execute):
+    """Test get_disk_usage with unexpected exception during execution."""
+    # Mock cmd.run to raise an unexpected exception
+    mock_execute.side_effect = RuntimeError("Unexpected error")
+
+    with pytest.raises(Exception) as exc_info:
+        await mcp_client.call_tool("get_disk_usage")
+
+    assert "error gathering disk usage information" in str(exc_info.value).casefold()
```
samdoran: When I'm raising exceptions intentionally for testing, I'll usually use `VauleError('Raised intentionally')` to make it easy to ensure it's my test exception that is being raised and not some other failure. This might not work since we're catching and raising another exception instead of letting it raise directly, but you get the idea.
```suggestion
    mock_execute.side_effect = ValueError("Unexpected error")

    with pytest.raises(ValueError, match="error gathering disk usage information"):
        await mcp_client.call_tool("get_disk_usage")
```

tests/tools/test_system_info.py
```diff
@@ -81,6 +105,177 @@ async def test_system_info_tools_unsuccessful_empty(tool, mcp_client, mock_execu
...
+
+async def test_get_hardware_information_unexpected_exception(mcp_client, mock_execute):
+    """Test get_hardware_information with unexpected exception from command execution."""
+    # Mock cmd.run to raise an unexpected exception (not FileNotFoundError)
+    mock_execute.side_effect = RuntimeError("Unexpected error")
+
+    with pytest.raises(Exception) as exc_info:
+        await mcp_client.call_tool("get_hardware_information")
+
+    assert "error gathering hardware information" in str(exc_info.value).casefold()
```
samdoran: The value to the `match` parameter is a string or regexp object, so you could pass in a case-insensitive regexp object if you want to do case-insensitive matching.

```suggestion
    with pytest.raises(ToolError, match="Error gathering hardware information"):
        await mcp_client.call_tool("get_hardware_information")
```

tests/tools/test_system_info.py
subpop: No, it's not a function, it doesn't return a value. It's a dictionary that
contains either a string or an Exception. Here's an example use:

```
command_responses = {failing_command: Exception("Command failed")}
```

In order to test whether the execute function raises an Exception, the responses
are dynamically loaded from a dictionary. In the case of a failing command, we
load the response with an Exception. The next line `raise response` will raise
that Exception if its an Exception. I admit it's kind of hacky, but it let me
cover those last few lines in order to get code coverage up to 100%. I'll try
typing the functions here to see if that makes more sense.

samdoran: Ok, I see what this is doing. Makes more sense when I have fresh eyes.

samdoran: Now that I better understand what this is doing, I think it can be written more clearly using an assignment expression. Pattern matching isn't adding much value here and makes it slightly more difficult to read.
```python
def mock_execute_side_effect(*args, **_kwargs):
    cmd = args[0][0]
    if response := command_responses.get(cmd):
        if isinstance(response, Exception):
            raise response
            
        return (0, response, "")
    
    raise AssertionError(f"Unexpected command in test mock: {cmd[0]}")
```

subpop: Yea, that's more concise. I like that.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -88,21 +87,26 @@ async def get_cpu_information(
...
-    memory = parse_free_output(stdout)
-    return format_memory_info(memory)
+        if not is_successful_output(returncode, stdout):
+            raise ToolError("Unable to retrieve memory information")
+
+        return parse_free_output(stdout)
+    except ToolError:
+        raise
+    except Exception as e:
+        raise ToolError(f"Error gathering memory information: {str(e)}") from e
```
samdoran, line 103: I think this might be a bad merge.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -115,20 +119,25 @@ async def get_memory_information(
...
-        raise ToolError("Unable to retrieve disk usage information")
+        if not is_successful_output(returncode, stdout):
+            raise ToolError("Unable to retrieve disk usage information")
 
-    return format_disk_usage(stdout)
+        return parse_df_output(stdout)
+    except ToolError:
+        raise
+    except Exception as e:
+        raise ToolError(f"Error gathering disk usage information: {str(e)}") from e
```
samdoran, line 131: Same here.

src/linux_mcp_server/tools/system_info.py
```diff
@@ -141,23 +150,28 @@ async def get_disk_usage(
...
+                if is_successful_output(returncode, stdout):
+                    results[name] = stdout if name == "lscpu" else stdout.splitlines()
+                else:
+                    results[name] = f"Error retrieving {name}: {stderr}"
+            except FileNotFoundError:
+                results[name] = f"{name} command not available"
+
+        return results
+    except Exception as e:
+        raise ToolError(f"Error gathering hardware information: {str(e)}") from e
```
samdoran: Here as well.

tests/parsers/test_parse_disk_usage.py
```diff
@@ -0,0 +1,113 @@
...
+
+    assert result == []
+
+
+def test_parse_df_output_header_only():
+    """Test parsing output with only header."""
+    stdout = """Filesystem     1M-blocks      Used Available Use% Mounted on"""
+    result = parse_df_output(stdout)
+
+    assert result == []
```
samdoran: These two tests can be combined.
```suggestion
@pytest.mark.parametrize("output",("", "Filesystem     1M-blocks      Used Available Use% Mounted on"))
def test_parse_df_output(output):
    """Test parsing empty output."""
    result = parse_df_output(output)
    
    assert result == []
```

src/linux_mcp_server/parsers.py
```diff
@@ -548,3 +549,45 @@ def parse_file_listing(
             entries.append(NodeEntry(name=line.strip()))
 
     return entries
+
+
+def parse_df_output(stdout: str) -> list[DiskUsage]:
```
samdoran: One thing that might make this much easier is using `findmnt --df --json` instead of `df`.

subpop: That would be useful, but `/usr/bin/findmnt` is part of the `util-linux-core` package. I’m being mindful of which binaries we use to keep our required package list manageable. `util-linux-core` sounds like a safe one to require though.

```
Summary     : The most essential utilities from the util-linux suite
Description :
This is a very basic set of Linux utilities that is necessary on
minimal installations.
```

src/linux_mcp_server/tools/system_info.py
subpop: Whoops. You're right. The rebase was grueling yesterday. I'm not surprised I made a mistake.

PR #198, opened 2026-01-13
refactor(logs): consolidate audit logs into get_journal_logs

major:
## Summary

- Add `transport` parameter to `get_journal_logs` to support filtering by journal transport (e.g., `audit`, `kernel`)
- Remove the now-redundant `get_audit_logs` tool since audit logs can be retrieved via `get_journal_logs(transport="audit")`
- Audit log header now displays additional filters (priority, since) when combined with transport=audit

This consolidates two separate log retrieval approaches into a single, more flexible tool.


major:
@coderabbit Come on, rip this PR to shreds. Tell me how you really feel. Let's have that review! 😜

src/linux_mcp_server/tools/logs.py
```diff
@@ -19,6 +18,9 @@
 from linux_mcp_server.utils.validation import is_empty_output
 
 
+VALID_TRANSPORTS = frozenset({"audit", "driver", "journal", "kernel", "stdout", "syslog"})
```
r0x0d: I think we should have a brief comment on each transport here, like, when they are used and in which case

src/linux_mcp_server/tools/logs.py
```diff
@@ -74,17 +77,23 @@ async def get_journal_logs(
         str | None,
         "Filter entries since specified time. Date/time filter (format: 'YYYY-MM-DD HH:MM:SS', 'today', 'yesterday', 'now', or relative like '-1h')",
     ] = None,
+    transport: t.Annotated[
+        str | None,
+        "Filter by journal transport (e.g., 'audit' for audit logs, 'kernel' for kernel messages, 'syslog' for syslog messages)",
+    ] = None,
```
r0x0d, line 103: uh, should we make a enum or something with the transport options? This way we can get this parameter typed and only choose one of the available options

r0x0d: also, maybe we should set a default for syslog? Not sure

src/linux_mcp_server/formatters.py
```diff
@@ -352,25 +354,23 @@ def format_journal_logs(
...
+    # Use "Audit Logs" header when transport=audit for clarity
+    if transport == "audit":
+        non_transport_filters = [f for f in filters if not f.startswith("transport=")]
+        if non_transport_filters:
+            extra_filters = ", ".join(non_transport_filters)
+            header = f"=== Audit Logs (last {lines_count} entries, {extra_filters}) ===\n"
+        else:
+            header = f"=== Audit Logs (last {lines_count} entries) ===\n"
+    else:
+        header = f"=== Journal Logs (last {lines_count} entries, {filter_desc}) ===\n"
```
r0x0d: Just a thought here... Could we use the transport name in the str to simplify this? 


```python
non_transport_filters = [f for f in filters if not f.startswith("transport=") and transport]
transport_name = transport.title() if transport else "Journal"
filters = f", {extra_fitlers}" if non_transport_filters else f", {filter_desc}"
header = f"=== {transport_name} Logs (last {lines_count} entries{filters}) ===\n"
```

Something around those lines... Although, in my suggestion above, line 369 get removed... Either we show the `extra_filters` or `filter_desc`

r0x0d: well, maybe that does not add a lot of clarity, but it's just a suggestion

src/linux_mcp_server/tools/logs.py
major: Oh man, I intended to go back and do that but forgot 🤦‍♂️

major: Actually, `None` is what we want here by default. You don't want to limit by transport unless you know what you're looking for.

PR #211, opened 2026-01-15
Add job for publishing to MCP registry

r0x0d:
Add a new job to publish to the mcp registry on every release

### DIFF
--- a/server.json
+++ b/server.json
@@ -82,7 +82,7 @@
         },
         {
           "name": "LINUX_MCP_LOG_LEVEL",
-          "description": "Path to SSH private key",
+          "description": "Log verbosity (DEBUG, INFO, WARNING)",
           "isRequired": false,
           "format": "string",
           "isSecret": false


server.json
```diff
@@ -0,0 +1,100 @@
...
+  "repository": {
+    "url": "https://github.com/rhel-lightspeed/linux-mcp-server",
+    "source": "github"
+  },
+  "version": "1.2.1",
+  "packages": [
+    {
+      "registryType": "pypi",
+      "identifier": "linux-mcp-server",
+      "version": "1.2.0",
```
jeffmaury: **question:** why is it different from the version at line 9 ?

r0x0d: This one comes from the actual version in pypi for the mcp server, and the above in line 9 is a version identifier for the mcp registry server, I believe. 

Also, had to bump because I pushed with wrong metadata :facepalm:

r0x0d: Also, that would get override anyway in the next release, since we have a step in the workflow to modify the `version` field accordingly with the tag name

.github/workflows/build-publish.yml
```diff
@@ -73,3 +73,31 @@ jobs:
...
+      - name: Authenticate to MCP Registry
+        run: ./mcp-publisher login github-oidc
+
+      - name: Set version in server.json
+        run: |
+           VERSION=${GITHUB_REF#refs/tags/v}
+           jq --arg v "$VERSION" '.version = $v' server.json > server.tmp && mv server.tmp server.json
+
+      - name: Publish server to MCP Registry
+        run: ./mcp-publisher publish
```
samdoran: Rather than inline shell scripts, can this be changed to a standalone script that this job calls?

.github/workflows/build-publish.yml
```diff
@@ -73,3 +73,31 @@ jobs:
...
+    environment:
+      name: mcp
+
+    permissions:
+      id-token: write
+
+    steps:
+      - name: Install mcp-publisher
+        run: |
+          curl -L "https://github.com/modelcontextprotocol/registry/releases/latest/download/mcp-publisher_$(uname -s | tr '[:upper:]' '[:lower:]')_$(uname -m | sed 's/x86_64/amd64/;s/aarch64/arm64/').tar.gz" | tar xz mcp-publisher
```
samdoran: Downloading this file every time may lead to spontaneous breakage. Could this be put into a container image that we use to run the job and then we decide when to update `mcp-publisher`?

server.json
```diff
@@ -0,0 +1,100 @@
...
+      "transport": {
+        "type": "stdio"
+      },
+      "environmentVariables": [
+        {
+          "name": "LINUX_MCP_USER",
+          "description": "Default username for SSH connections",
+          "isRequired": true,
```
samdoran, line 22: I'm curious where this comes from because this parameter is not required.

server.json
samdoran: I wonder if this should be a placeholder value and not a number. It would be a hassle to keep this file up to date with actual version.

.github/workflows/build-publish.yml
```diff
@@ -73,3 +73,31 @@ jobs:
...
+
+      - name: Authenticate to MCP Registry
+        run: ./mcp-publisher login github-oidc
+
+      - name: Set version in server.json
+        run: |
+           VERSION=${GITHUB_REF#refs/tags/v}
+           jq --arg v "$VERSION" '.version = $v' server.json > server.tmp && mv server.tmp server.json
```
samdoran: This command only changes the top level version but not the PyPI version. I understand why they are different but don't we want them to both be updated in an automated fashion on release?

server.json
r0x0d: That's a better idea, indeed. I will make it a placeholder

server.json
r0x0d: I think it was a leftover from copy & paste. Changed it to false.

samdoran: Ok. I wanted to make sure it wasn't being generated by the schema somewhere.

server.json
```diff
@@ -0,0 +1,100 @@
...
+        {
+          "name": "LINUX_MCP_LOG_DIR",
+          "description": "Server log directory",
+          "isRequired": false,
+          "format": "string",
+          "isSecret": false
+        },
+        {
+          "name": "LINUX_MCP_LOG_LEVEL",
+          "description": "Path to SSH private key",
```
jeffmaury: **suggestion:** See #222

r0x0d: oh, nice find. Let me update.

r0x0d: https://github.com/rhel-lightspeed/linux-mcp-server/pull/211/commits/f5eecfdf9190ab095bdfd84abe6b5bf98a116fe5

samdoran (approved):

PR #215, opened 2026-01-16
Omit '.venv/*' from coverage report

subpop:
Exclude the virtual environment from the coverage report to significantly reduce the coverage run time.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -108,7 +108,7 @@ source = [
 ]
 omit = [
     "src/linux_mcp_server/_vendor/*/*",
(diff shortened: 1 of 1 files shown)


samdoran:
This shouldn't be necessary since the coverage settings are only looking at the `src` and `tests` directories.

samdoran:
I would like to better understand why this is necessary. I wasn't able to duplicate the problem in VS Code, so it may be a local configuration. I'm not sure exactly.

We can add a comment why it's there and merge it in the mean time.

PR #216, opened 2026-01-19
Add semantic tags to all MCP tools

major:
## Summary

- Add semantic tags to all 19 MCP tools for improved LLM tool categorization and server-side filtering
- Tags are granular and alphabetically ordered (e.g., `{"connectivity", "network", "ports"}`)
- Cross-cutting tags like `troubleshooting` and `logs` applied where relevant

## Tag Taxonomy

| Module | Tags Applied |
|--------|--------------|
| logs | `files`, `journal`, `logs`, `systemd`, `troubleshooting` |
| network | `connections`, `connectivity`, `interfaces`, `network`, `ports` |
| processes | `performance`, `processes` |
| services | `logs`, `services`, `systemd`, `troubleshooting` |
| storage | `devices`, `directories`, `files`, `filesystem`, `storage` |
| system_info | `cpu`, `disk`, `filesystem`, `hardware`, `memory`, `performance`, `storage`, `system` |

## Why

FastMCP's `tags` parameter enables:
1. **LLM tool selection** - Tags appear in tool schemas, helping LLMs group related tools
2. **Server-side filtering** - Operators can filter tools via `include_tags`/`exclude_tags`
3. **Dynamic visibility** - Tools can be enabled/disabled per session based on tags

PR #217, opened 2026-01-19
feat(tools): add parameter examples for LLM guidance

major:
## Summary

Add `Field(examples=[...])` to tool parameters with non-obvious formats, helping LLMs understand expected input patterns.

### Parameters Updated (10 total)
- **logs.py**: `unit`, `priority`, `since`, `log_path`
- **services.py**: `service_name` (2 tools)
- **processes.py**: `pid`
- **storage.py**: `path` (3 tools)

### Examples Added
| Parameter | Examples |
|-----------|----------|
| `unit` | `sshd.service`, `nginx`, `httpd`, `systemd-*`, `audit*` |
| `priority` | `err`, `warning`, `info`, `debug`, `3`, `err..warning` |
| `since` | `today`, `yesterday`, `-1h`, `-30m`, `-7d`, `2025-01-15 10:00:00` |
| `log_path` | `/var/log/messages`, `/var/log/secure`, `/var/log/audit/audit.log` |
| `service_name` | `sshd`, `NetworkManager`, `auditd`, `rsyslog`, `crond`, `firewalld` |
| `pid` | `1`, `1234`, `65535` |
| `path` (dirs/files) | `/var/log`, `/etc`, `/home`, `/opt`, `/tmp` |
| `path` (read_file) | `/etc/hosts`, `/etc/resolv.conf`, `/etc/os-release`, `/proc/cpuinfo` |

### Testing
- Added `tests/test_tool_schemas.py` with parameterized tests verifying examples appear in tool schemas
- All 425 tests pass
- `make verify` passes

### Related
- Depends on #216 (adds tool tags)
- Part of Phase 2 from FastMCP feature usage refactor plan

### DIFF
--- a/tests/test_tool_schemas.py
+++ b/tests/test_tool_schemas.py
@@ -18,7 +18,7 @@ def tool_properties(mcp_tools: dict):
     def _tool_properties(tool_name: str) -> dict:
         tool = mcp_tools.get(tool_name)
         if tool is None:
-            raise ValueError(f"Tool '{tool_name}' not found")
+            raise ValueError(f"Tool '{tool_name}' not found")  # pragma: no cover
         return tool.parameters.get("properties", {})
 
     return _tool_properties


tests/test_tool_schemas.py
```diff
@@ -0,0 +1,40 @@
...
+
+from linux_mcp_server.server import mcp
+
+
+def get_tool_schema(tool_name: str) -> dict:
+    """Look up a tool's JSON schema by name."""
+    for tool in mcp._tool_manager._tools.values():
+        if tool.name == tool_name:
+            return tool.parameters  # type: ignore[return-value]
+    raise ValueError(f"Tool '{tool_name}' not found")
```
samdoran: Rather than using private attributes, it would be better to use the public interface and make use of fixtures.
```suggestion
@pytest.fixture(scope="module")
async def mcp_tools():
    return await mcp.get_tools()
```

In the test:
```python
def test_parameter_has_examples(tool_name, param_name, mcp_tools):
    tool = mcp_tools.get(tool_name, {})
    props = tool.parameters.get("properties", {})
```

This also removes the unused test code line.

samdoran: This could also be combined into a single fixture if you wanted a simpler interface.
```python
@pytest.fixture
def tool_properties(mcp_tools):
    def _tool_properties(tool_name):
        tool = mcp_tools.get(tool_name, {})
        return tool.parameters.get("properties", {})

    return _tool_properties
...

def test_parameter_has_examples(tool_name, param_name, tool_properties):
    props = tool_properties(tool_name)
```

major: Can do!

PR #223, opened 2026-01-20
[RSPEED-2268] Return typed objects from storage tools

samdoran:
Create Pydantic models and return those from storage functions instead of formatted strings.
Update commands when possible to return JSON data to make processing the command output much easier. Delete format functions that are no longer necessary.

Updated tools:
- [x] list_block_devices
- [x] list_directories
- [x] list_files

### DIFF
--- a/src/linux_mcp_server/tools/storage.py
+++ b/src/linux_mcp_server/tools/storage.py
@@ -3,7 +3,6 @@
 import os
 import typing as t
 
-from operator import attrgetter
 from pathlib import Path
 
 from fastmcp.exceptions import ToolError
@@ -38,6 +37,18 @@ class SortBy(StrEnum):
     DESCENDING = "descending"
 
 
+def attr_sorter(order_by: OrderBy):
+    """Sort based on the given attribute in a case-insensitive manner"""
+
+    def _attr_sorter(obj):
+        try:
+            return getattr(obj, order_by).casefold()
+        except AttributeError:
+            return getattr(obj, order_by)
+
+    return _attr_sorter
+
+
 async def _list_resources(
     path: Path,
     command: CommandSpec,
@@ -57,7 +68,7 @@ async def _list_resources(
     entries = parser(stdout, order_by)
 
     reverse = sort == SortBy.DESCENDING
-    entries = sorted(entries, key=attrgetter(order_by), reverse=reverse)
+    entries = sorted(entries, key=attr_sorter(order_by), reverse=reverse)
     entries = entries[:top_n]
 
     return StorageNodes(nodes=entries)
--- a/tests/tools/storage/test_list_files.py
+++ b/tests/tools/storage/test_list_files.py
@@ -9,7 +9,7 @@ from fastmcp.exceptions import ToolError
 async def test_list_files(setup_test_paths, mcp_client, tmp_path):
     file_specs = [
         ("file1.txt", 100, 1000.0),
-        ("file2.txt", 200, 2000.0),
+        ("File2.txt", 200, 2000.0),
         ("file3.txt", 300, 3000.0),
     ]
     setup_test_paths(file_specs)


src/linux_mcp_server/models.py
```diff
@@ -0,0 +1,36 @@
...
+    size: str
+    type: str
+    mountpoint: str | None = None
+    fstype: str | None = None
+    model: str | None = None
+    children: list["BlockDevice"] = []
+
+
+class BlockDevices(BaseModel):
+    blockdevices: list[BlockDevice]
```
r0x0d: Should we put a `_` here to make it better to read? 
```suggestion
    block_devices: list[BlockDevice]
```

samdoran: Yup, I can do that. `blockdevices` comes from the JSON output, so I'll need to make an alias so the initialization will work.

src/linux_mcp_server/models.py
```diff
@@ -0,0 +1,36 @@
...
+    size: str
+    type: str
+    mountpoint: str | None = None
+    fstype: str | None = None
+    model: str | None = None
+    children: list["BlockDevice"] = []
+
+
+class BlockDevices(BaseModel):
+    block_devices: list[BlockDevice] = Field(alias="blockdevices")
```
r0x0d, line 34: Would be good to document why this class only has one attribute? I think one could also assume that we could do list[BlockDevice] instead of having a class like this

src/linux_mcp_server/models.py
```diff
@@ -0,0 +1,36 @@
...
+    name: str
+    size: str
+    type: str
+    mountpoint: str | None = None
+    fstype: str | None = None
+    model: str | None = None
+    children: list["BlockDevice"] = []
+
+
+class BlockDevices(BaseModel):
```
r0x0d, line 33: maybe we should have a total property here like StorageNodes?

tests/tools/storage/conftest.py
```diff
@@ -0,0 +1,97 @@
...
+            content_file = tmp_path / name
+            content_file.write_text("x" * size)
+
+            os.utime(content_file, (modified_time, modified_time))
+
+            expected_names.append(name)
+
+        return tmp_path, expected_names
+
+    return _create_files
```
r0x0d: maybe this is not for this PR, but would it make sense if we tried to merge those two together? Or maybe just some portion of it. They have a similar idea between them, only changing "when" it stops.

tests/tools/storage/test_list_block_devices.py
```diff
@@ -0,0 +1,82 @@
...
+import pytest
+
+from fastmcp.exceptions import ToolError
+
+
+@pytest.mark.parametrize(
+    ("lsblk_output", "expected_content"),
+    [
+        pytest.param(
+            '{"blockdevices":[{"name":"sda","size":"1TB","type":"disk","mountpoint":null,"fstype":null,"model":null,"children":[{"name":"sda1","size":"512G","type":"part","mountpoint":"/","fstype":"ext4","module":null}]}]}',
```
major: Did you mean `model` rather than `model`?

tests/tools/storage/test_list_block_devices.py
```diff
@@ -0,0 +1,82 @@
...
+            ["=== Block Devices ===", "sda", "sda1"],
+            id="lsblk_success",
+        ),
+    ],
+)
+async def test_list_block_devices_lsblk_success(
+    lsblk_output,
+    expected_content,
```
major: You have `expected_content` here, but it's never used in the function as far as I can tell.

tests/tools/storage/test_list_block_devices.py
samdoran: Yup!

tests/tools/storage/test_list_block_devices.py
samdoran: Doh, that's correct. I'll fix it up.

tests/tools/storage/test_list_block_devices.py
major: Both sound similar in a southern accent. 😆

tests/tools/storage/conftest.py
samdoran: I can do that. These tests are getting better but still need a lot of love.

src/linux_mcp_server/models.py
samdoran: The advantage to having a model like this is that it validates every element as well as handles deserializing the text output. We could do this manually but it's cleaner to use a model.

```python
data = json.loads(stdout)
devices = data["blockdevices"]
return [BlockDevice.model_validate(device) for device in devices]
```

tests/tools/storage/test_list_block_devices.py
samdoran: I knew what you meant. 😉

PR #225, opened 2026-01-21
[RSPEED-2267] Convert service commands to return structured output

subpop:
Refactor service-related tools to return structured data instead of plain text. This includes converting the output of `list_services`, `get_service_status`, and `get_service_logs` into JSON. Convert `get_service_status` to use `systemctl show` instead of the human-readably focused `systemctl status`. Implement a parser for the `systemctl show` command to facilitate this structured output.


src/linux_mcp_server/commands.py
```diff
@@ -103,12 +105,14 @@ class CommandGroup(BaseModel):
         ),
         "service_status": CommandGroup(
             commands={
-                "default": CommandSpec(args=("systemctl", "status", "{service_name}", "--no-pager", "--full")),
+                "default": CommandSpec(args=("systemctl", "show", "{service_name}", "--no-pager")),
```
owtaylor: This isn't really the same thing at all ... `systemctl show` returns the complete expanded list of properties which includes some status information, but `systemctl status` provides the most relevant information included information about things like cgroups, and journalctl logs that aren't in the `systemctl show`. 

`show` also eats a lot more tokens of context.

```
$ systemctl show avahi-daemon.service --no-pager  | wc -c
8729
$ systemctl status avahi-daemon.service --no-pager  | wc -c
1830
```

(divide by 3 or so to get tokens.) From my perspective, returning structured output has some advantages we get extra validation, it works better if someone is trying to use the MCP in "code mode", and models are good at reading JSON - better than humans. but we *still* need to try and focus the result on the most relevant information or we are going to get bad perfomance.

[This is an advantage of guarded command execution - the model itself can figure out what information it needs and write commands to get just that.]

src/linux_mcp_server/commands.py
```diff
@@ -103,12 +105,14 @@ class CommandGroup(BaseModel):
...
             }
         ),
         "service_logs": CommandGroup(
             commands={
-                "default": CommandSpec(args=("journalctl", "-u", "{service_name}", "-n", "{lines}", "--no-pager")),
+                "default": CommandSpec(
+                    args=("journalctl", "-u", "{service_name}", "-n", "{lines}", "--no-pager", "--output=json")
```
owtaylor: See #275  - it uses `entries: list[str]` for log entries rather than JSON form, and that made sense to me - models are trained on human readable logs, and the JSON output from systemd logs have a lot of noise in them.

We don't want to expand:
```
Feb 22 15:56:09 fedora chronyd[1146]: Selected source 72.30.35.89 (2.fedora.pool.ntp.org)
```

to:

```
{
        "__SEQNUM_ID" : "9f9a06c9ff7f4d21aa0235bff47c87ad",
        "_SELINUX_CONTEXT" : "system_u:system_r:chronyd_t:s0",
        "_SYSTEMD_INVOCATION_ID" : "b62f1616c8b9453c980910f96f651610",
        "_HOSTNAME" : "fedora",
        "SYSLOG_FACILITY" : "3",
        "__REALTIME_TIMESTAMP" : "1771793899017757",
        "_COMM" : "chronyd",
        "SYSLOG_PID" : "1146",
        "_SYSTEMD_SLICE" : "system.slice",
        "_PID" : "1146",
        "SYSLOG_TIMESTAMP" : "Feb 22 15:58:19 ",
        "_BOOT_ID" : "60ef1101e7f34a23a1c486b7e72913ce",
        "__CURSOR" : "s=9f9a06c9ff7f4d21aa0235bff47c87ad;i=116ab3f;b=60ef1101e7f34a23a1c486b7e72913ce;m=728f9c938;t=64b6fe9f7561d;x=bfa5e635d89c531a",
        "_GID" : "992",
        "_EXE" : "/usr/bin/chronyd",
        "_MACHINE_ID" : "bbe3408bfdf24dae909eb91f836a87f4",
        "_CMDLINE" : "/usr/sbin/chronyd -n -F 2",
        "_RUNTIME_SCOPE" : "system",
        "_TRANSPORT" : "syslog",
        "__MONOTONIC_TIMESTAMP" : "30752229688",
        "_CAP_EFFECTIVE" : "2000400",
        "_SOURCE_REALTIME_TIMESTAMP" : "1771793899017664",
        "SYSLOG_IDENTIFIER" : "chronyd",
        "MESSAGE" : "Selected source 44.190.5.123 (2.fedora.pool.ntp.org)",
        "_UID" : "994",
        "PRIORITY" : "6",
        "_SYSTEMD_CGROUP" : "/system.slice/chronyd.service",
        "__SEQNUM" : "18262847",
        "_SYSTEMD_UNIT" : "chronyd.service"
}
```

owtaylor (changes requested):
I'm not quite sure how to to approach turning `systemctl status` into structured objects:

 A) Parse the output, turn that into objects
 B) Use lower-level API to obtain the information, reimplemen* the "summarization" of systemctl status inside our code. 
 C) Expose the low-level data sources, let the model parse through what it needs

A) is ugly and fragile. B) is ugly. C) will definitely lead to underperformance.

B) is probably best ,but considering how well trained models are on *actual* `systemctl status` output, I'm a bit skeptical about structured objects at all as applied to this tool. (It's also possible to have a tool return structured content *and* freeform content ... but I don't think that really helps things here.)

src/linux_mcp_server/tools/services.py
```diff
@@ -27,27 +26,25 @@
...
-) -> str:
+) -> list[dict[str, str]]:
     """List all systemd services.
 
     Retrieves all systemd service units with their load state, active state,
-    sub-state, and description. Also includes a count of currently running services.
+    sub-state, and description.
+
+    Returns:
+        list[dict[str, str]]: A list of dictionaries containing information about each service.
```
owtaylor, line 36: So the arguments I know of for doing the structured data conversion are:

 A) Not having to maintain fragile formatters. Accomplished, though we could just also pass through the data.
 B) Accomodating models that want to "code mode" transform the data programatically. Not accomplished - what is in this` list[dict[str, str]]?` (Maybe the model would first write a script to extract a few sample lines, but it's annoying to make them do it.)

I think we should do:

```python
class ServiceStatus(BaseModel):
    unit: str
    load: str
    active: str
    sub: str
    description: str

service_status_list_adapter = TypeAdapter(List[User])
...

return service_status_list_adapter.validate_json(stdout)
```

(systemd documents: "The list of possible LOAD, ACTIVE, and SUB states is not constant and new systemd releases may both add and remove values", so we probably shouldn't try use enums - a competent model will understand these mysterious load/active/sub values in terms of its systemd knowledge)

owtaylor (changes requested):
get_service_logs() and get_service_status() look good to me, but I have a bit of a hesititation about list_services.

PR #242, opened 2026-01-26
[RSPEED-2314] Custom accept/reject UI for run_script_modify

Jazzcort:
Here is the implementation of integrating mcp-app as a UI safeguard layer.
Please see each commit for more detail!

### DIFF
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -37,7 +37,7 @@ in read-only or read-write mode.
 
 ```
 export LINUX_MCP_TOOLSET="run_script"
-export LINUX_MCP_GATEKEEPER_MODEL="chatgpt/gpt-5.2"
+export LINUX_MCP_GATEKEEPER_MODEL="gpt-5.2"
 export OPENAI_API_KEY="Your API key"
 ```
 
@@ -57,6 +57,28 @@ the commands before running them. You must configure this model. Configuration i
 done by using the `LINUX_MCP_GATEKEEPER_MODEL` environment variable. Additional environment
 variables will be needed to configure credentials. See the
 [LiteLLM documentation](https://docs.litellm.ai/docs/providers).
+
+#### Try the mcp-app feature with compatible GUI application (i.e., Goose Desktop, Claude Desktop...etc)
+
+1. Generate the bundled app
+
+```shell
+cd mcp-app
+npm install
+npm run build:prod
+```
+
+2. Set up environment variables
+
+Please set this environment variable before launching the GUI, or add it to the environment configuration used by the application to spawn the linux-mcp-server.
+
+```
+export LINUX_MCP_USE_MCP_APPS=true
+```
+
+Note: Ensure LINUX_MCP_TOOLSET and LINUX_MCP_GATEKEEPER_MODEL are set in your environment, along with the specific API key or configuration variable required for your chosen gatekeeper model. 
+
+Once configured, `run_script_modify_interactive` will replace `run_script_modify`. This tool launches an integrated MCP app, allowing you to review, approve, or reject script modifications directly through the UI.
 
 
 ### Using with AI Agents


src/linux_mcp_server/server.py
```diff
@@ -58,8 +72,79 @@
...
+            except Exception as e:
+                logger.error(f"Failed to read UI resource from {resource_file}: {e}")
+                raise
+
+            content = TextResourceContents.model_validate(
+                {
+                    "uri": uri,
+                    "mimeType": MCP_APP_MIME_TYPE,
+                    "text": html,
+                    "_meta": {"ui": {"csp": {"resourceDomains": ["https://unpkg.com"]}}},
```
owtaylor: Do we need this - I don't see any references to unpkg.com in the built version.

owtaylor (changes requested):
So, at a high level, what I'd like to see here for our first version is:

 * we expose two different versions of run_script_modify depending on whether the client supports mcp-apps:
    1. like the current
    2. annotated to return the GUI, and with a JSON result with what the GUI needs, and with text in the description "When this tool completes successfully, this means that the user has been asked for approval - please ignore the tool call response and respond nothing to the user; the final result will be provided as a separate message later."
 * Since at least goose doesnt' implement the capabilities part of then mcp-apps spec, we allow overriding that with a LINUX_MCP_USE_MCP_APPS envvar - we can set that when launching `Goose` and it will be inherited by the MCP to avoid having to have two different `~/.config/goose/config.yaml`

I'm a little divided in my mind  on checking in run-script-app.html:

 Pro: developers that aren't changing the mcp-app directory don't need to do anything with npm
 Pro:  people can install the latest devel version with `uv tool install git+https://github.com/rhel-lightspeed/linux-mcp-server`
 Con: general problems with checking in generated content: conflicts and no verification that it matches the source

For now - when this is a separate branch - let's keep it not checked in, and we can see how annoying that is and decide what we want to do if/when we merge into `main`. If we don't switch to checking it in, we'd build it in the release workflow when pushing to pypi.

src/linux_mcp_server/server.py
Jazzcort: You're right! I was following some examples and forgot to remove this part. Will remove it!

src/linux_mcp_server/tools/run_script.py
```diff
@@ -124,11 +137,78 @@ async def run_script_readonly(
...
+        type="text",
+        text="When this tool completes successfully, this means that the user has been asked for approval - please ignore the tool call response and respond nothing to the user; the final result will be provided as a separate message later.",
+    )
+
+    return ToolResult(content=[text_block], structured_content=gatekeeper_result.model_dump())
+
+
+modify_with_mcp_apps = Tool.from_function(
+    run_script_intention_modify,
+    name="run_script_modify",
```
owtaylor: Thinking about this, want *do* want a different tool name here - the reason for this is that if the user is using both goose-desktop and goose-cli, then in the goose permission config , we want `always_approve` for `run_script_readonly` and `run_script_modify_interactive` and `ask_before` for `run_script_modify`.

.gitignore
```diff
@@ -59,3 +59,6 @@ config.sh
 
 .claude/settings.local.json
 CLAUDE.local.md
+
+# Include bundled app html file
+!mcp-app/dist
```
owtaylor: Does this need modification  - I would expect us to exclude this directory, not include it.

src/linux_mcp_server/server.py
```diff
@@ -58,8 +71,90 @@
...
+else:
+    mcp.add_tool(modify_plain)
+
+
+# This middleware can be used to dynamically inject tools based on client side compatibility
+class DynamicDiscoveryMiddleware(Middleware):
+    async def on_list_tools(self, context: MiddlewareContext, call_next):
+        tools = await call_next(context)
+
+        # Filter out the hidden tools
```
owtaylor: If I'm understanding this correctly, it needs a comment like:

```
      # Eventually, the tagging of the tools via _meta.ui.visiblity as "app" or "model" will
      # but Goose doesn't support this yet. On the other hand, goose is happy if the
      # app calls tools we don't list at all, so we just filter out the "app" tools
```

src/linux_mcp_server/tools/run_script.py
```diff
@@ -52,6 +56,15 @@ class ScriptType(StrEnum):
     + RUN_SCRIPT_COMMON_DESCRIPTION
 )
 
+RUN_SCRIPT_INTENTION_MODIFY_DESCRIPTION = (
```
owtaylor: Let's use RUN_SCRIPT_MODIFY_INTERACTIVE as the name in the source code for this (and exposed - see my comment below)

src/linux_mcp_server/tools/run_script.py
```diff
@@ -124,11 +137,78 @@ async def run_script_readonly(
 
 
 @mcp.tool(
+    tags={"run_script", "hidden_from_agent"},
+    description="This is a hidden tool. If you're an agent, PLEASE DO NO CALL THIS!",
+)
```
owtaylor: A) Is there any harm in going ahead and adding `meta={"ui": {"visiblity": "app"}}` here even though Goose doesn't support that yet?

B) Since the visibility field is app/model, let's use hidden_from_model as the name rather than hidden_from_agent

C) I don't like the description here - if the agent can see it, we have a problem - having a description here implies that we're counting on the comment to enforce the agent mot calling this tool. Maybe use `description="Execute a script; this is only available to the our mcp-app"`

src/linux_mcp_server/tools/run_script.py
```diff
@@ -124,11 +137,78 @@ async def run_script_readonly(
...
+    script: t.Annotated[
+        str,
+        Field(description="The script to run."),
+    ],
+    host: Host = None,
+) -> ToolResult:
+    gatekeeper_result = check_run_script(description, script_type, script, readonly=False)
+    text_block = TextContent(
+        type="text",
+        text="When this tool completes successfully, this means that the user has been asked for approval - please ignore the tool call response and respond nothing to the user; the final result will be provided as a separate message later.",
```
owtaylor: We've already completed successfully or failed, we should distinguish the two.


Success:
```
The user has been asked for approval; please respond nothing to the user yet; the final result will be provided as a separate message later.", 
```

Failure:
No text block at all - just let the model figure it out from the structured content?

src/linux_mcp_server/server.py
```diff
@@ -58,8 +71,90 @@
...
+
 from linux_mcp_server.tools import *  # noqa: E402, F403
 
 
+# TODO: Dynamically inject the 'modify' tool based on user compatibility.
+#
+# This is a temporary implementation. The injection logic should be moved to the
+# `on_initialize` handler in `DynamicDiscoveryMiddleware` once Goose starts
+# providing `mcp-app` compatibility during the initialize request.
+if CONFIG.use_mcp_apps:
```
owtaylor, line 142: Let's conditionally add the `execute_script` tool as well, the eventual goal would be that it only is available to clients that understand `meta.ui.visiblity`.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -52,6 +56,15 @@ class ScriptType(StrEnum):
     + RUN_SCRIPT_COMMON_DESCRIPTION
 )
 
+RUN_SCRIPT_INTENTION_MODIFY_DESCRIPTION = (
+    """
+Propose a script that modifies the system to users and wait for the approval. 
+This tool uses a gatekepper to diagnose the purposed script and provides the 
+user with an insight analysis about the script.
+"""
```
owtaylor, line 63: This is probably not useful to the LLM (and the gatekeeper is run for readonly as well) Suggest:

```
Run a script that modifies the system. The user will be asked for approval interactively.
```

mcp-app/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@modelcontextprotocol/ext-apps": "^0.2.2",
+    "@modelcontextprotocol/sdk": "^1.24.0",
+    "react": "^19.2.0",
+    "react-dom": "^19.2.0",
+    "zod": "^4.1.13"
+  },
+  "devDependencies": {
+    "@tailwindcss/postcss": "^4.1.18",
+    "@types/cors": "^2.8.19",
+    "@types/express": "^5.0.0",
```
owtaylor: Why?

mcp-app/package.json
```diff
@@ -0,0 +1,34 @@
...
+    "@types/node": "^22.0.0",
+    "@types/react": "^19.2.2",
+    "@types/react-dom": "^19.2.2",
+    "@vitejs/plugin-react": "^4.3.4",
+    "autoprefixer": "^10.4.23",
+    "cors": "^2.8.5",
```
owtaylor: Is this needed?

mcp-app/src/run-script-app.tsx
```diff
@@ -0,0 +1,247 @@
...
+        updatedExecutionState = "failed";
+      } else {
+        updatedExecutionState = "success";
+      }
+
+      updatedExecutionResult = extractText(result);
+    } catch (e) {
+      updatedExecutionState = "failed";
+      updatedExecutionResult = e as string;
+      console.error(`callServerTool failed: ${e}`);
```
owtaylor: console.error("calledServerTool failed", e);  is going to be better because it doesn't hide the error from the developer console.

.gitignore
Jazzcort: Thanks for the catch! Will remove it.

src/linux_mcp_server/server.py
Jazzcort: Will add this comment!

src/linux_mcp_server/tools/run_script.py
Jazzcort: Will change!

src/linux_mcp_server/tools/run_script.py
Jazzcort: I think there should be no harm to add `meta={"ui": {"visiblity": "app"}}`, but I will double check it so it will not crash. 😁 
I totally agree with all the renaming and description changes. I'm always bad at naming variables. will keep learning. 😁

src/linux_mcp_server/tools/run_script.py
Jazzcort: >No text block at all - just let the model figure it out from the structured content?

We can try this and see if the model can figure it out.

mcp-app/src/run-script-app.tsx
Jazzcort: Will change!

mcp-app/package.json
Jazzcort: I removed express but forgot the type package. Thanks for the catch!!

mcp-app/package.json
Jazzcort: I think we don't need it. Will double check!

PR #245, opened 2026-01-27
refactor(utils): extract sort_entries() to reduce duplication in formatters

major:
- Add `sort_entries()` function to `utils/format.py` for sorting `NodeEntry` lists by size, modified time, or name
- Update `format_directory_listing()` in `formatters.py` to use extracted function
- Update `format_file_listing()` in `formatters.py` to use extracted function
- Update `_list_resources()` in `storage.py` to use extracted function
- Add parameterized tests for `sort_entries()` covering all sort keys and directions

### DIFF
--- a/src/linux_mcp_server/tools/storage.py
+++ b/src/linux_mcp_server/tools/storage.py
@@ -4,6 +4,7 @@ import os
 import typing as t
 
 from collections.abc import Mapping
+from operator import attrgetter
 from pathlib import Path
 
 from fastmcp.exceptions import ToolError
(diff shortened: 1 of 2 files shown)


src/linux_mcp_server/tools/storage.py
```diff
@@ -94,13 +95,7 @@ async def _list_resources(
...
         entries = parser(stdout, order_by)
 
-        # Sort entries based on order_by criterion and sort direction
-        if order_by == OrderBy.SIZE:
-            entries = sorted(entries, key=lambda e: e.size, reverse=sort == SortBy.DESCENDING)
-        elif order_by == OrderBy.MODIFIED:
-            entries = sorted(entries, key=lambda e: e.modified, reverse=sort == SortBy.DESCENDING)
-        else:
-            entries = sorted(entries, key=lambda e: e.name.lower(), reverse=sort == SortBy.DESCENDING)
+        entries = sort_entries(entries, order_by, sort == SortBy.DESCENDING)
```
samdoran: This doesn't really need to be a standalone function. The existing code can be simplified into one or two lines.
```python
reverse = sort == SortBy.DESCENDING
entries = sorted(entries, key=attrgetter(order_by), reverse=reverse)
```

major: `attrgetter` works okay for size/modified, but it doesn't work well for names with case-insensitive name sorting. Let me see if I can find a happy medium.

samdoran: In that case, a custom sorting function would be better. Something like [this](https://github.com/rhel-lightspeed/linux-mcp-server/commit/2b3c86cad164a20e68b67a29d74e9fbc1f4768a0). I updated the tests to cover that as well.

PR #246, opened 2026-01-27
Adding CLI to linux-mcp-server to make it modular

### LINKED ISSUE #173: [RFE] Add the ability to remotely connect to the MCP server
I was wondering how much development it would be to add the feature for the mcp server to be be able to be accessed over http / https like you can with other MCP tools like FastMCP. The Idea is to host the MCP servers as a container or on their own system then have the AI agent on something like N8N connect to the MCP server remotely to access the tools.

I am open to discussing this more and helping provide PR's as need but wanted to see if this is something worth doing.

aheath1992:
Summary of Changes

1. Modified src/linux_mcp_server/__main__.py

Added command-line argument parsing with the following options:
- --transport: Choose between stdio (default), sse, http, or streamable-http
- --host: Host address to bind to (default: 127.0.0.1) for HTTP/SSE transports
- --port: Port to bind to (default: 8000) for HTTP/SSE transports
- --path: Optional endpoint path for HTTP/SSE transports
- --log-level: Log level for HTTP/SSE transports
- --show-banner: Show the FastMCP server banner on startup
- --version: Display version information

2. Modified src/linux_mcp_server/server.py

Updated the main() function to accept transport parameters and pass them to mcp.run().

Usage Examples

Default stdio transport (current behavior):
linux-mcp-server

SSE transport:
linux-mcp-server --transport sse --host 0.0.0.0 --port 8080

HTTP transport with custom path:
linux-mcp-server --transport streamable-http --host localhost --port 3000 --path /mcp

With banner and debug logging:
linux-mcp-server --transport sse --show-banner --log-level debug

The implementation uses argparse (built-in to Python) so no additional dependencies are needed.

Fixes #173

### DIFF
--- a/src/linux_mcp_server/utils/enum.py
+++ b/src/linux_mcp_server/utils/enum.py
@@ -1,29 +1,16 @@
-import sys
-
 from enum import Enum
 
 
-# StrEnum is available in Python 3.11+
-if sys.version_info >= (3, 11):
-    from enum import StrEnum as _StrEnumBase
-else:
-    _StrEnumBase = None  # type: ignore[assignment,misc]
+try:
+    from enum import StrEnum  # pyright: ignore[reportAttributeAccessIssue]
+except ImportError:
+    StrEnum = None
 
 
 class StringEnum(str, Enum):
-    """String-based enum for Python < 3.11 compatibility."""
-
-    def __str__(self) -> str:
+    def __str__(self):
         return self.value
 
 
-# Use native StrEnum if available, otherwise use our StringEnum
-StrEnum = _StrEnumBase if _StrEnumBase is not None else StringEnum  # type: ignore[assignment,misc]
-
-
-class TransportType(StrEnum):  # type: ignore[misc,valid-type]
-    """Transport protocol types for the MCP server."""
-
-    STDIO = "stdio"
-    HTTP = "http"
-    STREAMABLE_HTTP = "streamable-http"
+if StrEnum is None:
+    StrEnum = StringEnum
--- a/tests/test__main__.py
+++ b/tests/test__main__.py
@@ -1,4 +1,0 @@
-"""Tests for the __main__ module (legacy file, most tests are in test_cli.py)."""
-
-# This file is kept for backwards compatibility but tests have moved to test_cli.py
-# The old tests relied on Typer's 'app' which no longer exists


src/linux_mcp_server/server.py
```diff
@@ -39,5 +42,16 @@
 from linux_mcp_server.tools import *  # noqa: E402, F403
 
 
-def main():
-    mcp.run(show_banner=False)
+def main(
+    transport: Literal["stdio", "http", "sse", "streamable-http"] = "stdio",
```
samdoran: This could be a string enum to reduce the likelihood of typo errors. Enums also integrate nicely with configuration validation and remove the need for using a `Literal` type hint in multiple places.

src/linux_mcp_server/utils/enum.py
```diff
@@ -4,7 +4,7 @@
 try:
     from enum import StrEnum  # pyright: ignore[reportAttributeAccessIssue]
 except ImportError:
-    StrEnum = None
+    StrEnum = None  # type: ignore[assignment,misc]
```
samdoran: Why are these type ignore statements necessary?

src/linux_mcp_server/utils/enum.py
```diff
@@ -13,4 +13,13 @@ def __str__(self):
...
 if StrEnum is None:
-    StrEnum = StringEnum
+    StrEnum = StringEnum  # type: ignore[assignment,misc]
+
+
+class TransportType(StrEnum):  # type: ignore[misc]
+    """Transport protocol types for the MCP server."""
+
+    STDIO = "stdio"
+    SSE = "sse"
```
samdoran: Since SSE is deprecated, we don't want to support that transport type.

src/linux_mcp_server/__main__.py
```diff
@@ -3,21 +3,89 @@
 import logging
 import sys
 
+from typing import Literal
+from typing import Optional
```
samdoran: Import typing once in order to avoid the need to import types individually.
```suggestion
import typing as t
```

src/linux_mcp_server/__main__.py
```diff
@@ -3,21 +3,89 @@
...
+
+import typer
```
samdoran: Is it necessary to turn this into a Typer application only to add command line parameters? Enabling command line parsing on the settings class should be all that is necessary. It's also not clear how these command line parameters align with the current environment variable based configuration options.

src/linux_mcp_server/__main__.py
```diff
@@ -3,21 +3,89 @@
...
 
+    # Prepare transport kwargs based on transport type
+    transport_kwargs = {}
+    if transport in {TransportType.HTTP, TransportType.SSE, TransportType.STREAMABLE_HTTP}:
+        transport_kwargs["host"] = host
+        transport_kwargs["port"] = port
+        if path:
+            transport_kwargs["path"] = path
+        if log_level:
+            transport_kwargs["log_level"] = log_level
```
samdoran: This is unnecessary since this is being done by the configuration class.

src/linux_mcp_server/config.py
```diff
@@ -1,20 +1,43 @@
 """Settings for linux-mcp-server"""
 
+import sys
+
 from pathlib import Path
+from typing import Any
```
samdoran, line 6:
```suggestion
import typing as t
```

src/linux_mcp_server/config.py
```diff
@@ -1,20 +1,43 @@
...
 
+from linux_mcp_server.utils.enum import StrEnum
 from linux_mcp_server.utils.types import UpperCase
 
 
-class Config(BaseSettings):
+class Transport(StrEnum):
+    stdio = "stdio"
+    http = "http"
+    streamable_http = "streamable-http"
```
samdoran, line 19: This seems like a duplicate and inconsistent implementation of the `TransportType` enum from `linux_mcp_server.utils.enum`. Only one should be used.

src/linux_mcp_server/server.py
```diff
@@ -2,8 +2,12 @@
 
 import logging
 
+from typing import Any
```
samdoran:
```suggestion
import typing as t
```

src/linux_mcp_server/server.py
```diff
@@ -39,5 +43,14 @@
...
+    """Run the MCP server with specified transport and options.
+
+    Args:
+        transport: Transport protocol to use (stdio, sse, http, streamable-http)
+        show_banner: Whether to show the FastMCP banner on startup
+        **kwargs: Additional transport-specific keyword arguments
+    """
+    # Use provided transport or fall back to CONFIG
+    actual_transport = transport if transport is not None else CONFIG.transport.value
+    mcp.run(show_banner=show_banner, transport=actual_transport, **kwargs)  # type: ignore[arg-type]
```
samdoran: The default value for `transport` can be used from config making this check unnecessary. The value for `transport` should not be `None` would would remove the need for the type ignore.
```suggestion
    mcp.run(show_banner=show_banner, transport=CONFIG.transport.value, **kwargs)
```

src/linux_mcp_server/config.py
```diff
@@ -1,20 +1,43 @@
...
+# Only enable CLI parsing when not running under pytest
+_enable_cli_parsing = "pytest" not in sys.modules
+
+
+class Config(
+    BaseSettings,
+    cli_parse_args=_enable_cli_parsing,
+    cli_implicit_flags=_enable_cli_parsing,
+    cli_kebab_case=_enable_cli_parsing,
+):
```
samdoran, line 31: This seems odd. Is there a way to change the tests so that the runtime code doesn't require this global?

src/linux_mcp_server/config.py
```diff
@@ -38,6 +38,7 @@ class Config(
     host: str | None = None
     port: int | None = None
     path: str | None = None
+    show_banner: bool = False
```
samdoran: This is an internal option that does not need to be exposed as a user-configurable setting.

src/linux_mcp_server/utils/enum.py
```diff
@@ -1,16 +1,29 @@
...
 
-try:
-    from enum import StrEnum  # pyright: ignore[reportAttributeAccessIssue]
-except ImportError:
-    StrEnum = None
+# StrEnum is available in Python 3.11+
+if sys.version_info >= (3, 11):
+    from enum import StrEnum as _StrEnumBase
+else:
+    _StrEnumBase = None  # type: ignore[assignment,misc]
```
samdoran: Refactoring this code is unrelated to adding command line options and should be discussed in another PR.

src/linux_mcp_server/server.py
```diff
@@ -39,5 +41,7 @@
 from linux_mcp_server.tools import *  # noqa: E402, F403
 
 
-def main():
-    mcp.run(show_banner=False)
+def main() -> None:
+    """Run the MCP server with configuration from CONFIG."""
+    # show_banner defaults to False in FastMCP
+    mcp.run(transport=CONFIG.transport.value, show_banner=False, **CONFIG.transport_kwargs)  # type: ignore[arg-type]
```
samdoran, line 47: Is the type ignore needed here?

tests/test__main__.py
```diff
@@ -1,14 +1,4 @@
...
-    with pytest.raises(SystemExit):
-        cli()
-
-
-def test_cli_keyboard_interrupt(mocker):
-    mocker.patch("linux_mcp_server.__main__.main", side_effect=KeyboardInterrupt)
-    with pytest.raises(SystemExit):
-        cli()
+# This file is kept for backwards compatibility but tests have moved to test_cli.py
+# The old tests relied on Typer's 'app' which no longer exists
```
samdoran: If this file serves no purpose it should be removed.

src/linux_mcp_server/server.py
aheath1992: yes, mcp.run() has transport-specific typing, and mypy can’t verify that CONFIG.transport.value matches the right overload or that CONFIG.transport_kwargs are valid for it. The code is runtime-safe, but the types aren’t provable.

PR #252, opened 2026-01-29
Conditionally use a Pydantic model as a response format

subpop:
If the LLM model supports the "response_format" field and a response schema, include it in the completion function call. Otherwise, fall back to None for a text response.


owtaylor (approved):
Looks great!

PR #253, opened 2026-01-29
Dynamically construct instructions

subpop:
Depending on the value of CONFIG.toolset, include instructions for the read-only, or writable tools.

### DIFF
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -29,6 +29,9 @@ These tools map to six areas:
 - **Remote execution:** Every tool accepts an optional `host` argument. When set, the work runs on that host over SSH instead of locally.
 - **Containers:** If the `container` environment variable is set, tools refuse to run locally; a remote `host` must be used.
 - **Read-only vs destructive:** All tools are marked read-only. Do not expect to be able to modify the system.
+- **Log file access:** requires explicit allowlist configuration via LINUX_MCP_ALLOWED_LOG_PATHS
+- **Service names:** automatically append '.service' suffix if not provided
+- **File paths:** must be absolute
 """
 
 INSTRUCTIONS_RUN_SCRIPT = """You have access to tools that execute Python or Bash scripts you supply on the target system, for inspection or for making changes.
@@ -52,6 +55,9 @@ INSTRUCTIONS_RUN_SCRIPT = """You have access to tools that execute Python or Bas
 - **Remote execution:** Every tool accepts an optional `host` argument. When set, the work runs on that host over SSH instead of locally.
 - **Containers:** If the `container` environment variable is set, tools refuse to run locally; a remote `host` must be used.
 - **Read-only vs destructive:** run_script_readonly is marked read-only; the modify script tool is marked destructive.
+- **Log file access:** requires explicit allowlist configuration via LINUX_MCP_ALLOWED_LOG_PATHS
+- **Service names:** automatically append '.service' suffix if not provided
+- **File paths:** must be absolute
 """
 
 INSTRUCTIONS_BOTH = """You have access to two kinds of tools: predefined commands that inspect the system, and script runners that execute Python or Bash you supply.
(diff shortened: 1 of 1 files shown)


src/linux_mcp_server/server.py
```diff
@@ -11,6 +11,47 @@
...
+
+If running inside a container, which is indicated by the the environment variable
+"container", local execution is not allowed.
+
+IMPORTANT NOTES:
+    - Most tools are read-only operations (indicated by readOnlyHint)
+    - Log file access requires explicit allowlist configuration via LINUX_MCP_ALLOWED_LOG_PATHS
+    - Service names automatically append '.service' suffix if not provided
+    - File paths must be absolute
+    - Some hardware info may require elevated privileges
```
owtaylor: I feel these instructions are liable to confuse things in the RUN_SCRIPT case. Maybe  we can do it as:

```
FIXED_INSTRUCTIONS = <old instructions>
RUN_SCRIPT_INSTRUCTIONS = <as you have it here>
BOTH_INSTRUCTIONS = """\
This server provides two sets of tools - a set of predefined tools for inspecting
a server, and flexible tools for running bash and Python scripts: 

## Predefined Tools
""" + \
FIXED INSTRUCTIONS + \
"""
## Running Scripts
""" + \
RUN_SCRIPT_INSTRUCTIONS
```

That doesn't handle "All tools support optional 'host' parameter for SSH-based remote diagnostics.
When provided, commands execute on the remote system instead of locally." correctly, but I think that part should be pretty clear from the individual parameter descriptions.

subpop: I attempted to rewrite the instructions using a 3-way approach like you suggested. I *think* it's better.

src/linux_mcp_server/server.py
```diff
@@ -32,29 +87,22 @@
...
+            else (RUN_SCRIPT_INTRO if CONFIG.toolset == Toolset.RUN_SCRIPT else "")
+        ),
+        instructions=BOTH_INSTRUCTIONS
+        if CONFIG.toolset == Toolset.BOTH
+        else (
+            FIXED_INSTRUCTIONS
+            if CONFIG.toolset == Toolset.FIXED
+            else (RUN_SCRIPT_INSTRUCTIONS if CONFIG.toolset == Toolset.RUN_SCRIPT else "")
+        ),
+    ),
```
owtaylor: I find that hard to read, what about something like:

```
introduction, instructions = {
    Toolset.BOTH: (BOTH_INTRO, BOTH_INSTRUCTIONS),
    Toolset.FIXED: (FIXED_INTRO, FIXED_INSTRUCTIONS),
    Toolset.RUN_SCRIPT: (RUN_SCRIPT_INTRO, RUN_SCRIPT_INSTRUCTIONS),
}[CONFIG.toolset]

# Initialize FastMCP server
mcp = FastMCP(
    "linux-diagnostics",
    instructions=INSTRUCTIONS.format(
        introduction=introduction,
        instructions=instructions,
    ),
    **kwargs,
)
```

(I also probably would name `BOTH_INSTRUCTIONS` to `BOTH_DETAILS` and so forth)

src/linux_mcp_server/server.py
owtaylor: It was actually the IMPORTANT_NOTES at the end that I was wondering about - it would confuse me as a human reader if those appeared in the RUN_SCRIPT case, so potentially will confuse a model as well. What if we simply moved those into FIXED_INSTRUCTIONS?

owtaylor (changes requested):

src/linux_mcp_server/server.py
subpop: Oh, I like the dictionary approach. I'll refactor.

src/linux_mcp_server/server.py
subpop: Good point. When running with `RUN_SCRIPT`, the tools themselves define whether they are modify or read-only. And in the `FIXED` case, all the tools are read only; not "some". I also noticed we still include that note about "some tools may require elevated privileges". That won't work in `FXED` mode at all.

src/linux_mcp_server/server.py
```diff
@@ -29,34 +108,7 @@
...
-
-        If running inside a container, which is indicated by the the environment variable
-        "container", local execution is not allowed.
-
-        IMPORTANT NOTES:
-            - Most tools are read-only operations (indicated by readOnlyHint)
-            - Log file access requires explicit allowlist configuration via LINUX_MCP_ALLOWED_LOG_PATHS
-            - Service names automatically append '.service' suffix if not provided
-            - File paths must be absolute
-            - Some hardware info may require elevated privileges
```
owtaylor, line 56: The last few items here are no longer in the instructions. Were they not needed?

subpop: Ah good catch. I had forgotten about that list when I rewrote the instructions.

PR #256, opened 2026-01-30
[RSPEED-2335] Wrap executed commands in sudo+systemd-run where possible

subpop:
The readonly tool additionally includes flags that deny access to AF_INET and AF_INET6 address families and makes the root filesystem appear readonly.

The command is wrapped in a bash script that checks for sudo and systemd-run and gracefully degrades to running the script directly if either program is absent.

### DIFF
--- a/src/linux_mcp_server/tools/run_script.py
+++ b/src/linux_mcp_server/tools/run_script.py
@@ -37,9 +37,9 @@ SYSTEMD_RUN_ARGS = [
     "--wait",
     "--service-type=exec",
     "--property=PrivateTmp=true",
+    "--property=NoNewPrivileges=true",
 ]
 SYSTEMD_RUN_READONLY_ARGS = [
-    "--property=NoNewPrivileges=true",
     "--property=ReadOnlyPaths=/",
     "--property=RestrictAddressFamilies=AF_UNIX",
 ]
@@ -50,7 +50,7 @@ SYSTEMD_RUN_COMMAND = "/usr/bin/sudo /usr/bin/systemd-run {args}"
 WRAPPER_TEMPLATE = """\
 set -euo pipefail
 SCRIPT={script}
-if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo whoami >/dev/null 2>&1; then
+if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo -l whoami >/dev/null 2>&1; then
   exec {systemd_run_command} {script_type} -c "$SCRIPT"
 else
   exec {script_type} -c "$SCRIPT"


src/linux_mcp_server/tools/run_script.py
```diff
@@ -85,15 +115,12 @@ async def run_script_readonly(
...
-    elif script_type == ScriptType.BASH:
-        command = ["bash", "-c", script]
+    wrapper_script = WRAPPER_TEMPLATE.format(
+        systemd_run_command=SYSTEMD_RUN_COMMAND.format(args=" ".join(SYSTEMD_RUN_ARGS + SYSTEMD_RUN_READONLY_ARGS)),
+        script_type=script_type.value,
+        script=shlex.quote((BASH_STRICT_PREAMBLE + script) if script_type == ScriptType.BASH else script),
+    )
+    command = ["bash", "-c", wrapper_script]
 
     gatekeeper_result = check_run_script(description, script_type, script, readonly=True)
```
owtaylor: It was intentional that the `set -euo pipefail` was included in what was passed to the gatekeeper - it should be able to do a better job if it knows the script is being run with those parameters.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -26,11 +27,40 @@ class ScriptType(StrEnum):
...
+    "--property=RestrictAddressFamilies=AF_UNIX",
+]
+SYSTEMD_RUN_COMMAND = "/usr/bin/sudo /usr/bin/systemd-run {args}"
+
+# Wrapper for run_script_readonly: use sudo+systemd-run when available, else run script
+# directly. Template uses {script_type} and {script}; script is escaped via shlex.quote().
+WRAPPER_TEMPLATE = """\
+set -euo pipefail
+if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo whoami >/dev/null 2>&1; then
+  exec {systemd_run_command} --uid=$USER {script_type} -c {script}
```
owtaylor: Instead of having {script} here twice, which could be painful in the logs, I'd do:
```sh
SCRIPT={script}
[...]
  exec {systemd_run_command} --uid=$USER {script_type} -c "$SCRIPT"
```
Another approach that avoids `shlex.quote` would be to do:
```sh
  exec {systemd_run_command} --uid=$USER {script_type} -c "$1"
```

and run it as:
```python
command = ["bash", "-c", wrapper_script, "linux-mcp-server-wrapper", script]
```

They seem about the same level of OK to me.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -149,15 +176,12 @@ async def run_script_modify(
...
-    command = []
-
-    if script_type == ScriptType.BASH:
-        script = "# added by linux-mcp-server\nset -euo pipefail\n\n" + script
-
-    if script_type == ScriptType.PYTHON:
-        command = ["python3", "-c", script]
-    elif script_type == ScriptType.BASH:
-        command = ["bash", "-c", script]
+    wrapper_script = WRAPPER_TEMPLATE.format(
```
owtaylor: I see I had existing ugly duplication - instead of perpetuating that, let's pull this out into a helper.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -149,15 +187,7 @@ async def run_script_modify(
...
-    if script_type == ScriptType.BASH:
-        script = "# added by linux-mcp-server\nset -euo pipefail\n\n" + script
-
-    if script_type == ScriptType.PYTHON:
-        command = ["python3", "-c", script]
-    elif script_type == ScriptType.BASH:
-        command = ["bash", "-c", script]
+    command = _wrap_script(script_type, script)
 
     gatekeeper_result = check_run_script(description, script_type, script, readonly=False)
```
owtaylor: The preamble is also needed here

src/linux_mcp_server/tools/run_script.py
```diff
@@ -26,11 +27,41 @@ class ScriptType(StrEnum):
     BASH = "bash"
 
 
+BASH_STRICT_PREAMBLE = "set -euo pipefail; "
+
+SYSTEMD_RUN_ARGS = [
+    "--quiet",
+    "--pipe",
+    "--same-dir",
```
owtaylor: Does `--same-dir` make sense if running as root?
Maximally restrictive would be `--working-directory=/tmp --property=PrivateTmp=true`

src/linux_mcp_server/tools/run_script.py
```diff
@@ -26,11 +27,41 @@ class ScriptType(StrEnum):
...
+    "--property=RestrictAddressFamilies=AF_UNIX",
+]
+SYSTEMD_RUN_COMMAND = "/usr/bin/sudo /usr/bin/systemd-run {args}"
+
+# Wrapper for run_script_readonly: use sudo+systemd-run when available, else run script
+# directly. Template uses {script_type} and {script}; script is escaped via shlex.quote().
+WRAPPER_TEMPLATE = """\
+set -euo pipefail
+SCRIPT={script}
+if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo whoami >/dev/null 2>&1; then
```
owtaylor: `sudo -l whoami` would avoid audit-log spam.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -26,11 +27,41 @@ class ScriptType(StrEnum):
...
+SYSTEMD_RUN_ARGS = [
+    "--quiet",
+    "--pipe",
+    "--same-dir",
+    "--collect",
+    "--wait",
+    "--service-type=exec",
+]
+SYSTEMD_RUN_READONLY_ARGS = [
+    "--property=NoNewPrivileges=true",
```
owtaylor: If it doesn't break anything, I'd move this to the main args - since we're running as root with full capabilities for both sandboxes, it doesn't make any real difference, but the model I have mentally is "sandbox is fixed, can't privilege-escalate from it"

src/linux_mcp_server/tools/run_script.py
subpop: Because it's running over SSH, it's going to end up being $HOME. I suppose it's harmless, but at the same time is unnecessary. Do we want to be maximally restrictive? If a script wrote files, if we set the working directory to a private temp, there's less risk of the script spewing files around permanent locations like `/root`.

owtaylor: I don't think we want the llm doing things like running one command to write to a temp file and then running a separate command to read from it - so it seems to me that starting from the PrivateTmp version should be fine.

PR #260, opened 2026-02-04
[RSPEED-2364] Enable HTTP transport options

### LINKED ISSUE #176: [RFE] Add the ability to remotely connect to the MCP server
I was wondering how much development it would be to add the feature for the mcp server to be be able to be accessed over http / https like you can with other MCP tools like FastMCP. The Idea is to host the MCP servers as a container or on their own system then have the AI agent on something like N8N connect to the MCP server remotely to access the tools.

I am open to discussing this more and helping provide PR's as need but wanted to see if this is something worth doing.

samdoran:
Add a new option to enable HTTP and streamable HTTP transports as well as options for host, port, and URL path.

Running with either of the HTTP transports does mean the server needs to be started before the LLM client since it expects the MCP server to be available at the specified host, port, and path.

```
> linux-mcp-server --transport streamable-http --port 8123 --path /my-mcp
2026-02-04 16:18:59 | INFO | root | Logging initialized: /Users/sdoran/.local/share/linux-mcp-server/logs
2026-02-04 16:18:59 | INFO | linux-mcp-server | Running Linux MCP Server 1.2.2.dev24+gbc80a46ba.d20260127. Press Ctrl+C, Enter to stop the server.
[02/04/26 16:18:59] INFO     Starting MCP server 'linux-diagnostics' with transport 'streamable-http' on http://127.0.0.1:8123/my-mcp                                        server.py:2580
INFO:     Started server process [63800]
INFO:     Waiting for application startup.
2026-02-04 16:18:59 | INFO | docket.worker | Starting worker 'rhmbp-2021#63800' with the following tasks:
2026-02-04 16:18:59 | INFO | docket.worker | * trace(message: str, ...)
2026-02-04 16:18:59 | INFO | docket.worker | * fail(message: str, ...)
2026-02-04 16:18:59 | INFO | docket.worker | * sleep(seconds: float, ...)
2026-02-04 16:18:59 | INFO | mcp.server.streamable_http_manager | StreamableHTTP session manager started
INFO:     Application startup complete.
INFO:     Uvicorn running on http://127.0.0.1:8123 (Press CTRL+C to quit)
```

Update documentation to reflect new options.

Closes #246.
Fixes #176

### DIFF
--- a/src/linux_mcp_server/config.py
+++ b/src/linux_mcp_server/config.py
@@ -27,7 +27,7 @@ class Config(BaseSettings):
         cli_hide_none_type=True,
         # Only ignore errors for incorrect/extra parameters when testing
         # https://github.com/pydantic/pydantic-settings/issues/391
-        cli_ignore_unknown_args="pytest" in sys.argv[0],
+        cli_ignore_unknown_args=sys.argv[0].endswith("pytest"),
         cli_implicit_flags=True,
         cli_kebab_case=True,
         cli_parse_args=True,


src/linux_mcp_server/config.py
```diff
@@ -6,28 +6,40 @@
...
     )
 
+    # FIXME: When the next version of pydantic-settings is released, change this
+    # to CliToggleFlag in order to remove the '--no-' option.
+    # https://github.com/pydantic/pydantic-settings/pull/717/changes
     version: bool = False
 
     user: str = ""
+    transport: Transport = Transport.stdio
+    host: str | None = None
```
owtaylor: Hmm, so if we have it this way, then the default of host isn't 127.0.0.1 - it will default to what is set in FASTMCP_HOST and if that isn't set, then it (assuming that fastmcp isn't changed) defaults to 127.0.0.1. It seems like it would be better and more predictable to just have the defaults that we document hardcoded here and avoid the inheritance.

src/linux_mcp_server/config.py
```diff
@@ -54,5 +66,16 @@ def effective_known_hosts_path(self) -> Path:
...
         return self.known_hosts_path or Path.home() / ".ssh" / "known_hosts"
 
+    @property
+    def transport_kwargs(self):
+        result = {}
+        if self.transport in {Transport.http, Transport.streamable_http}:
+            result["host"] = self.host
+            result["port"] = self.port
+            result["path"] = self.path
+            result["log_level"] = self.log_level
```
owtaylor: Hmm, log_level as a kwargs is supported by stdio transport_kwargs too ... if there's a reason that we only pass it for the http transports, that should be documented.

src/linux_mcp_server/config.py
```diff
@@ -6,28 +6,40 @@
...
+    streamable_http = "streamable-http"
+
+
+class Config(BaseSettings):
+    # The '_' is required in the env_prefix, otherwise, pydantic would
+    # interpret the prefix as LINUX_MCPLOG_DIR, instead of LINUX_MCP_LOG_DIR
     model_config = SettingsConfigDict(
         env_prefix="LINUX_MCP_",
         env_ignore_empty=True,
+        cli_exit_on_error=False,  # Ignore errors for incorrect/extra parameters
```
owtaylor: Why? That can lead to much user frustration.

src/linux_mcp_server/config.py
samdoran: I wasn't sure which was the better approach: using the underlying defaults are hard coding default values. I like the idea of being explicit in our configuration.

src/linux_mcp_server/config.py
samdoran: That's an oversight. We can always pass `log_level` through if it's valid for all the transport types.

src/linux_mcp_server/config.py
samdoran: I did put those values in the documentation, so it seems to make more sense to explicitly add it to your config.

src/linux_mcp_server/config.py
samdoran: Actually, this is not the correct option. That comment should actually be on `cli_ignore_unknown_args=True`. This is why I love code review!

I do think it's better to error for invalid options but here's the problem: when running with `pytest`, we end up with two things parsing `sys.argv`. When Pydantic Settings parses `sys.argv` it will error on the options passed to `pytest` and the message is quite confusing.


```
> pytest --cov-report xml
usage: pytest [-h] [--version | --no-version] [--user str]
              [--transport {stdio,http,streamable_http}] [--host str]
              [--port int] [--path str] [--log-dir Path] [--log-level str]
              [--log-retention-days int] [--allowed-log-paths str]
              [--ssh-key-path Path] [--key-passphrase SecretStr]
              [--search-for-ssh-key | --no-search-for-ssh-key]
              [--verify-host-keys | --no-verify-host-keys]
              [--known-hosts-path Path] [--command-timeout int]
pytest: error: unrecognized arguments: --cov-report xml
```

I looked into a lot of different ways to solve this. There is [quite a bit of discussion](https://github.com/pydantic/pydantic-settings/issues/391) around this same issue. I didn't really like the [solution](https://github.com/pydantic/pydantic-settings/issues/391#issuecomment-2349150941), but it may be the best option.

I suppose the main question is how much do we dislike `cli_ignore_unknown_args="pytest" in sys.argv[0]`? Maybe it's ok if it avoids user frustration.

samdoran: Maybe `cli_ignore_unknown_args=sys.argv[0].endswith("pytest")` is less error prone.

samdoran: I pushed some changes. Let me know what you think.

owtaylor (approved):
I like the approach of checking if argv[0] ends with "pytest" - it's explicit about what problem it is handling and won't leak into causing problems for users who accidentally typo an argument name.

Generally seems ready to land.

PR #263, opened 2026-02-07
feat(network): add ip route tool

Albe83:
## Summary
Adds a new read-only MCP tool to fetch routing tables via `ip route` (IPv4/IPv6), with unit tests and documentation updates. Also adjusts a storage test to avoid false negatives when running as root.

## Motivation
Routing tables are a core diagnostic for network issues. Today the server exposes interfaces, connections, and listening ports but no route visibility. This tool completes the basic network triage set without introducing any write capability.

## Changes
- **New tool:** `get_ip_route_table`
  - Parameters: `family` = `ipv4` | `ipv6` | `all` (default: `ipv4`)
  - Uses allowlisted `ip route show` commands in `commands.py`
  - Read-only, `@log_tool_call`, `@disallow_local_execution_in_containers`, `readOnlyHint=True`
  - Returns a labeled output block per family
- **Commands registry:** added `ip_route` group with `ipv4` and `ipv6`
- **Tests:** cover success, `all` aggregation, **partial failure** (one family fails), command failure/empty output, and error propagation
- **Docs:** usage, cheatsheet, API index updated with the new tool
- **Test fix:** skip restricted-path storage test when `geteuid()==0` to prevent false failures in root environments

## Security / Read-only
- No user-supplied shell args; commands are fixed in `commands.py`
- Tool annotated as read-only; no writes or side effects

## Testing
- `make verify`

## Notes for Reviewer
- The `family=all` path runs both commands and returns a combined output, with per-family error messages if one fails.
- The root-skip change is deliberately small and avoids masking behavior on non-root runs.

### DIFF
--- a/.github/workflows/build-publish.yml
+++ b/.github/workflows/build-publish.yml
@@ -92,3 +92,87 @@ jobs:
 
       - name: Publish to MCP Registry
         run: ./scripts/publish_mcp_registry.py
+
+  create-release:
+    name: Create GitHub Release
+    runs-on: ubuntu-latest
+
+    permissions:
+      contents: write
+
+    steps:
+      - name: Create GitHub Release
+        uses: softprops/action-gh-release@v2.5.0
+        with:
+          tag_name: ${{ github.ref_name }}
+          generate_release_notes: true
+
+  zipapp:
+    name: Zipapp - ${{ matrix.python-version }} ${{ matrix.image.family }} ${{ matrix.image.arch }}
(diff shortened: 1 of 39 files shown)


subpop (changes requested):
There's a lot more to this PR than just adding `ip route` commands. Can you separate the `dnf` tools into another PR? I think the idea of adding an `ip route` tool is a good one, but I don't want to get it mixed up with other tools.

PR #264, opened 2026-02-07
feat(dnf): add read-only dnf tools and list output limits

Albe83:
## Summary
Adds read-only DNF tooling to the Linux MCP Server, enabling system diagnostics for RPM-based (RH-like) systems with DNF v4+. Includes list/detail workflows, repositories, groups and modules, plus safe output limiting for large datasets.

## Changes
- Add DNF tools: list installed/available packages, package info, repo list/info, provides, groups, modules
- Enforce read-only execution and input validation
- Add list output controls (`limit`, `offset`, `no_limit`) with safe defaults and truncation note
- Update usage documentation
- Add comprehensive tests for all new tool paths and limits

## Tests
- `make verify`

## Notes
- Read-only by design (`readOnlyHint=True`)
- Output is kept raw to avoid early normalization
- Works on RH-like systems with DNF >= 4

### DIFF
--- a/src/linux_mcp_server/utils/validation.py
+++ b/src/linux_mcp_server/utils/validation.py
@@ -1,3 +1,5 @@
+import re
+
 from pathlib import Path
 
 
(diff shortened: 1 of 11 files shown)


src/linux_mcp_server/utils/validation.py
```diff
@@ -56,6 +58,131 @@ def validate_path(path: str) -> Path:
     return Path(path)
 
 
+def validate_dnf_package_name(value: str) -> str:
```
r0x0d, line 61: I wonder if we should let the LLM figure out this instead of us

subpop:
I have concerns about the scalability of distro-specific tools like this. I like that this MCP server is very distro-agnostic; that's one of its strengths. Adding dnf-specific tools means we're preferring Fedora-based distributions over others. We should add support for `apt` (and probably `zypper` and `pacman`) to cover the vast majority of package managers on Linux installs that this server might encounter.

PR #266, opened 2026-02-09
Adapt to host theme and styling

owtaylor:
Propagate theme/colors/etc from the host to the mcp-app. [SDK docs](https://github.com/modelcontextprotocol/ext-apps/blob/main/docs/patterns.md#adapting-to-host-context-theme-styling-fonts-and-safe-areas) Goose doesn't set much of this yet, but it does set the theme, so this avoids a black background for our mcp-app when Goose has a white background.


Jazzcort (approved):
Tested it with switching theme in Goose Desktop and it works! Haven't found a way to test font ...etc, but we can always come back and fix stuffs if we find the styles are not consistent later. LGTM 😁

PR #269, opened 2026-02-10
[RSPEED-2404] persistence for mcp-apps

Jazzcort:
Adds an ExecutionPersistenceStore and hidden tools (get/set_execution_state) to the server to track script status. Updates the run-script-app frontend to sync state changes and modifies run_script_modify_interactive to include the request ID in the result.

### DIFF
--- a/mcp-app/src/run-script-app.tsx
+++ b/mcp-app/src/run-script-app.tsx
@@ -6,30 +6,24 @@ import {
   useApp,
 } from "@modelcontextprotocol/ext-apps/react";
 import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
-import { StrictMode, useCallback, useEffect, useState } from "react";
+import { StrictMode, useEffect, useMemo, useState } from "react";
 import { createRoot } from "react-dom/client";
+import {
+  ExecuteScriptResultSchema,
+  GetExecutionStateResultSchema,
+  McpAppToolParamsSchema,
+  McpAppToolResultSchema,
+  type ExecutionState,
+} from "./types";
+import { extractText, formatOutput, formatOutputForToolError } from "./utils";
 
 const IMPLEMENTATION = { name: "Run Script App", version: "1.0.0" };
-
-type GateKeeperStatus =
-  | "OK"
-  | "BAD_DESCRIPTION"
-  | "POLICY"
-  | "MODIFIES_SYSTEM"
-  | "UNCLEAR"
-  | "DANGEROUS"
-  | "MALICIOUS";
 
 const log = {
   info: console.log.bind(console, "[APP]"),
   warn: console.warn.bind(console, "[APP]"),
   error: console.error.bind(console, "[APP]"),
 };
-
-function extractText(callToolResult: CallToolResult): string {
-  const { text } = callToolResult.content?.find((c) => c.type === "text")!;
-  return text;
-}
 
 function RunScriptApp() {
   const [toolResult, setToolResult] = useState<CallToolResult | undefined>(
@@ -109,16 +103,11 @@ interface RunScriptAppInnerProps {
   toolResult: CallToolResult | undefined;
 }
 
-type ExecutionState =
-  | "initialized"
-  | "success"
-  | "failed"
-  | "rejected"
-  | "executing";
-
+// TODO: This function would be deleted in the following UI rework PR
 const pickStateColor = (state: ExecutionState): string => {
   switch (state) {
-    case "initialized":
+    case "initial":
+    case "waiting-approval":
       return "text-white";
     case "executing":
       return "text-yellow-500";
@@ -135,62 +124,133 @@ function RunScriptAppInner({
   toolResult,
 }: RunScriptAppInnerProps) {
   const [executionState, setExecutionState] =
-    useState<ExecutionState>("initialized");
+    useState<ExecutionState>("initial");
   const [executionResult, setExecutionResult] = useState<string>("");
 
-  const handleAccept = useCallback(async () => {
-    setExecutionState("executing");
+  const validatedToolRequestParams = useMemo(() => {
+    if (!toolRequestParams) {
+      return undefined;
+    }
 
-    const params = toolRequestParams || {};
-    const script = (params["script"] || "") as string;
-    const scriptType = (params["script_type"] || "bash") as string;
-    const host = params["host"] || null;
+    const parsedResult = McpAppToolParamsSchema.safeParse(toolRequestParams);
+
+    if (parsedResult.success) {
+      return parsedResult.data;
+    } else {
+      log.error(parsedResult.error);
+      return undefined;
+    }
+  }, [toolRequestParams]);
+
+  const validatedToolResult = useMemo(() => {
+    if (!toolResult) {
+      return undefined;
+    }
+
+    const parsedResult = McpAppToolResultSchema.safeParse(
+      toolResult?.structuredContent,
+    );
+
+    if (parsedResult.success) {
+      return parsedResult.data;
+    } else {
+      log.error(parsedResult.error);
+      return undefined;
+    }
+  }, [toolResult]);
+
+  useEffect(() => {
+    if (!validatedToolResult) return;
+
+    const getExecutionStateFromServer = async () => {
+      const result = await app.callServerTool({
+        name: "get_execution_state",
+        arguments: {
+          id: validatedToolResult.id,
+        },
+      });
+
+      if (result.isError) return;
+
+      const validatedResult = GetExecutionStateResultSchema.safeParse(
+        result.structuredContent,
+      );
+
+      if (validatedResult.success) {
+        setExecutionState(validatedResult.data.state);
+      }
+    };
+
+    getExecutionStateFromServer();
+  }, [app, validatedToolResult]);
+
+  const handleAccept = async () => {
+    if (!validatedToolRequestParams || !validatedToolResult) return;
+
+    setExecutionState("executing");
 
     let updatedExecutionState: ExecutionState = executionState;
     let updatedExecutionResult = executionResult;
+    let outputToModel = "";
 
     try {
       const result = await app.callServerTool({
         name: "execute_script",
-        arguments: { script: script, script_type: scriptType, host: host },
+        arguments: { id: validatedToolResult.id },
       });
 
-      if (!!result.isError) {
-        updatedExecutionState = "failed";
-      } else {
-        updatedExecutionState = "success";
+      if (result.isError) {
+        throw new Error(extractText(result));
       }
 
-      updatedExecutionResult = extractText(result);
+      const executeScriptResult = ExecuteScriptResultSchema.parse(
+        result.structuredContent,
+      );
+
+      updatedExecutionResult = executeScriptResult.output;
+      updatedExecutionState = executeScriptResult.state;
+
+      outputToModel = formatOutput(
+        validatedToolRequestParams,
+        validatedToolResult,
+        executeScriptResult,
+      );
     } catch (e) {
-      updatedExecutionState = "failed";
-      updatedExecutionResult = e as string;
-      console.error("callServerTool failed", e);
+      outputToModel = formatOutputForToolError(
+        validatedToolRequestParams,
+        validatedToolResult,
+        JSON.stringify(e),
+      );
+
+      updatedExecutionState = "failure";
+      updatedExecutionResult = JSON.stringify(e);
     }
 
     setExecutionState(updatedExecutionState);
     setExecutionResult(updatedExecutionResult);
 
-    try {
-      await app.sendMessage({
-        role: "user",
-        content: [
-          {
-            type: "text",
-            text: `The script ${script} has been executed. Here is the detail of the execution.\n--------------------------------------------\nExecution status: ${updatedExecutionState}\nExecution result: ${updatedExecutionResult}`,
-          },
-        ],
-      });
-    } catch (e) {
-      console.error(`sendMessage failed: ${e}`);
-    }
-  }, [app, toolRequestParams]);
-
-  const handleReject = () => {
-    setExecutionState("rejected");
+    app.sendMessage({
+      role: "user",
+      content: [
+        {
+          type: "text",
+          text: outputToModel,
+        },
+      ],
+    });
   };
 
-  if (!toolResult || !toolRequestParams) {
+  const handleReject = async () => {
+    if (!validatedToolResult) return;
+
+    setExecutionState("rejected-user");
+    app.callServerTool({
+      name: "reject_script",
+      arguments: { id: validatedToolResult.id },
+    });
+  };
+
+  if (!validatedToolResult || !validatedToolRequestParams) {
     return (
       <div className="flex justify-center items-center min-h-[200px]">
         <div className="flex items-center">
@@ -204,7 +264,7 @@ function RunScriptAppInner({
     );
   }
 
-  switch (toolResult.structuredContent!["status"] as GateKeeperStatus) {
+  switch (validatedToolResult.status) {
     case "OK":
       return (
         <div className="p-2">
@@ -217,15 +277,15 @@ function RunScriptAppInner({
                 </span>
               </p>
               <p>
-                <strong>Script:</strong> {toolRequestParams["script"] as string}
+                <strong>Script:</strong> {validatedToolRequestParams.script}
               </p>
               <p>
                 <strong>Script Type:</strong>{" "}
-                {toolRequestParams["script_type"] as string}
+                {validatedToolRequestParams.scriptType}
               </p>
               <p>
                 <strong>Description:</strong>{" "}
-                {toolRequestParams["description"] as string}
+                {validatedToolRequestParams.description}
               </p>
               {executionResult !== null && (
                 <p>
@@ -236,14 +296,14 @@ function RunScriptAppInner({
             <div className="flex-none">
               <button
                 className="btn-primary"
-                disabled={executionState !== "initialized"}
+                disabled={executionState !== "waiting-approval"}
                 onClick={handleAccept}
               >
                 Allow
               </button>
               <button
                 className="btn-primary"
-                disabled={executionState !== "initialized"}
+                disabled={executionState !== "waiting-approval"}
                 onClick={handleReject}
               >
                 Deny
@@ -257,13 +317,10 @@ function RunScriptAppInner({
         <div className="p-2 whitespace-pre-wrap break-all">
           <p>
             <strong>Gatekeeper State:</strong>{" "}
-            <span className="text-red-500">
-              {toolResult.structuredContent!["status"] as string}
-            </span>
+            <span className="text-red-500">{validatedToolResult.status}</span>
           </p>
           <p>
-            <strong>Detail:</strong>{" "}
-            <span>{toolResult.structuredContent!["detail"] as string}</span>
+            <strong>Detail:</strong> <span>{validatedToolResult.detail}</span>
           </p>
         </div>
       );
--- a/mcp-app/src/utils.ts
+++ b/mcp-app/src/utils.ts
@@ -1,0 +1,52 @@
+import type {
+  McpAppToolParams,
+  ExecuteScriptResult,
+  McpAppToolResult,
+} from "./types";
+
+import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+
+export const formatOutput = (
+  toolRequestParams: McpAppToolParams,
+  toolResult: McpAppToolResult,
+  executeScriptResult: ExecuteScriptResult,
+) => {
+  let result = "";
+
+  if (toolRequestParams.script.trim().split("\n").length > 1) {
+    result += `Script id=${toolResult.id} (${toolRequestParams.description}) executed`;
+  } else {
+    result += `Script \`${toolRequestParams.script}\` executed`;
+  }
+
+  if (executeScriptResult.state === "success") {
+    result += " successfully, ";
+  } else {
+    result += ", ";
+  }
+
+  result += `and returned ${executeScriptResult.state}. Output:\n${executeScriptResult.output}`;
+  return result;
+};
+
+export const formatOutputForToolError = (
+  toolRequestParams: McpAppToolParams,
+  toolResult: McpAppToolResult,
+  message: string,
+) => {
+  return toolRequestParams.script.trim().split("\n").length > 1
+    ? `Script id=${toolResult.id} (${toolRequestParams.description}) failed to execute: ${message}`
+    : `Script \`${toolRequestParams.script}\` failed to execute: ${message}`;
+};
+
+/**
+ * Extracts text content from a CallToolResult.
+ *
+ * @param callToolResult - The CallToolResult containing content to extract text from
+ * @returns The extracted text string
+ * @throws Runtime error if no text content is found in the content array
+ */
+export function extractText(callToolResult: CallToolResult): string {
+  const { text } = callToolResult.content?.find((c) => c.type === "text")!;
+  return text;
+}
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -202,7 +202,10 @@ from linux_mcp_server.tools import *  # noqa: E402, F403
 # providing `mcp-app` compatibility during the initialize request.
 if CONFIG.use_mcp_apps:
     mcp.add_tool(run_script_modify_interactive)
+    mcp.add_tool(get_execution_state)
     mcp.add_tool(execute_script)
+    mcp.add_tool(reject_script)
+
 else:
     mcp.add_tool(run_script_modify)
 
--- a/src/linux_mcp_server/tools/run_script.py
+++ b/src/linux_mcp_server/tools/run_script.py
@@ -1,6 +1,10 @@
 import logging
+import secrets
 import shlex
 import typing as t
+
+from dataclasses import asdict
+from dataclasses import dataclass
 
 from fastmcp import Context
 from fastmcp.exceptions import ToolError
@@ -11,6 +15,7 @@ from mcp.types import TextContent
 from mcp.types import ToolAnnotations
 
 # from pydantic import BaseModel
+from pydantic import BaseModel
 from pydantic import Field
 
 from linux_mcp_server.audit import log_tool_call
@@ -25,6 +30,9 @@ from linux_mcp_server.utils.types import Host
 
 logger = logging.getLogger("linux-mcp-server")
 
+ExecutionState = t.Literal[
+    "waiting-approval", "success", "failure", "executing", "rejected-user", "rejected-gatekeeper"
+]
 
 # This would make more sense as a StrEnum, but that generates a schema with
 # the enum in $defs, which deeply confuses (at least) the combination of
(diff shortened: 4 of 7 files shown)


mcp-app/src/run-script-app.tsx
```diff
@@ -129,6 +133,23 @@ const pickStateColor = (state: ExecutionState): string => {
...
+  options?: RequestOptions,
+): Promise<CallToolResult> => {
+  try {
+    const result = await app.callServerTool(params, options);
+    return result;
+  } catch (e) {
+    log.error(`${params.name} failed`, e);
+    return {
+      content: [{ type: "text", text: `${JSON.stringify(e)}` }],
+      isError: true,
```
owtaylor: I don't like this much - the SDK authors made some determination about what throwing an error means (transport failures) versus returning a result with isError true ( tool execution failures), and by changing that here, we've introduced specialized knowledge about how *we* use the SDK.

mcp-app/src/run-script-app.tsx
```diff
@@ -138,39 +159,64 @@ function RunScriptAppInner({
     useState<ExecutionState>("initialized");
   const [executionResult, setExecutionResult] = useState<string>("");
 
+  useEffect(() => {
+    const getExecutionHistory = async (id: number) => {
```
owtaylor: Is it "execution history" or "execution state" - should be consistent.

Also suggest `getExecutionStateFromServer` or something like that - `getExecutionState` seems too paired with `setExecutionState`

mcp-app/src/run-script-app.tsx
```diff
@@ -138,39 +159,64 @@ function RunScriptAppInner({
...
-        updatedExecutionState = "success";
-      }
+    const result = await safeguardCallServerTool(app, {
+      name: "execute_script",
+      arguments: { script: script, script_type: scriptType, host: host },
+    });
 
-      updatedExecutionResult = extractText(result);
-    } catch (e) {
+    if (!!result.isError) {
```
owtaylor: Why the `!!` ?

mcp-app/src/run-script-app.tsx
```diff
@@ -138,39 +159,64 @@ function RunScriptAppInner({
...
+      updatedExecutionState = "success";
     }
 
+    updatedExecutionResult = extractText(result);
+
     setExecutionState(updatedExecutionState);
     setExecutionResult(updatedExecutionResult);
 
+    safeguardCallServerTool(app, {
+      name: "set_execution_state",
```
owtaylor: I don't think the app should be driving the logic here - it shouldn't be responsible for changing the state here. But we do need to change the state if Deny is clicked.

What I'd lean to is:

```
 run_script_modify_interactive(description, script, script_type, ...) => {
    structured_content: { id: "asd311a" }, content: "... instructions to LLM ... "
 set_script_approval(id, action: "approve" | "deny") => {}
 get_script_state(id) => {
    status: "rejected" | "waiting_approval" | | "denied" | "executing" | "success" | "failure"
 }
```

Note:
 * id should be a cryptographically secure random value (`secrets.token_url_safe(16)`), not an integer. This doesn't matter for stdio, but does matter for a shared http server. Future proofing.
 * I've shown `set_script_approval` not taking the script/script_type parameters. I think this is cleaner, but will require storing the script/script_type under the id, not just the status. If it ends up being hard for any reason, please feel free to add script/script_type back.
 * In the case where it's rejected, we'll never get to calling get_script_state(). We also don't care about the difference between "executing" | "success" | "failure" but I think it's simpler to have the enum have all possible states rather than whittling it down.
 * I made get_script_state() return an object with the status field so we could compatibly extend it to have the script output or other things in it.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -243,7 +245,11 @@ async def _run_script_modify_interactive(
             )
         )
 
-    result = ToolResult(content=content, structured_content=gatekeeper_result.model_dump())
+    gatekeeper_result_dict = gatekeeper_result.model_dump()
+    # Attach request ID to the result to keep the further execution persistent
+    gatekeeper_result_dict["id"] = int(ctx.request_id)
+
+    result = ToolResult(content=content, structured_content=gatekeeper_result_dict)
```
owtaylor: I don't like how the result has the type of "GatekeeperResult + id" - I'd actually have a type that defines our public interface.

```
class RunScriptInteractiveResult(BaseModel):
     id: str
     status: GatekeeperStatus
     detail: str
```

And build that out of the gatekeeper result. Then we can create the tool with `output_schema=GatekeeperResult`. (We can't count on FastMCP automatically setting the schema from the function signature since we are returning ToolResult)

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,3 +1,21 @@
...
+
+
 RUN_SCRIPT_APP_URI = "ui://run_script_readonly_with_mcp_app/run-script-app.html"
 ALLOWED_UI_RESOURCE_URIS = set([RUN_SCRIPT_APP_URI])
 MCP_APP_MIME_TYPE = "text/html;profile=mcp-app"
+
+ExecutionState = Literal["initialized", "success", "failed", "rejected"]
+
+
+class ExecutionPersistenceStore(BaseModel):
```
owtaylor: This doesn't seem like it needs to be a pydantic type.

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,3 +1,21 @@
...
+    def __init__(self):
+        self._execution_persistence_store: dict[int, ExecutionState] = {}
```
owtaylor: maybe this can be something shorter like `_values` ?

src/linux_mcp_server/server.py
```diff
@@ -195,14 +198,42 @@ async def _read_resource_with_meta(req: ReadResourceRequest):
...
+
 # TODO: Dynamically inject the 'modify' tool based on user compatibility.
 #
 # This is a temporary implementation. The injection logic should be moved to the
 # `on_initialize` handler in `DynamicDiscoveryMiddleware` once Goose starts
 # providing `mcp-app` compatibility during the initialize request.
 if CONFIG.use_mcp_apps:
+
+    def _get_execution_state(id: int):
+        return get_execution_state(id, execution_persistence_store)
```
owtaylor: Let's keep this stuff in run_script.py ... it's very closely linked to what's already in there.

mcp-app/src/run-script-app.tsx
Jazzcort: I hesitated on this change a bit, but I ended up having it because I want to get rid of having try-catch blocks everywhere. However, I agree that we shouldn't introduce specialized knowledge and potentially confuse the model.
Will revert it. 😁

mcp-app/src/run-script-app.tsx
Jazzcort: This one got left out from my renaming work. Thanks for the catch and suggestion! 😁  I'll use `getExecutionStateFromServer` which explains itself more explicitly.

mcp-app/src/run-script-app.tsx
Jazzcort: My OCD kicks in and makes me want to have pure boolean in the if clause. 😆 Will change it.

mcp-app/src/run-script-app.tsx
Jazzcort: I first tried implementing this on the server side but encountered a problem that I can't catch the "failed" (execution failed) case in the middleware. 
```python
response = await call_next(content)
```
`call_next` here will just raise an error instead of returning a ToolResult with `isError: true`. And I also found that the "rejected" state can only be set from the app side so I end up shifting the setting state work to the app side. 

Anyway, now I have a new idea. I want to move the store instance into the run_script module and we just directly access the store instance within `_execute_script` or `set_script_approval`. What do you think?

By introducing `set_script_approval`, I think we're removing `execute_script` from the tool and just call it with in `set_script_approval`. Am I correct? I'll try to implement the version without script/script_type first and see how it goes. 😁 

>status: "rejected" | "waiting_approval" | | "denied" | "executing" | "success" | "failure"

One more thing I'd like to be clarified. The "rejected" here means rejected by the gatekeeper right? 🤔

src/linux_mcp_server/tools/run_script.py
Jazzcort: Totally agree! Will change!

src/linux_mcp_server/mcp_app.py
Jazzcort: You're right! I somehow revealed this parameter to FastMcp and I got complains from pydantic, so I made this a BaseModel to get rid of the error but forgot to remove it after I hid the parameter. Will remove it!

src/linux_mcp_server/mcp_app.py
Jazzcort: sounds good!

src/linux_mcp_server/server.py
Jazzcort: Will move it to run_script with the store like what I mention in the previous comment. 😁

mcp-app/src/run-script-app.tsx
owtaylor: > Anyway, now I have a new idea. I want to move the store instance into the run_script module and we just directly access the store instance within `_execute_script` or `set_script_approval`. What do you think?

Yes, that's what I'd expect - from my perspective anything we have in server.py should be there because we can't do it within basic framework - but as much as possible we should keep actual tool logic within run_script.py. I don't think there's any reason to handle this at the middleware layer - the state is canonically is controlled by the tools in run_script.py, and if they don't run, nothing needs to change for the state.

> By introducing set_script_approval, I think we're removing execute_script from the tool and just call it with in set_script_approval. Am I correct? I'll try to implement the version without script/script_type first and see how it goes. 😁

Yes, in this model, the app just says "ok, go ahead" or "the user said no", and the MCP server is responsible for knowing what goes next.

> One more thing I'd like to be clarified. The "rejected" here means rejected by the gatekeeper right? 🤔

Yeah, "rejected" is rejected by gatekeeper, "denied" is denied by user. Maybe it should be "rejected-gatekeeper", "rejected-user" instead - probably would help people reading the code in the future.

mcp-app/src/run-script-app.tsx
```diff
@@ -135,37 +138,70 @@ function RunScriptAppInner({
...
+        }
+      } catch (e) {
+        log.error("get_execution_state failed", e);
+      }
+    };
+
+    if (toolResult?.structuredContent) {
+      getExecutionStateFromServer(toolResult.structuredContent["id"] as string);
+    }
+  }, [app, toolResult]);
```
owtaylor: This is somehow quite hairy to read - I think the main problem is that you have to read down to the bottom of it to figure out what this function is for. I threw it at Claude and it came up with:

```tsx
useEffect(() => {
  const executionId = toolResult?.structuredContent?.id;
  if (!executionId) return;

  const fetchExecutionState = async () => {
    try {
      const result = await app.callServerTool({
        name: "get_execution_state",
        arguments: { id: executionId },
      });

      if (!result.isError && result.structuredContent) {
        setExecutionState(result.structuredContent.state as ExecutionState);
      }
    } catch (e) {
      log.error("get_execution_state failed", e);
    }
  };

  fetchExecutionState();
}, [app, toolResult?.structuredContent?.id]);
```

Which is almost the same thing, but pulling the id in a variable up to the top and using it from the inner scrope rather than passing it in as as a parameter makes it more readable for me.  [variable name would be better as scriptId or something, I didn't give much context when asking for a restructure.]

mcp-app/src/run-script-app.tsx
```diff
@@ -177,18 +213,31 @@ function RunScriptAppInner({
...
-      console.error(`sendMessage failed: ${e}`);
+      log.error(`sendMessage failed: ${e}`);
     }
-  }, [app, toolRequestParams]);
+  }, [app, toolRequestParams, toolResult]);
 
-  const handleReject = () => {
-    setExecutionState("rejected");
-  };
+  const handleReject = useCallback(() => {
```
owtaylor: it's weird to me that handleAccept uses an async function, and this doesn't. This makes reading the two paths hard to read.

mcp-app/src/run-script-app.tsx
```diff
@@ -135,37 +138,70 @@ function RunScriptAppInner({
...
+    if (toolResult?.structuredContent) {
+      getExecutionStateFromServer(toolResult.structuredContent["id"] as string);
+    }
+  }, [app, toolResult]);
+
   const handleAccept = useCallback(async () => {
+    // Do nothing if either toolRequestParams or toolResult are missing
+    if (!toolRequestParams || !toolResult?.structuredContent) {
+      return;
+    }
```
owtaylor: I don't think we actually need tooLRequestParams here - again extracting what we need into an early return will improve readability.

```
 const scriptId = toolResult?.structuredContent?.id;
 if (!scriptId) {
      return;
 }
```

mcp-app/src/run-script-app.tsx
```diff
@@ -135,37 +138,70 @@ function RunScriptAppInner({
...
+        log.error("get_execution_state failed", e);
+      }
+    };
+
+    if (toolResult?.structuredContent) {
+      getExecutionStateFromServer(toolResult.structuredContent["id"] as string);
+    }
+  }, [app, toolResult]);
+
   const handleAccept = useCallback(async () => {
```
owtaylor: Probably better not to use useCallback here .... there's nothing heavy that depends on this callback.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -263,6 +260,7 @@ async def _run_script_modify_interactive(
     title="Propose to run a script that modifies system",
     description=RUN_SCRIPT_MODIFY_INTERACTIVE,
     annotations=ToolAnnotations(destructiveHint=True),
+    output_schema=RunScriptInteractiveResult.model_json_schema(),
```
owtaylor, line 397: Is the .model_json_schema() needed here? I thought that FastMCP converted automatically.

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,3 +1,47 @@
...
+
+@dataclass()
+class ExecutionDetail:
+    state: ExecutionState
+    script: str
+    script_type: ScriptType
+    host: Host
+
+
+class ExecutionPersistenceStore:
```
owtaylor: If we keep this in run_script.py we don't need to mve ScriptType around - this is specifically about the run_script functionality and not really about "mcp-app machinery".

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,3 +1,47 @@
...
+    This store maps request IDs to their execution details (script, host, state) so we can
+    retrieve and execute the script after user approval, even though the original request
+    context is no longer available. It also preserve execution state (waiting-approval, executing,
+    success, failure, rejected) for the MCP app UI.
+    """
+
+    def __init__(self):
+        self._details: dict[str, ExecutionDetail] = {}
+
+    def initialize_detail(self, id: str, script: str, script_type: ScriptType, host: Host):
```
owtaylor: I'd suggest making the Store allocate the IDs, you could do something like:

```
def add_script(*, script: str, script_type: ScriptType, host: Host) -> str
```

Note the * to force the caller to use keyword parameters, and avoid the possibility of getting the order wrong.

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,3 +1,47 @@
...
+        self._details[id] = ExecutionDetail(state="waiting-approval", script=script, script_type=script_type, host=host)
+
+    def get_execution_detail(self, id: str) -> ExecutionDetail | None:
+        return self._details.get(id)
```
owtaylor: If we let this throw a KeyError from `_details[id]` then we can get consistent behavior between this and `set_execution_state`

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,3 +1,47 @@
...
+
+    def set_execution_state(self, id: str, new_state: ExecutionState):
+        if id not in self._details:
```
owtaylor: Then we can remove this check and let the KeyError throw - silently doing nothing on a bad id isn't something we want to do.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -330,3 +328,53 @@ async def _run_script_modify(
...
+async def _set_script_approval(id: str, action: ScriptApprovalAction):
+    match action:
+        case "approve":
+            execution_detail = execution_persistence_store.get_execution_detail(id)
+            if execution_detail is None:
+                raise RuntimeError("Tried to execute script before it's initialized")
+
+            result = await _execute_script(execution_detail.script_type, execution_detail.script, execution_detail.host)
+
+            if result.startswith("Error executing script:"):
```
owtaylor: Why don't we move updating the execution state into _execute_script() so we don't have to do this.

mcp-app/src/run-script-app.tsx
Jazzcort: Didn't know the dependency can be used like this way `toolResult?.structuredContent?.id`. It does make it cleaner! Will change!

src/linux_mcp_server/tools/run_script.py
Jazzcort: So I did get a syntax error with `output_schema=RunScriptInteractiveResult`. It complained `output_schema` needs to be `dict[str, Any]` or some sort.

src/linux_mcp_server/tools/run_script.py
Jazzcort: Good idea! 😁

src/linux_mcp_server/mcp_app.py
Jazzcort: Agree! Will change it

mcp-app/src/run-script-app.tsx
Jazzcort: @owtaylor But don't we need toolRequestParams to construct the ugly content for sendMessage? 😆

mcp-app/src/run-script-app.tsx
```diff
@@ -135,40 +138,56 @@ function RunScriptAppInner({
   toolResult,
 }: RunScriptAppInnerProps) {
   const [executionState, setExecutionState] =
-    useState<ExecutionState>("initialized");
+    useState<ExecutionState>("initial");
   const [executionResult, setExecutionResult] = useState<string>("");
+  const getExecutionStateFromServer = async (scriptId: string) => {
```
Jazzcort: I tweak it a bit and move it out of the useEffect so if can be triggered in the handleAccept to get the updated state. What do you think?

mcp-app/src/run-script-app.tsx
```diff
@@ -177,17 +196,28 @@ function RunScriptAppInner({
         content: [
           {
             type: "text",
-            text: `The script ${script} has been executed. Here is the detail of the execution.\n--------------------------------------------\nExecution status: ${updatedExecutionState}\nExecution result: ${updatedExecutionResult}`,
+            text: `Here is the detail of the execution.\n--------------------------------------------\nExecution status: ${updatedExecutionState}\nExecution result: ${updatedExecutionResult}`,
```
Jazzcort: Not quite sure about what we should post to the model here. I feel like we should at least provide the execution state and result. In case if the result is empty model can know the execution state.

Jazzcort (commented):
Thanks for the review! Let me know if any changes are needed!

mcp-app/src/run-script-app.tsx
```diff
@@ -135,40 +138,56 @@ function RunScriptAppInner({
...
 
-  const handleAccept = useCallback(async () => {
-    setExecutionState("executing");
+      if (!result.isError && result.structuredContent) {
+        setExecutionState(result.structuredContent.state as ExecutionState);
+        return result.structuredContent.state;
+      }
+    } catch (e) {
+      log.error("get_execution_state failed", e);
+    }
```
owtaylor: it's better to not catch at all here - it will get logged anyways - that will avoid the function having a null result in case of an error and a caller needing to handle it.

mcp-app/src/run-script-app.tsx
owtaylor: Hmm, I think moving it out here is a bit of a react anti-pattern. The issue here is that the getExecutionStateFromServer function implicitly depends on app, so it's the needed as a dependency of the useEffect (which you have correctly), but you can't tell by looking just at the contents of useEffectthat it is needed. So it will work as you have it, but very easy to get wrong This is a case where you would need useCallback.
 getExecutionStateFromServer = useCallback(...., [app]);
 useEffect(......, [getExecutionStateFromServer, ....])

But I think my preference would to be make execute_script return an object with the state as a property and avoid having to make two tool calls, and then we'll avoid complexity and you can move this back into useEffect.

There's a nasty edge case that is somewhat realistic - what if the app is reinitialized *while* the script is executing (by reloading the web page, or whatever). We want to "rejoin" the execution and catch the update and message the model. This is certainly possible to handle, but let's ignore it for now.

mcp-app/src/run-script-app.tsx
```diff
@@ -177,17 +196,28 @@ function RunScriptAppInner({
...
+
+    setExecutionState("rejected-user");
+    app
+      .callServerTool({
+        name: "reject_script",
+        arguments: { id: scriptId },
+      })
+      .catch((err) => {
+        log.error("reject_script failed", err);
+      });
```
owtaylor: Again, there's no point in catch simply to log the exception - the additional context of "reject_script failed" here isn't worth the extra code. 

There's some argument that we shouldn't consider this final until it is acknowledge by the server - and do:

```ts
 callServerTool({ name: "reject_script", .... }).then({state} => setExecutionState(state))
```

But then we need additional code to debounce the interaction (avoid the user hitting a button multiple times). Let's keep it the way you have it.

mcp-app/src/run-script-app.tsx
owtaylor: Hmm, this is making me think! - I'd like to say that we return a json structured result to the model, but as long as we're using the message setup, then we need to keep things human readable. We also have to deal with the fact that we don't control the order of execution, so we have to be clear about *which* tool call (if the model made multiple tool calls, or the user ignored the app and proceeded with the conversation.)

Proposal:

```
case 1: execute_script return error (failed to look up hostname, say)

Script ${script} failed to execute: {message} 

case 2: execute_script executed command, but it failed (exit status=1)

Script ${script} executed, and returned {exit_status}. Output:\n${output}

case 2: execute_script executed command, but it failed (exit status 2)

Script ${script} executed successfully, and returned {exit_status}. Output:\n${output}
```

*However* this only makes sense when ${script} is a single line. It could be quite extension, so I'd suggest that in thta case we return this differently:

```
Script ${script_id} ($script_description) executed succesfully, and returned {exit_status}. Output:\n${output}
```

We include ${script_id} here to give the model at least the chance to figure things out if it's iterating the script to fix it, and ${script_description} to help the human.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -183,23 +279,18 @@ async def run_script_readonly(
...
-    script: t.Annotated[
-        str,
-        Field(description="The script to run."),
-    ],
-    host: Host = None,
+    id: t.Annotated[str, Field(description="The associated ID of the script to be executed")],
 ) -> str:
-    command = _wrap_script(script_type, script)
+    script_details = script_store.get_script_details(id)
+    command = _wrap_script(script_details.script_type, script_details.script)
```
owtaylor, line 302: I would make set the state here to "executing" since this operation could be long running.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -183,23 +279,18 @@ async def run_script_readonly(
...
 
-    returncode, stdout, stderr = await execute_command(command, host=host)
+    returncode, stdout, stderr = await execute_command(command, host=script_details.host)
```
owtaylor: This needs a `try .... catch: ....` if execute_command raises on a DNS lookup failure or something.

mcp-app/src/run-script-app.tsx
Jazzcort: > But I think my preference would to be make execute_script return an object with the state as a property and avoid having to make two tool calls, and then we'll avoid complexity and you can move this back into useEffect.

I agree! Having `execute_script` return an object that contains the execution state and output is more robust than making two tool calls. I will do that and move `getExecutionStateFromServer` back to useEffect.

 > There's a nasty edge case that is somewhat realistic - what if the app is reinitialized while the script is executing (by reloading the web page, or whatever). We want to "rejoin" the execution and catch the update and message the model. This is certainly possible to handle, but let's ignore it for now.

I did think about this nasty edge case before when I was implementing the draft of this persistence feature, but back then I thought we can not address this edge case because we only preserve the execution state. However, since we have the `ScriptStore` now, it's possible to resume the execution!

mcp-app/src/run-script-app.tsx
Jazzcort: Got it! Will change

mcp-app/src/run-script-app.tsx
Jazzcort: >Again, there's no point in catch simply to log the exception - the additional context of "reject_script failed" here isn't worth the extra code.

Agree! will change it!

> There's some argument that we shouldn't consider this final until it is acknowledge by the server - and do:
>```typescript
>callServerTool({ name: "reject_script", .... }).then({state} => >setExecutionState(state))
>```
>But then we need additional code to debounce the interaction (avoid the user hitting a button multiple times). Let's keep it the way you have it.

I agree that it's more valid to get acknowledged by the server, but I also feel like it's not worth the code to make `reject_script` debounce since it's just a simple state update for the server. As you said, we can keep it this way and if we encounter some problems in the future we can adjust it! 😁 

Another way of solve this that I can think of is to have a intermediate state before "rejected-user" - maybe "rejecting" to prevent the user from hitting the button multiply times from the UI perspective. However, there's also some argument that if having the state acknowledged by the server is worth adding a state to address it.

mcp-app/src/run-script-app.tsx
Jazzcort: I like this idea! Will change! 😁

src/linux_mcp_server/tools/run_script.py
Jazzcort: You're right! we should definitely catch that. 😁

src/linux_mcp_server/tools/run_script.py
```diff
@@ -182,25 +279,39 @@ async def run_script_readonly(
         return f"Error executing script: return code {returncode}, stderr: {stderr}"
 
 
+@dataclass
+class ExecuteScriptResult:
+    state: t.Literal["success", "failure"]
+    output: str
```
Jazzcort, line 286: I think we can only return the updated state and output here. The script, script_type and host can rely on the the initial tool input in the app. What do you think? @owtaylor

Jazzcort (commented):
Sorry for introducing these type validating stuffs at the very last moment. 😅 I was planing to have these in the UI rework PR but since I need to implement the function to format the output to post to the model, I just feel if I implemented without the validated parameters, I have to almost rewrite it completely in the next PR which I think is not quite efficient. 😁 Therefore, I try to squeeze these in this PR.

I think adding data validation improves the readability and maintainability of our codebase. Instead of checking each nested field individually wherever we need it, we define a schema that validates the entire object as a whole. This keeps related fields tied together, makes the intent clearer, and reduces scattered validation logic throughout the code.

What do you think? @owtaylor

mcp-app/src/run-script-app.tsx
```diff
@@ -135,62 +123,105 @@ function RunScriptAppInner({
...
   const [executionState, setExecutionState] =
-    useState<ExecutionState>("initialized");
+    useState<ExecutionState>("initial");
   const [executionResult, setExecutionResult] = useState<string>("");
 
-  const handleAccept = useCallback(async () => {
-    setExecutionState("executing");
+  const validatedToolRequestParams = useMemo(() => {
+    const parsedResult = McpAppToolParamsSchema.safeParse(toolRequestParams);
+    return parsedResult.success ? parsedResult.data : undefined;
```
owtaylor: This probably would be better off logging `parseResult.error`, since we otherwise mysteriously swallow the error state but then we'd want to avoid calling `safeParse(undefined)` since that's not an error state.

mcp-app/src/run-script-app.tsx
```diff
@@ -135,62 +123,105 @@ function RunScriptAppInner({
...
+      const result = await app.callServerTool({
+        name: "get_execution_state",
+        arguments: {
+          id: validatedToolResult.id,
+        },
+      });
+
+      if (!result.isError && result.structuredContent) {
+        setExecutionState(result.structuredContent.state as ExecutionState);
+        return result.structuredContent.state;
```
owtaylor: You don't use this return.

mcp-app/src/run-script-app.tsx
```diff
@@ -135,62 +123,105 @@ function RunScriptAppInner({
...
+        arguments: { id: validatedToolResult.id },
       });
 
-      if (!!result.isError) {
-        updatedExecutionState = "failed";
-      } else {
-        updatedExecutionState = "success";
-      }
+      const executeScriptResult = ExecuteScriptResultSchema.parse(
+        result.structuredContent,
```
owtaylor, line 207: If the tool fails inside the MCP server, don't we have `result.isError == true`, `result.structuredContent == undefined`, `result.content == <error message>` ?

mcp-app/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+export const formatOutput = (
+  toolRequestParams: McpAppToolParams,
+  toolResult: McpAppToolResult,
+  executeScriptResult: ExecuteScriptResult,
+) => {
+  let result = "";
+
+  if (toolRequestParams.script.trim().split("\n").length > 1) {
+    result += `Script ${toolResult.id} ${toolRequestParams.description} executed`;
```
owtaylor: `Script a12341234dfa restart the httpd service executed successfully` seems hard to read for human or machine, maybe:
```
Script id=a12341234dfa (restart the httpd service) executed successfully
```

mcp-app/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  } else {
+    result += `Script ${toolRequestParams.script} executed`;
```
owtaylor: Again need some punctuation
```
Script `service httpd start` executed succesfully
```

src/linux_mcp_server/tools/run_script.py
```diff
@@ -182,25 +279,39 @@ async def run_script_readonly(
...
+) -> ToolResult:
+    script_details = script_store.get_script_details(id)
+    command = _wrap_script(script_details.script_type, script_details.script)
+    script_store.set_script_state(id, "executing")
+    content: list[ContentBlock] = []
+
+    try:
+        returncode, stdout, stderr = await execute_command(command, host=script_details.host)
+    except Exception as e:
+        raise ToolError(f"Execution failed: {e}")
```
owtaylor: The point here wasn't to turn the error into a ToolError - that happens automatically within FastMCP - the point was that the script state should be updated in this case and not left as executing.

mcp-app/src/run-script-app.tsx
Jazzcort: So it should be looking something like this? 
```typescript
  const validatedToolRequestParams = useMemo(() => {
    if (!toolRequestParams) {
      return undefined;
    }

    const parsedResult = McpAppToolParamsSchema.safeParse(toolRequestParams);

    if (parsedResult.success) {
      return parsedResult.data;
    } else {
      log.error(parsedResult.error);
      return undefined;
    }
  }, [toolRequestParams]);
```
and same for `validatedToolResult`

mcp-app/src/run-script-app.tsx
Jazzcort: Thanks for the catch! Will remove it!

mcp-app/src/run-script-app.tsx
Jazzcort: Yes, you are right! So I think we should do something like this?
```typescript
if (result.isError) {
  throw new Error(result.content[0].text)
}
```
Therefore, it can be caught at the catch block.

What do you think?

src/linux_mcp_server/tools/run_script.py
Jazzcort: Oh right! I wasn't thinking straight. So I think we should do this?
```python
    try:
        returncode, stdout, stderr = await execute_command(command, host=script_details.host)
    except Exception as e:
        returncode, stdout, stderr = 1, "", str(e)
```

owtaylor: I was thinking something more like:

```
    try:
        returncode, stdout, stderr = await execute_command(command, host=script_details.host)
    except Exception as e:
        script_store.set_script_state(id, "failure")
        raise
```

PR #271, opened 2026-02-10
Update python deps

samdoran:
Update the lock file, increase the minimum FastMCP and mkdocstrings version.

Closes #267.
Closes #280.

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -72,7 +72,7 @@ test = [
 docs = [
     "mkdocs>=1.6.1",
     "mkdocs-material>=9.7.1",
-    "mkdocstrings[python]>=1.0.2",
+    "mkdocstrings[python]>=1.0.3",
 ]
 
 [tool.hatch.build.targets.wheel]
--- a/uv.lock
+++ b/uv.lock
@@ -1197,7 +1197,7 @@ dev = [
 docs = [
     { name = "mkdocs", specifier = ">=1.6.1" },
     { name = "mkdocs-material", specifier = ">=9.7.1" },
-    { name = "mkdocstrings", extras = ["python"], specifier = ">=1.0.2" },
+    { name = "mkdocstrings", extras = ["python"], specifier = ">=1.0.3" },
 ]
 lint = [
     { name = "pyright" },


pyproject.toml
```diff
@@ -11,10 +11,10 @@ authors = [
...
 requires-python = ">=3.10"
 dependencies = [
-    "asyncssh[bcrypt] >= 2.22.0",
-    "fastmcp >=2.14.4, <3",
-    "pydantic-settings >= 2.12.0",
-    "pydantic >= 2.12.5",
+    "asyncssh[bcrypt] ~= 2.0",
+    "fastmcp ~= 2.0, <3",
+    "pydantic-settings ~= 2.0",
+    "pydantic ~= 2.0",
```
owtaylor: I'm not really following the reasoning here.

My understanding is that `fastmcp =~ 2.0` is roughly the same as `fastmcp >= 2.0, < 3` Which is a lot looser than `fastmcp >= 2.14.4, <3`  Do we want to let people use such old versions of fastmcp... presumably we have some dependencies on things added in 2.Y? 

And how does this reduce our need to update the versions here. If we don't care about 2.14.3 vs 2.14.4, we can run `uv lock --upgrade` and check in the result, no matter how we specify the constraints in pyproject.toml, right?

samdoran: The main goal is to set an accurate floor for the project requirements and reduce the number of PRs that renovate bot opens. I don't feel the need to update `pyproject.toml` for every single .Z release, but renovate bot does.

>My understanding is that fastmcp =~ 2.0 is roughly the same as fastmcp >= 2.0, < 3 Which is a lot looser than fastmcp >= 2.14.4, <3 Do we want to let people use such old versions of fastmcp... presumably we have some dependencies on things added in 2.Y?

That's correct. The `~=` operator essentially means any 2.Y.Z release. I don't believe we are relying on specific 2.X features but I would have to double check. We do rely or 2.Y features.

In practice, when people `pip install linux-mcp-server` there is little risk of them getting older releases. The greater risk is getting something too new.

> And how does this reduce our need to update the versions here.

The lock file serves a different purpose than `pyproject.toml`. End users installing with `pip` or `uv` will use the dependencies and constraints in `pyproject.toml`. Developers and container image builds use the lock file with `uv sync` for consistency and repeatability.

We will still need to update the lock file regularly since it provides detailed constraints for consistent environments. But we shouldn't have to update the `pyproject.toml` file as often as the lock file. Because `pyproject.toml` has such unnecessarily detailed requirements, we end up updating both all the time. This seems incorrect and these changes aim to fix that.

owtaylor: > The main goal is to set an accurate floor for the project requirements and reduce the number of PRs that renovate bot opens. I don't feel the need to update `pyproject.toml` for every single .Z release, but renovate bot does.

Is a floor of versions we haven't tested with an accurate floor? How do we know?

> > My understanding is that fastmcp =~ 2.0 is roughly the same as fastmcp >= 2.0, < 3 Which is a lot looser than fastmcp >= 2.14.4, <3 Do we want to let people use such old versions of fastmcp... presumably we have some dependencies on things added in 2.Y?
> 
> That's correct. The `~=` operator essentially means any 2.Y.Z release. I don't believe we are relying on specific 2.X features but I would have to double check. We do rely or 2.Y features.
>
> In practice, when people `pip install linux-mcp-server` there is little risk of them getting older releases. The greater risk is getting something too new.

I think `pip install linux-mcp-server` is inherently questionable. There's no guarantee at all that everything a user tries to install that way will have compatible versions. So then you get into the vendoring swamp. I'd much rather people used `uv tool install`.

> > And how does this reduce our need to update the versions here.
> 
> The lock file serves a different purpose than `pyproject.toml`. End users installing with `pip` or `uv` will use the dependencies and constraints in `pyproject.toml`. Developers and container image builds use the lock file with `uv sync` for consistency.

If someone (whether developer or end-user) is installing with uv, they are going to get the uv.lock versions, right? pip will obviously not care about what's in uv.lock...

> We will still need to update the lock file regularly since it provides detailed constraints for consistent environments. But we shouldn't have to update the `pyproject.toml` file as often as the lock file. Because `pyproject.toml` has such unnecessarily detailed requirements, we end up updating both all the time. This seems incorrect and these changes aim to fix that.

Where does our dependabot configuration live? I wouldn't really expect any difference between `>= 2.14, < 3` and `~= 2.14` in terms of what dependabot does. The change from a 3-digit version to a 2-digit version might make a difference though.

The main disadvantage of `~=` seems to be that there's no way with that to express `>= 2.14.4, < 3` `~= 2.14.4` means `>= 2.14.4, < 2.15` - but as long as we don't need to express that level of granularity, I'm find with switching to `~=`.

So, in summary:

 - I'm not comfortable with backing down the version requirements so far unless we have some way of testing that in CI
 - I'm fine with changing to `~=`
 - I'm fine with switching to two-digit specifiers if that reduces dependabot churn
 
[But if we don't want dependabot to update versions in pyproject.yaml, I think we can just configure it that way]

samdoran: > Is a floor of versions we haven't tested with an accurate floor? How do we know?

We don't, really. I don't know of any project that tests with every version of every dependency within the constraints range. It's really up to projects to adhere to conventions and not break things between versions. When we find incompatibilities, we can adjust the project requirements.

It's an imperfect system for sure.

>  I'd much rather people used uv tool install.

This is no different than `pip install`, it's just that `uv` is faster and takes care of creating a virtual environment for you. `uv install` installs from PyPI where there is no `uv.lock` file. You can see this with `uv tool install` or inspect after the fact with `uv pip list --python ~/.local/share/uv/tools/linux-mcp-server/bin/python`. You'll see it does not match the project lock file because that is not used by `uv` which installing/running tools.

<details>
  <summary>uv tool installation output</summary>
  
```
> uv tool install linux-mcp-server --python $(pyenv which python)
Resolved 87 packages in 119ms
Installed 87 packages in 119ms
 + annotated-doc==0.0.4
 + annotated-types==0.7.0
 + anyio==4.12.1
 + asyncssh==2.22.0
 + attrs==25.4.0
 + authlib==1.6.7
 + backports-tarfile==1.2.0
 + bcrypt==5.0.0
 + beartype==0.22.9
 + cachetools==7.0.1
 + certifi==2026.1.4
 + cffi==2.0.0
 + charset-normalizer==3.4.4
 + click==8.3.1
 + cloudpickle==3.1.2
 + croniter==6.0.0
 + cryptography==46.0.5
 + cyclopts==4.5.2
 + diskcache==5.6.3
 + dnspython==2.8.0
 + docstring-parser==0.17.0
 + docutils==0.22.4
 + email-validator==2.3.0
 + exceptiongroup==1.3.1
 + fakeredis==2.33.0
 + fastmcp==2.14.5  # one example that is newer than the lock file
 + h11==0.16.0
 + httpcore==1.0.9
 + httpx==0.28.1
 + httpx-sse==0.4.3
 + idna==3.11
 + importlib-metadata==8.7.1
 + jaraco-classes==3.4.0
 + jaraco-context==6.1.0
 + jaraco-functools==4.4.0
 + jsonref==1.1.0
 + jsonschema==4.26.0
 + jsonschema-path==0.3.4
 + jsonschema-specifications==2025.9.1
 + keyring==25.7.0
 + linux-mcp-server==1.3.0
 + lupa==2.6
 + markdown-it-py==4.0.0
 + mcp==1.26.0
 + mdurl==0.1.2
 + more-itertools==10.8.0
 + openapi-pydantic==0.5.1
 + opentelemetry-api==1.39.1
 + packaging==26.0
 + pathable==0.4.4
 + pathvalidate==3.3.1
 + platformdirs==4.5.1
 + prometheus-client==0.24.1
 + py-key-value-aio==0.3.0
 + py-key-value-shared==0.3.0
 + pycparser==3.0
 + pydantic==2.12.5
 + pydantic-core==2.41.5
 + pydantic-settings==2.12.0
 + pydocket==0.17.7
 + pygments==2.19.2
 + pyjwt==2.11.0
 + pyperclip==1.11.0
 + python-dateutil==2.9.0.post0
 + python-dotenv==1.2.1
 + python-json-logger==4.0.0
 + python-multipart==0.0.22
 + pytz==2025.2
 + pyyaml==6.0.3
 + redis==7.1.1
 + referencing==0.36.2
 + requests==2.32.5
 + rich==14.3.2
 + rich-rst==1.3.2
 + rpds-py==0.30.0
 + shellingham==1.5.4
 + six==1.17.0
 + sortedcontainers==2.4.0
 + sse-starlette==3.2.0
 + starlette==0.52.1
 + typer==0.23.0
 + typing-extensions==4.15.0
 + typing-inspection==0.4.2
 + urllib3==2.6.3
 + uvicorn==0.40.0
 + websockets==16.0
 + zipp==3.23.0
Installed 1 executable: linux-mcp-server
``` 
</details>


This is why we use `uv sync` for development setup and container image builds. That command uses the lock file.

> Where does our dependabot configuration live? 

`renovate.json` is the repo root


I think I'll just leave it like it is and we can go back to updating both the lock file and `pyproject.toml` file. I think that's an odd way to have project specifiers but it's not bad. Just more work to maintain.

owtaylor: Ah, yeah, I didn't think about that `uv tool install` doesn't have access to uv.lock. But still it's quite different (and more likely to work) to install the *newest* in-range version of everything in a venv, compared to what `pip install` will do.

> I think I'll just leave it like it is and we can go back to updating both the lock file and `pyproject.toml` file. I think that's an odd way to have project specifiers but it's not bad. Just more work to maintain.

If you don't want renovate to bump the minimum versions, then we should change `"rangeStrategy": "bump"` to something else in `renovate.json`?

https://docs.renovatebot.com/configuration-options/#rangestrategy

I think their recommendation would be to use "update-lockfile" combined with something like:

```json
{
  "packageRules": [
    {
      "matchPackageNames": ["fastmcp"],
      "matchUpdateTypes": ["major"],
      "enabled": false
    }
  ]
}
```

to avoid fastmcp getting renovate-updated to v3.

samdoran: How is `pip install` different than `uv tool install` with regards to dependency resolution? I don't see any difference between the installed versions just some subtle differences in the package names.

```
> uv pip freeze --python ~/.local/share/uv/tools/linux-mcp-server/bin/python | string lower | sort > uv.txt
> pip freeze | string lower | sort > pip.txt
> diff --unified {pip,uv}.txt
--- pip.txt	2026-02-12 09:13:30
+++ uv.txt	2026-02-12 09:13:16
@@ -4,7 +4,7 @@
 asyncssh==2.22.0
 attrs==25.4.0
 authlib==1.6.7
-backports.tarfile==1.2.0
+backports-tarfile==1.2.0
 bcrypt==5.0.0
 beartype==0.22.9
 cachetools==7.0.1
@@ -18,7 +18,7 @@
 cyclopts==4.5.2
 diskcache==5.6.3
 dnspython==2.8.0
-docstring_parser==0.17.0
+docstring-parser==0.17.0
 docutils==0.22.4
 email-validator==2.3.0
 exceptiongroup==1.3.1
@@ -29,10 +29,10 @@
 httpx-sse==0.4.3
 httpx==0.28.1
 idna==3.11
-importlib_metadata==8.7.1
-jaraco.classes==3.4.0
-jaraco.context==6.1.0
-jaraco.functools==4.4.0
+importlib-metadata==8.7.1
+jaraco-classes==3.4.0
+jaraco-context==6.1.0
+jaraco-functools==4.4.0
 jsonref==1.1.0
 jsonschema-path==0.3.4
 jsonschema-specifications==2025.9.1
@@ -50,11 +50,11 @@
 pathable==0.4.4
 pathvalidate==3.3.1
 platformdirs==4.5.1
-prometheus_client==0.24.1
+prometheus-client==0.24.1
 py-key-value-aio==0.3.0
 py-key-value-shared==0.3.0
 pycparser==3.0
-pydantic_core==2.41.5
+pydantic-core==2.41.5
 pydantic-settings==2.12.0
 pydantic==2.12.5
 pydocket==0.17.7
@@ -79,7 +79,7 @@
 sse-starlette==3.2.0
 starlette==0.52.1
 typer==0.23.0
-typing_extensions==4.15.0
+typing-extensions==4.15.0
 typing-inspection==0.4.2
 urllib3==2.6.3
 uvicorn==0.40.0
```

owtaylor: The difference is that unless you are doing `pip install --upgrade-strategy=eager`, if libraries are already in the destination Python environment, and meet our minimum requirements, they won't be upgraded. So pip install into a fresh venv should be basically identical to what you get with `uv tool install`, but `pip install --user` could be very different.

samdoran: That's really good to know. I didn't think about that. Thank you for explaining. I was incorrectly thinking of the user virtual environment as equivalent to a clean virtual environment but it most certainly is not.

I would much rather tell folks to create a dedicated virtual environment and install in that, but those instructions get tedious.

I suppose that is why everyone just wants to use `uv` 🤪.

PR #273, opened 2026-02-11
Avoid StrEnum for script_type

owtaylor:
We want to use Claude Desktop as a second reference for our work with mcp-apps, but it has trouble with schemas for tools that reference `$defs`. Avoid that by making `script_type` a `Literal[]` rather than a `StrEnum`.

Future versions of FastMCP may automatically dereference references (https://github.com/jlowin/fastmcp/pulls/2814) but that doesn't currently happen as of v2.14.5 (https://github.com/jlowin/fastmcp/issues/3153).

(This is not the only usage of StrEnum in our tool interface, but since problems with the other usages aren't immediately affecting us, and the eventual fix is probably on the FastMCP side instead of changing our definitions, let's restrict the changes to this for now.)


subpop (approved):
This looks reasonable to me.

PR #274, opened 2026-02-11
Merge main into arbitrary-commands (2025-02-11)

owtaylor:
This merges changes in head (such as transport options) into experiments/arbitrary-commands.

Rationale for merging rather than rebasing:

 - Easier to review the merge
 - People with `arbitrary-commands` checked out can just pull the new version, instead of having to reset to the rebased version.
 - Repeated merges are much easier than repeated rebases - you typically only have to resolve conflicts *once*

We'll probably want to do a rebase before merging into main to avoid complicated git history.


r0x0d (approved):
it looks good to me on first glance, but @subpop would be a better person to have a final ack on this.

owtaylor:
I think it's pretty safe, so if you've given it a glance over, that should be good enough. 

> it looks good to me on first glance, but @subpop would be a better person to have a final ack on this.

I'm willing to push this on one glance over :-) ... I think it's pretty safe. [Pushing from the cli, since we can't squash-and-merge here]

PR #275, opened 2026-02-11
[RSPEED-2264] Return structured output from log tools

samdoran:
Add new models `LogFile` and `JournalLogs` to format the returned data from `read_log_file` and `get_journal_logs` and remove formatting functions that are no longer necessary.

<details>
  <summary>Both tools now have output schema.</summary>
  
```json
    {
      "name": "get_journal_logs",
      "title": "Get journal logs",
      "description": "Get systemd journal logs.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "unit": {
            "default": "",
            "description": "Filter by systemd unit name or pattern",
            "examples": [
              "sshd.service",
              "nginx",
              "httpd",
              "systemd-*",
              "audit*"
            ],
            "type": "string"
          },
          "priority": {
            "default": "",
            "description": "Filter by syslog priority level (0-7), name, or range",
            "examples": [
              "err",
              "warning",
              "info",
              "debug",
              "3",
              "err..warning"
            ],
            "type": "string"
          },
          "since": {
            "default": "",
            "description": "Filter entries since specified time (absolute or relative)",
            "examples": [
              "today",
              "yesterday",
              "-1h",
              "-30m",
              "-7d",
              "2025-01-15 10:00:00"
            ],
            "type": "string"
          },
          "transport": {
            "anyOf": [
              {
                "$ref": "#/$defs/Transport"
              },
              {
                "type": "null"
              }
            ],
            "default": null,
            "description": "Filter by journal transport (e.g., 'audit' for audit logs, 'kernel' for kernel messages, 'syslog' for syslog messages)"
          },
          "lines": {
            "default": 100,
            "description": "Number of log lines to retrieve. Default: 100",
            "maximum": 10000,
            "minimum": 1,
            "type": "integer"
          },
          "host": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "default": null,
            "description": "Remote host to connect to via SSH"
          }
        },
        "$defs": {
          "Transport": {
            "description": "Valid journalctl transport types for filtering journal entries.\n\nTransports identify the source/mechanism that submitted log messages to the journal.\n\nAUDIT - Linux audit subsystem messages (security events, syscall auditing)\nDRIVER - Kernel driver messages logged via dev_printk().\nJOURNAL - Messages logged directly to the journal via sd_journal_* APIs.\nKERNEL - Kernel ring buffer messages (dmesg/printk).\nSTDOUT - stdout/stderr from services with StandardOutput/StandardError=journal.\nSYSLOG - Messages received via the syslog socket (/dev/log).",
            "enum": [
              "audit",
              "driver",
              "journal",
              "kernel",
              "stdout",
              "syslog"
            ],
            "type": "string"
          }
        }
      },
      "outputSchema": {
        "type": "object",
        "properties": {
          "entries": {
            "items": {
              "type": "string"
            },
            "type": "array"
          },
          "unit": {
            "default": "",
            "type": "string"
          },
          "path": {
            "anyOf": [
              {
                "format": "path",
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "default": null
          },
          "lines_count": {
            "type": "integer"
          }
        },
        "required": [
          "entries"
        ]
      },
      "annotations": {
        "readOnlyHint": true
      },
      "_meta": {
        "_fastmcp": {
          "tags": [
            "journal",
            "logs",
            "systemd",
            "troubleshooting"
          ]
        }
      }
    },

...

    {
      "name": "read_log_file",
      "title": "Read log file",
      "description": "Read a specific log file.",
      "inputSchema": {
        "type": "object",
        "properties": {
          "log_path": {
            "description": "Absolute path to the log file (must be in allowed list)",
            "examples": [
              "/var/log/messages",
              "/var/log/secure",
              "/var/log/audit/audit.log",
              "/var/log/dnf.log"
            ],
            "format": "path",
            "type": "string"
          },
          "lines": {
            "default": 100,
            "description": "Number of lines to retrieve from the end.",
            "maximum": 10000,
            "minimum": 1,
            "type": "integer"
          },
          "host": {
            "anyOf": [
              {
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "default": null,
            "description": "Remote host to connect to via SSH"
          }
        },
        "required": [
          "log_path"
        ]
      },
      "outputSchema": {
        "type": "object",
        "properties": {
          "entries": {
            "items": {
              "type": "string"
            },
            "type": "array"
          },
          "unit": {
            "default": "",
            "type": "string"
          },
          "path": {
            "anyOf": [
              {
                "format": "path",
                "type": "string"
              },
              {
                "type": "null"
              }
            ],
            "default": null
          },
          "lines_count": {
            "type": "integer"
          }
        },
        "required": [
          "entries"
        ]
      },
      "annotations": {
        "readOnlyHint": true
      },
      "_meta": {
        "_fastmcp": {
          "tags": [
            "files",
            "logs",
            "troubleshooting"
          ]
        }
      }
    },
... 
</details>

### DIFF
--- a/tests/tools/test_logs.py
+++ b/tests/tools/test_logs.py
@@ -27,7 +27,7 @@ class TestGetJournalLogs:
             (
                 {},
                 ["-n", "100", "--no-pager"],
-                {"unit": ""},
+                {"unit": None},
             ),
             # Unit filter
             (
@@ -39,19 +39,19 @@ class TestGetJournalLogs:
             (
                 {"priority": "err"},
                 ["--priority", "err"],
-                {"unit": ""},
+                {"unit": None},
             ),
             # Since filter
             (
                 {"since": "today"},
                 ["--since", "today"],
-                {"unit": ""},
+                {"unit": None},
             ),
             # Custom line count
             (
                 {"lines": 50},
                 ["-n", "50"],
-                {"unit": ""},
+                {"unit": None},
             ),
             # All filters combined
             (
@@ -63,17 +63,17 @@ class TestGetJournalLogs:
             (
                 {"transport": "audit"},
                 ["_TRANSPORT=audit"],
-                {"unit": ""},
+                {"unit": None},
             ),
             (
                 {"transport": "kernel"},
                 ["_TRANSPORT=kernel"],
-                {"unit": ""},
+                {"unit": None},
             ),
             (
                 {"transport": "audit", "priority": "err", "lines": 50},
                 ["_TRANSPORT=audit", "--priority", "err", "-n", "50"],
-                {"unit": ""},
+                {"unit": None},
             ),
         ],
     )

### CODE REVIEW

File: src/linux_mcp_server/tools/logs.py [hunk 1]
```diff
@@ -12,8 +12,8 @@
 from linux_mcp_server.audit import log_tool_call
 from linux_mcp_server.commands import get_command
 from linux_mcp_server.config import CONFIG
-from linux_mcp_server.formatters import format_journal_logs
-from linux_mcp_server.formatters import format_log_file
+from linux_mcp_server.models import JournalLogs
+from linux_mcp_server.models import LogFile
```
owtaylor: I find the naming here a inconsistent- both of these types are "a set of log lines from a source" - maybe `JournalLogs` and `LogFileLogs` (Or `...Entries` instead of `...Logs`)

src/linux_mcp_server/models.py
```diff
@@ -184,3 +185,19 @@ def human_values(self):
...
+
+
+### Log models ###
+class JournalLogs(BaseModel):
+    entries: list[str]
+    unit: str | None = None
+    priority: str | None = None
+    since: str | None = None
+    transport: str | None = None
+    lines_count: int = Field(default_factory=field_length("entries"))
```
owtaylor: I do somewhat wonder what is the utility here of these extra fields - of repeating the parameters of the tool call and adding `lines_count`? how is that going to make the model perform better? But my guess is that it's pretty much a wash - will have no real effect on performance either way. Wish we had thorough evals :-)

With the journal, we do have the *ability* to make entries not just a List[str], but to have subset of `journalctl -o json` - however, I think that's likely to degrade performance - the model will have been trained on human formatted logs, not JSON logs.

owtaylor (changes requested):
I went through this pretty carefully, and it looks good to me. The only thing I'd like to see changed is a nit-picky one - the names of the structured result types seem inconsistent to me.

src/linux_mcp_server/tools/logs.py
samdoran: Part of me wondered whether these could be combined into a single object for lines of log output, but `JournalLogs` originally had a bunch of different attributes in the output I thought would be good to preserve in the structured return.

src/linux_mcp_server/models.py
samdoran: > I do somewhat wonder what is the utility here of these extra fields - of repeating the parameters of the tool call and adding lines_count?

I thought the same thing about the parameters and mainly aimed to preserve what was already in the text output. Since the LLM has both the input parameters and the result in the context, maybe it is unnecessary to have them in the output. That would make it much easier to create a single object to house the list of log lines.

From an object design perspective, I think it makes sense to have the attributes of how the log lines were derived in addition to the log lines. It may not matter for how the object is used in this exact context, but it does make the object itself a more complete entity. If the object just has on attribute that is a list of strings, it's not doing much to justify its existence.

The total count I figured would be helpful to the LLM and save it from having to do math or just add some value to the object beyond just containing a list of strings.

samdoran: > With the journal, we do have the ability to make entries not just a List[str], but to have subset of journalctl -o json

I didn't know `journalctl` support JSON output. That may be helpful but only if there is a consistent schema and only if we need to manipulate the results before the final return.

If we are _only_ returning lines and not doing any filtering, it's probably not worth the cost in time and complexity to serialize the individual lines even thought that would be much more structured.

That's something we can keep in mind for the future, though. It's always very handy when tools return JSON because it's so easy to turn that JSON into objects.

src/linux_mcp_server/tools/logs.py
samdoran: I combined these into a single class, removed most of the attributes, and used it for both tool returns. Let me know what you think.

src/linux_mcp_server/models.py
samdoran: I went ahead and removed most of those fields.

samdoran:
That's a good point. Let me see if I can fix that up. I think the keys should always be in the return schema but `null` would be more appropriate that an empty string for JSON I think. Empty strings as falsy values are more Pythonic than idiomatic JavaScript I believe.

samdoran:
How does this look:
```
In [5]: entries = LogEntries(entries=['one', 'two'])

In [6]: entries.model_dump_json()
Out[6]: '{"entries":["one","two"],"unit":null,"path":null,"lines_count":2}'
```

Not sure if we want to keep the empty string if provided, though. Might be splitting hairs at this point.
```
In [7]: LogEntries(entries=['one', 'two',], unit='').model_dump_json()
Out[7]: '{"entries":["one","two"],"unit":null,"path":null,"lines_count":2}'
```

samdoran:
> I don't like "" as a flag value for unspecified

The thinking behind an empty string as the parameter default is specifically because we do _not_ need a sentinel/flag value in this particular case. It simplifies the code since we know it will be a string and avoids having to test or inspect the value before calling string methods on it.

It helps avoid things like this:
```python
if unit is not None:
    unit.lower()  # or other string manipulations that would fail on a None object
```

We are giving up a clear sentinel value for unspecified, but I don't think that is necessary here.

PR #281, opened 2026-02-17
Update doc string for execute_command

samdoran:
This method does not raise `ToolError`.


src/linux_mcp_server/connection/ssh.py
```diff
@@ -354,7 +354,6 @@ async def execute_command(
     Raises:
         ValueError: If host is provided without username
         ConnectionError: If remote connection fails
```
Jazzcort, line 356: Just one quick question here. Do you think it's relevant to also mention it raises `FileNotFoundError` when bin can not be found here?

PR #288, opened 2026-02-24
[RSPEED-2331] Figure out how to do different things for clients w/wo mcp-app support

Jazzcort:
Since both https://github.com/block/goose/pull/6927 and https://github.com/block/goose/pull/6931 are merged into Goose, we can now detect mcp app compatibility from Goose client during initialize phase.

feat: dynamically inject tools based on client UI capabilities
- Change `use_mcp_apps` config default from `False` to `None` to enable auto-detection, while preserving manual overrides.
- Add `MCP_UI_EXTENSION` constant (`io.modelcontextprotocol/ui`) to check for UI extension support.
- Register both interactive and non-interactive script modification tools by default.
- Update `DynamicDiscoveryMiddleware` to inspect `client_params.capabilities.extensions` during `on_list_tools`.
- Dynamically filter out either `run_script_modify` or `run_script_modify_interactive` based on whether the client supports the `MCP_APP_MIME_TYPE`

### DIFF
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -196,16 +196,6 @@ _low_level_server.request_handlers[ReadResourceRequest] = _read_resource_with_me
 from linux_mcp_server.tools import *  # noqa: E402, F403
 
 
-# With mcp-app feature
-mcp.add_tool(run_script_modify_interactive)
-mcp.add_tool(get_execution_state)
-mcp.add_tool(execute_script)
-mcp.add_tool(reject_script)
-
-# Without mcp-app feature
-mcp.add_tool(run_script_modify)
-
-
 class DynamicDiscoveryMiddleware(Middleware):
     async def on_list_tools(self, context: MiddlewareContext, call_next):
         tools = await call_next(context)
@@ -240,9 +230,9 @@ class DynamicDiscoveryMiddleware(Middleware):
         # The configuration can overwrite the MCP app support detection, so we have the flexibility to
         # manually turn the Mcp app feature on/off for developing/testing purposes.
         if CONFIG.use_mcp_apps or (CONFIG.use_mcp_apps is None and MCP_APP_MIME_TYPE in mime_types):
-            filtered_tools = [t for t in filtered_tools if t.name != "run_script_modify"]
+            filtered_tools = [t for t in filtered_tools if "mcp_apps_exclude" not in t.tags]
         else:
-            filtered_tools = [t for t in filtered_tools if t.name != "run_script_modify_interactive"]
+            filtered_tools = [t for t in filtered_tools if "mcp_apps_only" not in t.tags]
 
         return filtered_tools
 
--- a/src/linux_mcp_server/tools/run_script.py
+++ b/src/linux_mcp_server/tools/run_script.py
@@ -8,7 +8,6 @@ from dataclasses import dataclass
 
 from fastmcp import Context
 from fastmcp.exceptions import ToolError
-from fastmcp.tools.tool import Tool
 from fastmcp.tools.tool import ToolResult
 from mcp.types import ContentBlock
 from mcp.types import TextContent
@@ -286,7 +285,14 @@ class ExecuteScriptResult:
     output: str
 
 
-async def _execute_script(
+@mcp.tool(
+    tags={"run_script", "hidden_from_model"},
+    description="Execute a script; this is only available to the our mcp-app",
+    meta={"ui": {"visibility": ["app"]}},
+)
+@log_tool_call
+@disallow_local_execution_in_containers
+async def execute_script(
     id: t.Annotated[str, Field(description="The associated ID of the script to be executed")],
 ) -> ToolResult:
     script_details = script_store.get_script_details(id)
@@ -316,33 +322,30 @@ async def _execute_script(
     return ToolResult(content=content, structured_content=asdict(result))
 
 
-execute_script = Tool.from_function(
-    _execute_script,
-    name="execute_script",
+@mcp.tool(
     tags={"run_script", "hidden_from_model"},
-    description="Execute a script; this is only available to the our mcp-app",
+    description="Reject a script; this is only available to the our mcp-app",
     meta={"ui": {"visibility": ["app"]}},
 )
-
-
-async def _reject_script(
+@log_tool_call
+@disallow_local_execution_in_containers
+async def reject_script(
     id: t.Annotated[str, Field(description="The associated ID of the script to be rejected")],
 ):
     script_store.set_script_state(id, "rejected-user")
 
 
-reject_script = Tool.from_function(
-    _reject_script,
-    name="reject_script",
-    tags={"run_script", "hidden_from_model"},
-    description="Reject a script; this is only available to the our mcp-app",
-    meta={"ui": {"visibility": ["app"]}},
+@mcp.tool(
+    tags={"run_script", "mcp_apps_only"},
+    title="Propose to run a script that modifies system",
+    description=RUN_SCRIPT_MODIFY_INTERACTIVE,
+    annotations=ToolAnnotations(destructiveHint=True),
+    output_schema=RunScriptInteractiveResult.model_json_schema(),
+    meta={"ui": {"resourceUri": RUN_SCRIPT_APP_URI}},
 )
-
-
 @log_tool_call
 @disallow_local_execution_in_containers
-async def _run_script_modify_interactive(
+async def run_script_modify_interactive(
     ctx: Context,
     description: t.Annotated[
         str,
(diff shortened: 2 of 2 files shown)


src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
 
         return filtered_tools
 
+    async def on_initialize(self, context: MiddlewareContext, call_next):
+        assert isinstance(context.message, InitializeRequest), (
```
Jazzcort: Not sure if using `assert` here is the best solution. Open to change it if there is a better way of doing this.

src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+        # This is a temperory implementation. When "extensions" property is officially introduced in FastMcp,
+        # it will be easier for us to verify mime type.
+        capabilities = context.message.params.capabilities.model_dump()
+        extensions = capabilities.get("extensions") or {}
+        mcp_ui_extension = extensions.get(MCP_UI_EXTENSION) or {}
+        mime_types = mcp_ui_extension.get("mimeTypes") or []
+
+        # The configuration can overwrite the MCP app support detection, so we have the flexibility to
+        # manually turn the Mcp app feature on/off for developing/testing purposes.
+        if CONFIG.use_mcp_apps or MCP_APP_MIME_TYPE in mime_types:
```
Jazzcort: I think it's helpful to keep the env configuration working so we can easily switch the feature on/off for local testing purpose.

src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+        fastmcp_context = context.fastmcp_context
+        assert fastmcp_context is not None, (
+            "FastMCP framework error: context.fastmcp_context should not be None inside on_initialize"
+        )
+        fastmcp_instance = fastmcp_context.fastmcp
+
+        # This is a temperory implementation. When "extensions" property is officially introduced in FastMcp,
+        # it will be easier for us to verify mime type.
+        capabilities = context.message.params.capabilities.model_dump()
+        extensions = capabilities.get("extensions") or {}
```
owtaylor: You should be able to just do `getattr(capabilities, "extensions", {})`

src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+
+        fastmcp_context = context.fastmcp_context
+        assert fastmcp_context is not None, (
+            "FastMCP framework error: context.fastmcp_context should not be None inside on_initialize"
+        )
+        fastmcp_instance = fastmcp_context.fastmcp
+
+        # This is a temperory implementation. When "extensions" property is officially introduced in FastMcp,
+        # it will be easier for us to verify mime type.
+        capabilities = context.message.params.capabilities.model_dump()
```
owtaylor: As discussed in a private slack conversation, the extra("allow") was removed from python-sdk's types in https://github.com/modelcontextprotocol/python-sdk/commit/a1d330de9f625ae153bfc4c322731eb1fa0087ad - I think we assume that `extensions` will be be there properly once v2 of python-sdk is released - hopefully in a compatible way.

Note that this is *not* FastMCP - it's the  base `mcp` module:

```
        # For python-sdk -1.x, count on extensibility of protocol types - while this is being
        # removed for v2, hopefully extensions will be there properly.
```

src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+    async def on_initialize(self, context: MiddlewareContext, call_next):
+        assert isinstance(context.message, InitializeRequest), (
+            f"FastMCP routing error: Expected InitializeRequest in on_initialize, got {type(context.message)}"
+        )
+
+        fastmcp_context = context.fastmcp_context
+        assert fastmcp_context is not None, (
+            "FastMCP framework error: context.fastmcp_context should not be None inside on_initialize"
+        )
+        fastmcp_instance = fastmcp_context.fastmcp
```
owtaylor: Isn't this just the same as our global `mcp` variable? Retrieving it this way makes it look like something special is going on.

src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+
+        # The configuration can overwrite the MCP app support detection, so we have the flexibility to
+        # manually turn the Mcp app feature on/off for developing/testing purposes.
+        if CONFIG.use_mcp_apps or MCP_APP_MIME_TYPE in mime_types:
+            fastmcp_instance.add_tool(run_script_modify_interactive)
+            fastmcp_instance.add_tool(get_execution_state)
+            fastmcp_instance.add_tool(execute_script)
+            fastmcp_instance.add_tool(reject_script)
+        else:
+            fastmcp_instance.add_tool(run_script_modify)
```
owtaylor: This isn't going to work with http transport. Can we A) add everything B) set a flag on the context C) Filter list_tools? If that ends up being too hard, we might need to upgrade to FastMCP 3.0 for it's builtin tool visibility support.

src/linux_mcp_server/server.py
```diff
@@ -224,6 +209,36 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+            f"FastMCP routing error: Expected InitializeRequest in on_initialize, got {type(context.message)}"
+        )
+
+        fastmcp_context = context.fastmcp_context
+        assert fastmcp_context is not None, (
+            "FastMCP framework error: context.fastmcp_context should not be None inside on_initialize"
+        )
+        fastmcp_instance = fastmcp_context.fastmcp
+
+        # This is a temperory implementation. When "extensions" property is officially introduced in FastMcp,
```
owtaylor:
```suggestion
        # python-sdk v1.x uses pydantic's extra("allow") capability to extend the protocol types
        # with unknown fields. This is being removed for v2.0
        #    https://github.com/modelcontextprotocol/python-sdk/pull/1937
        # but hopefully that will support `extensions` properly when released.
```

src/linux_mcp_server/server.py
owtaylor: I think you can simply do:

```
    extensions = capabilities.getattr("extensions", {})
```

src/linux_mcp_server/server.py
owtaylor: Yes, but I'd make use_mcp_apps `bool | None` and make this:

```python
   if CONFIG.use_mcp_apps or (CONFIG.use_mcp_apps is None and MCP_APP_MIME_TYPE in mime_types):
```

so you can switch the feature *off* and not just on.

src/linux_mcp_server/server.py
owtaylor: narrowing with `assert` seems fine to me.

src/linux_mcp_server/server.py
Jazzcort: Thanks for the catch! 😁 Will change!

src/linux_mcp_server/server.py
Jazzcort: Yes, it's the same as our global `mcp`. I was thinking "what if we don't have the global mcp access" when I implemented this. 😅 Will change it to use the global variable so it looks neat.

src/linux_mcp_server/server.py
Jazzcort: It would be great if we can upgrade it to 3.0. 
>B) set a flag on the context

@owtaylor  Are you referring to `set_status` to set a flag? Last time I tried it, the status was not persistent among requests for 2.x - I set status for Client A in the initialize request but can't get the status I set in the following tool call request. I have to manually create a store to save the session_id. However, `set_status` works fine for 3.0.

Let me test `set_status` again for both 2.x and 3.0! Will report to you later!

src/linux_mcp_server/server.py
```diff
@@ -197,22 +196,16 @@ async def _read_resource_with_meta(req: ReadResourceRequest):
...
-# This is a temporary implementation. The injection logic should be moved to the
-# `on_initialize` handler in `DynamicDiscoveryMiddleware` once Goose starts
-# providing `mcp-app` compatibility during the initialize request.
-if CONFIG.use_mcp_apps:
-    mcp.add_tool(run_script_modify_interactive)
-    mcp.add_tool(get_execution_state)
-    mcp.add_tool(execute_script)
-    mcp.add_tool(reject_script)
+# With mcp-app feature
+mcp.add_tool(run_script_modify_interactive)
```
Jazzcort: @owtaylor Since we're now adding all tools, there is actually no need to transform functions into Tool and inject tools like this. Do you think we should change these back to decorator format `@mcp.tool`?

owtaylor: Yes, I do

src/linux_mcp_server/server.py
```diff
@@ -222,6 +215,35 @@ async def on_list_tools(self, context: MiddlewareContext, call_next):
...
+        extensions = getattr(capabilities, "extensions", {})
+        mcp_ui_extension = extensions.get(MCP_UI_EXTENSION) or {}
+        mime_types = mcp_ui_extension.get("mimeTypes") or []
+
+        # The configuration can overwrite the MCP app support detection, so we have the flexibility to
+        # manually turn the Mcp app feature on/off for developing/testing purposes.
+        if CONFIG.use_mcp_apps or (CONFIG.use_mcp_apps is None and MCP_APP_MIME_TYPE in mime_types):
+            filtered_tools = [t for t in filtered_tools if t.name != "run_script_modify"]
+        else:
+            filtered_tools = [t for t in filtered_tools if t.name != "run_script_modify_interactive"]
```
owtaylor: Even though there's only one tool, I think I'd still do it with tags:

```suggestion
        if CONFIG.use_mcp_apps or (CONFIG.use_mcp_apps is None and MCP_APP_MIME_TYPE in mime_types):
            filtered_tools = [t for t in filtered_tools if "mcp_apps_exclude" not in t.tags]
        else:
            filtered_tools = [t for t in filtered_tools if "mcp_apps_only" not in t.tags]
```

PR #290, opened 2026-02-27
run_script_modify_interactive: use default FastMCP rendering of content

owtaylor:
In the case where the gatekeeper rejects a script, we were returning content=[], structuredContent={...} to the model, which at least for some providers, results in the model seeing an empty result from the tool call. By instead return content=None, FastMCP will render structuredContent as the content, letting the model know what happened.


Jazzcort (approved):
LGTM! 😁

PR #296, opened 2026-03-09
fix: update model tool instructions and gatekeerper prompt

panyamkeerthana:
- added INSTRUCTIONS_RUN_SCRIPT_MODIFY
- updated toolset logic to us instructions based on use_mcp_apps config
- added note to INSTRUCTIONS_BOTH about mutual run_script_modify* tool exclusivity
- update gatekeeper prompt for remote script execution validation

### DIFF
--- a/src/linux_mcp_server/gatekeeper/check_run_script.py
+++ b/src/linux_mcp_server/gatekeeper/check_run_script.py
@@ -76,6 +76,11 @@ as it fails cleanly. The model may have previously investigated the system
 and found out things like the operating system version and configuration
 file paths.
 
+The description of the script may reference querying or modifying a specific host;
+this is *OK* - the MCP server will take care of executing the script on the
+appropriate host. The script does not need to, for example, use SSH to connect to
+the target host.
+
 Return a JSON object with the following keys, `status` (required) and `detail` (optional).
 
 status should be one of the following:


src/linux_mcp_server/server.py
```diff
@@ -92,6 +118,9 @@
 
 - **run_script_readonly:** Run a Python or Bash script that only inspects the system. Use for queries and data collection.
 - **run_script_modify:** Run a Python or Bash script that may change files or configuration. Use only when changes are required.
+- **run_script_modify_interactive:** Propose a script that modifies the system. The user will be asked for approval before execution.
+
+Note: Only one of the modify variants (run_script_modify or run_script_modify_interactive) will be available at any given time, depending on the LINUX_MCP_USE_MCP_APPS configuration.
```
owtaylor: I'm just very uncertain that weaker models and/or models where tool calling generation isn't forced by the provided list of tools will figure this out. Also the model will have no idea what "LINUX_MCP_USE_MCP_APPS configuration" is.

To prioritize not confusing the model above all else, I came up with #311 - it is super-hacky for FastMCP 2, but once we switch to FastMCP 3 it will only be a bit hacky.

If we go that approach, this PR will reduce to just the gatekeeper part.

src/linux_mcp_server/server.py
```diff
@@ -120,7 +149,7 @@
         instructions = INSTRUCTIONS_FIXED
         kwargs["exclude_tags"] = {"run_script"}
     case Toolset.RUN_SCRIPT:
-        instructions = INSTRUCTIONS_RUN_SCRIPT
+        instructions = INSTRUCTIONS_RUN_SCRIPT_MODIFY if CONFIG.use_mcp_apps else INSTRUCTIONS_RUN_SCRIPT
```
owtaylor: The default value of CONFIG.use_mcp_apps is None - which means "auto-detect", so we'd need to use the combined instructions unless mcp-app usage was explicitly disabled - this even more inclines me to the approach in #311

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -46,6 +46,10 @@ def get_model() -> str:
    No new repositories may be added.
  * Except for installing software from pre-configured repositories,
    nothing may be downloaded from the internet.
+ * Scripts can be executed on remote hosts. The remote connection is
+   handled automatically by the execution infrastructure and do not
+   need to contain SSH commands. The script should contain only the
+   commands to run on the target system.
```
owtaylor, line 49: This doesn't seem to fit under General Policies - it should be probably be next to "It is *OK* if the script does not consider every possible failure mode, as long as it fails cleanly..."

The wording is a bit off - there seems to be a few missing words "and do not need to contain SSH commands". Maybe make it specifically about the description.

```
The description of the script may reference modifying a specific host;
this is *OK* - the MCP server will take care of executing the script
on the appropriate host. The script does not need to, for example,
use SSH to connect to the target host.
```

(Eventually we might want the gatekeeper to validate that any host mentioned in the description matches where the script is going to be executed. Like if the description said "Change testvm to allow logging in without a password, but it actually ran with host=None. But let's leave that to the future, especially since it will conflict with the gatekeeper eval work.)

src/linux_mcp_server/server.py
panyamkeerthana: Just went through the PR and honestly it's a much better workaround, thanks for coming up w that! 😊

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -76,6 +76,11 @@ def get_model() -> str:
 and found out things like the operating system version and configuration
 file paths.
 
+The description of the script may reference modifying a specific host; this is
```
owtaylor:
```suggestion
The description of the script may reference querying or modifying a specific host; this is
```

owtaylor: Hmm, I didn't think about this being used for run_script_readonly as well as run-script_modify - tiny change with that in mind.

PR #309, opened 2026-03-19
feat: add gatekeeper evaluation framework

owtaylor:
Add scripts and test cases for evaluating the gatekeeper's script approval/denial behavior. Includes extract-from-goose.py for extracting test cases from Goose session logs, run-eval.py for running evaluations against a model, and initial test cases (crafted, adhoc, and scenario-based).

```
Running adhoc/rpmfusion-policy.yaml  [####################################]  100%
Running crafted/bad-description.yaml  [####################################]  100%          
Running scenarios/selinux-port-denial-1.yaml  [####################################]  100%          
Running scenarios/selinux-port-denial-2.yaml  [####################################]  100%          
[...details of exceptions/failing cases...]
file                                  same  ok_to_forbidden  forbidden_to_ok  other_mismatch  exception
------------------------------------  ----  ---------------  ---------------  --------------  ---------
adhoc/rpmfusion-policy.yaml           1     0                0                0               0        
crafted/bad-description.yaml          2     0                0                0               0        
scenarios/selinux-port-denial-1.yaml  3     0                0                0               1        
scenarios/selinux-port-denial-2.yaml  8     0                0                0               1        
------------------------------------  ----  ---------------  ---------------  --------------  ---------
TOTAL                                 14    0                0                0               2        
```

### DIFF
--- a/eval/gatekeeper/run-eval.py
+++ b/eval/gatekeeper/run-eval.py
@@ -1,0 +1,259 @@
+#!/usr/bin/env python3
+
+import os
+import sys
+
+from pathlib import Path
+from typing import Annotated
+
+import typer
+import yaml
+
+from utils import BlockStyleDumper
+
+from linux_mcp_server.gatekeeper import check_run_script
+
+
+app = typer.Typer()
+
+
+def run_test_file(test_case_file: str, label: str | None = None) -> tuple[list[dict], list[dict]]:
+    """Run a single test case file through the gatekeeper.
+
+    Returns (test_cases, results) where test_cases are the inputs and results
+    are the corresponding output dicts.
+    """
+    if label is None:
+        label = test_case_file
+    with open(test_case_file, "r") as f:
+        data = yaml.safe_load(f)
+
+    if isinstance(data, dict) and "cases" in data:
+        test_cases = data["cases"]
+    else:
+        test_cases = None
+
+    if not test_cases:
+        typer.echo(f"No test cases found in {test_case_file}")
+        raise typer.Exit(code=1)
+
+    results = []
+
+    with typer.progressbar(test_cases, label=f"Running {label}") as progress:
+        for test_case in progress:
+            description = test_case["description"]
+            script_type = test_case["script_type"]
+            script = test_case["script"]
+            readonly = test_case.get("readonly", False)
+
+            result = {
+                "description": description,
+                "script_type": script_type,
+                "script": script,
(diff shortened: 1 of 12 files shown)


eval/gatekeeper/run-eval.py
```diff
@@ -0,0 +1,271 @@
...
+    pass
+
+
+def _str_representer(dumper, data):
+    if "\n" in data:
+        return dumper.represent_scalar("tag:yaml.org,2002:str", data, style="|")
+    return dumper.represent_scalar("tag:yaml.org,2002:str", data)
+
+
+BlockStyleDumper.add_representer(str, _str_representer)
```
Jazzcort: This block also appears in `eval/gatekeeper/extract-from-goose.py`. We can move this block into `eval/gatekeeper/utils.py` or `eval/utils.py` if we have more common utility stuffs in the future. For now it's fine I think. 😁

eval/gatekeeper/run-eval.py
```diff
@@ -0,0 +1,271 @@
...
+        yaml_output = yaml.dump(output, indent=2, sort_keys=False, Dumper=BlockStyleDumper)
+
+        if output_file:
+            with open(output_file, "w") as f:
+                f.write(yaml_output)
+            typer.echo(f"Wrote {len(all_output_cases)} results to {output_file}")
+            print_summary_table(file_summaries, file=sys.stdout)
+        else:
+            sys.stdout.write(yaml_output)
+            print_summary_table(file_summaries, file=sys.stderr)
```
Jazzcort, line 239: We print the summary to stderr here is because we want the stdout to have a valid yaml format which is good for pipe and redirect usage right?

Jazzcort (approved):
Everything looks great to me! I tried both scripts and they works without any issues. Let me know if you want to have a table in the README file that displays the testing results of some common models or some models we suggest for users to pick from. 😁

eval/gatekeeper/run-eval.py
owtaylor: Yes, that's the idea. I don't know that the output format here is completely right - it's something that we'll need to tune some more especially when we hook things up to CI. But it's a start.

eval/gatekeeper/run-eval.py
owtaylor: I went ahead and added a `eval/gatekeeper/utils.py` for BlockStyleDumper... better not duplicated in any case.

PR #310, opened 2026-03-19
fix: Only parse CLI args when running linux-mcp-server

owtaylor:
Changed cli_parse_args to only enable CLI argument parsing when the script being run is linux-mcp-server itself. This prevents conflicts when other scripts (like eval scripts) import from linux_mcp_server and have their own CLI arguments. Settings can still be configured via environment variables in all cases.

---

I wanted to have some scripts for evaluation that imported files from linux_mcp_server; parsing the CLI args in this case is pretty annoying, so came up with this change (cherry-picked from #309). What do you think, @samdoran ?

### DIFF
--- a/src/linux_mcp_server/config.py
+++ b/src/linux_mcp_server/config.py
@@ -25,12 +25,11 @@ class Config(BaseSettings):
         env_prefix="LINUX_MCP_",
         env_ignore_empty=True,
         cli_hide_none_type=True,
-        # Only ignore errors for incorrect/extra parameters when testing
-        # https://github.com/pydantic/pydantic-settings/issues/391
-        cli_ignore_unknown_args=sys.argv[0].endswith("pytest"),
         cli_implicit_flags=True,
         cli_kebab_case=True,
-        cli_parse_args=True,
+        # Only parse CLI args when running linux-mcp-server itself, not when
+        # importing the module for other scripts (like eval scripts)
+        cli_parse_args=sys.argv[0].endswith("linux-mcp-server"),
     )
 
     # FIXME: When the next version of pydantic-settings is released, change this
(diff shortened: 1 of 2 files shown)


samdoran (approved):
This change looks really good. I like how you used `monkeypatch` for everything in the tests. I would not have not to use `setattr` on `sys.argv`

My only minor concern is that command line parsing doesn't happen with `python -m linux_mcp_server` as you pointed out. We can fix that later if it becomes an issue.

src/linux_mcp_server/config.py
```diff
@@ -25,12 +25,11 @@ class Config(BaseSettings):
...
         cli_hide_none_type=True,
-        # Only ignore errors for incorrect/extra parameters when testing
-        # https://github.com/pydantic/pydantic-settings/issues/391
-        cli_ignore_unknown_args=sys.argv[0].endswith("pytest"),
         cli_implicit_flags=True,
         cli_kebab_case=True,
-        cli_parse_args=True,
+        # Only parse CLI args when running linux-mcp-server itself, not when
+        # importing the module for other scripts (like eval scripts)
+        cli_parse_args=sys.argv[0].endswith("linux-mcp-server"),
```
samdoran, line 32: Since we required Python >= 3.10, we would check `sys.orig_argv` to make argument parsing working when the module is run directly.
```suggestion
        cli_parse_args=sys.argv[0].endswith("linux-mcp-server") or "linux_mcp_server" in sys.orig_argv,
```

samdoran: Not a blocker, just wanted to document it for the future.

owtaylor: Thanks for the further review! I think I'd rather commit it as is for now, and see what problems actually pop up; maybe there's some other clever way someone has of running the server.

PR #311, opened 2026-03-20
Make MCP server instructions depend on mcp-apps status

owtaylor:
This replaces part of #296 ... taking a hit in complexity to try and avoid leaking our implementation details to the model.

---

We refer to run_script_modify within the server instructions; this confuses some models, which trust the server instructions over the actual list of tools they have available.

To avoid this problem, dynamically modify the server instructions to refer to run_script_modify_interactive when applicable. This is just a bit hacky for FastMCP 3.x - we have to reach into the ServerSession object in our on_initialize middleware and change the instructions.

For FastMCP 2.x which doesn't make the session available, we have to use details of how CPython handles closures to dig out the session - it should work for now until we want to upgrade to FastMCP 3.

### DIFF
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -5,11 +5,18 @@ import sys
 
 from importlib import resources
 from pathlib import Path
+from types import CellType
+from types import CodeType
 
 from fastmcp import FastMCP
 from fastmcp.server.middleware import Middleware
 from fastmcp.server.middleware import MiddlewareContext
+from fastmcp.server.middleware.middleware import CallNext
+from mcp import ServerSession
 from mcp.types import BlobResourceContents
+from mcp.types import InitializeRequest
+from mcp.types import InitializeRequestParams
+from mcp.types import InitializeResult
 from mcp.types import ReadResourceRequest
 from mcp.types import ReadResourceResult
 from mcp.types import ServerResult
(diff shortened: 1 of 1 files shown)


src/linux_mcp_server/server.py
```diff
@@ -196,7 +203,65 @@ async def _read_resource_with_meta(req: ReadResourceRequest):
...
+            session: ServerSession | None = getattr(context.fastmcp_context, "_session", None)
+            if session is None:
+                # FastMCP 2.x - let's pull out the hacks! call_next is a closure within a method
+                # of fastmcp.server.low_level.MiddlewareServerSession. The "self" variable used
+                # in the closure is what we need. Assuming CPython, we can dig and and get it!
+                code: CodeType | None = getattr(call_next, "__code__", None)
+                closure: tuple[CellType, ...] | None = getattr(call_next, "__closure__", None)
+                if code and closure:
+                    # co_freevars gives us the names of the variables captured in __closure__
+                    closure_dict = dict(zip(code.co_freevars, [c.cell_contents for c in closure]))
```
Jazzcort, line 251: Wow, this is basically black magic. Today officially marks @owtaylor earning the title of Python Black Magician 😆

Jazzcort (approved):
Kind of a good place for me to use my mcp inspector 😁 LGTM!

### Goose Cli
Image

### Goose Desktop
Image

PR #312, opened 2026-03-20
Merge main into experiment/guard-command-execution (2026-03-20)

owtaylor:
This merges current main into the guarded-command-execution branch.

There are no conflicts other than uv.lock.

---

Review instructions

```sh
git fetch origin pull/312/head:pull-312-review

# Show 3-way diff for files that had changes in both branches
git show pull-312-review

# Alternate - show diff between automatic resolution and the final result
git show pull-312-review --diff-merges=remerge
```


subpop (approved):
Looks reasonable to me.

PR #313, opened 2026-03-21
fix(commands): limit journal_logs output to specified number of lines using head instead of --lines

umglurf:
Journalctl with `--lines` combined with other filters such as `--since` filters lines first, then applies filtering logic, ref https://github.com/systemd/systemd/pull/26669
This makes the journalctl tool not work as intended, as the agent is not able to query logs from a given time.
This PR replaces `--lines` with using head, making it possible to query logs from a given timestamp.


owtaylor:
Filed https://github.com/rhel-lightspeed/linux-mcp-server/issues/472 - closing the PR since we want to do something a bit different.

PR #328, opened 2026-04-01
test: migrate functional tests into monorepo and configure CI

alexxa:
- Moved functional tests, utilities, and FMF plans from external testing repository.
- Simplified .github/workflows/functional-tests.yml to test the current PR/branch directly without context variable workarounds.
- Added Red Hat copyright headers to all test files to comply with open-source policy guidelines.
- Updated main.fmf to install the server locally for testing.


tests/functional/general/test_list_tools.py
```diff
@@ -0,0 +1,40 @@
+# Copyright Red Hat
```
owtaylor: Sorry for not giving clearer guidance about this when you asked on Slack - I'm not really a fan of this approach (though it is one of the recommended approaches in our  "Copyright notices for open source projects" document.) The thing is, we want outside contributors to fix and improve tests along with everyone else, and we neither want to *require* people to add their names, or *forbid* people from adding their names.

My general preference would be to do what we do so far - do *not* add any per-file notices. The only thing that the Red Hat guidelines *require* is a toplevel LICENSE file - they basically say that adding per-file copyright/license notices just make your code look better to someone reviewing it and have no legal purpose.

But if you'd prefer to add something to these files, what I'd ask for the other approach from that document, and also to include the license, since as the document says "If you add a copyright notice without an associated license notice, users may reasonably assume that your code is not open source."

```
# Copyright Contributors to the linux-mcp-server project
# SPDX-License-Identifier: Apache-2.0
```

.github/workflows/functional-tests.yml
```diff
@@ -1,40 +1,89 @@
...
       TESTING_FARM_API_TOKEN: ${{ secrets.TESTING_FARM_API_TOKEN }}
 
     steps:
-      - name: Checkout
-        uses: actions/checkout@v6.0.2
       - name: Run TF request tests
         run: |
           echo "Starting functional tests..."
+
+          # Now that tests are in-repo, Testing Farm handles checking out the correct commit natively.
```
owtaylor: This comment wouldn't make sense to someone reading this in the future?

.github/workflows/functional-tests.yml
```diff
@@ -1,40 +1,89 @@
...
+
+          if [[ "${{ github.event_name }}" == "pull_request_target" ]]; then
+            GIT_URL="${{ github.event.pull_request.head.repo.clone_url }}"
+            GIT_REF="${{ github.event.pull_request.head.sha }}"
+          elif [[ "${{ github.event_name }}" == "issue_comment" ]]; then
+            # We would need a small script to fetch the PR head info here, similar to before.
+            # However, for the first pass, we focus on the core merge benefit.
+            # Simplified for now assuming push/PR-target usage.
+            GIT_URL="https://github.com/${{ github.repository }}"
+            GIT_REF="${{ github.sha }}"
```
owtaylor: Using github.sha (tip of default branch) doesn't make sense to me - if someone types /retest in a PR, they *don't* want to run the tests against the main branch and it will be super confusing if that happens.

You could just omit issue_comment triggers for the first pass, but I'd get an LLM to fix this properly and include the full version in this PR - it shouldn't be more than a few lines of code and they'll know how to do it.

plans/main.fmf
```diff
@@ -0,0 +1,80 @@
...
+  elif dnf info python3.11 &>/dev/null; then
+      dnf install -y python3.11 python3.11-pip
+      PYTHON_CMD=python3.11
+  else
+      dnf install -y python3-pip
+      PYTHON_CMD=python3
+  fi
+  # Install from the current directory (the repo under test)
+  $PYTHON_CMD -m pip install .
+  $PYTHON_CMD -m pip install -r tests/functional/requirements.txt
```
owtaylor, line 16: This is ok for now, but I don't want to have a file full of manually maintained versions going forward. We should add any additional dependencies for the functional tests to dependency-groups in pyproject.toml and do:

```sh
pip install '.[functional]'
```

*Or* just use uv with an isolated virtual environment. The point of these tests is testing compatibility of our code with the target system, not testing compatibility with the host system's python environment. Note that the Python discovery above *already* means that we frequently won't be testing the host system's Python environment. In particular, for Fedora 43, system python is python3.13.

owtaylor (changes requested):
Mostly looks very good.

* Were you able to test this in practice, or are we going to land this and do a follow-up to make it work? It's annoying from this perspective that github won't run the workflow until it a first version is merged into the default branch. (I'm OK with that.)
* Do you know why a plain `pytest` doesn't recurse in and try to run these tests? From the looks of the CI run it doesn't, but I don't know why it doesn't.

alexxa:
> Mostly looks very good.
> 
>     * Were you able to test this in practice, or are we going to land this and do a follow-up to make it work? It's annoying from this perspective that github won't run the workflow until it a first version is merged into the default branch. (I'm OK with that.)
> 
>     * Do you know why a plain `pytest` doesn't recurse in and try to run these tests? From the looks of the CI run it doesn't, but I don't know why it doesn't.

thanks for the review!

- The CI setup was verified locally using tmt and pytest within the target environment. The pull_request_target is only triggered once the file is merged to the main branch, since the change comes from a fork. Let's see how it goes and rework if any issues?
- Pytest is configured to ignore this directory via norecursedirs = ["tests/functional"] in pyproject.toml to prevent it from trying to run functional tests as unit tests. At first this PR was without this change and then unit tests just hanged.

tests/functional/general/test_list_tools.py
alexxa: Even though there is a LICENSE file at the root, it is standard practice and specifically noted in our policy to include the SPDX identifier and a generic copyright notice in all source files to ensure the license travels with the code if copied. Therefore I changed it to your suggested instead of removing it entirely, ok?

plans/main.fmf
alexxa: I understand the desire to centralize dependencies. However, having pinned versions dedicated to the functional tests ensures exact reproducibility in our tests without muddying the main project dependencies. That being said, migrating to uv with an isolated environment and managing these via dependency-groups in pyproject.toml is a great idea. Since this PR focuses on the initial migration, could we track this specific improvement in a follow-up issue?

.github/workflows/functional-tests.yml
alexxa: I updated the workflow file to properly fetch the correct PR ref. For issue_comment events, the workflow now uses a small inline python script that calls the GitHub API by using the injected GITHUB_TOKEN to dynamically pull the head.repo.clone_url and head.sha for the associated PR, avoiding the github.sha default branch mix-up.

.github/workflows/functional-tests.yml
alexxa: right, removed.

plans/main.fmf
owtaylor: Sounds good! Can I get you to file the issue?

.github/workflows/functional-tests.yml
```diff
@@ -1,40 +1,92 @@
...
+            export GITHUB_TOKEN="${{ secrets.GITHUB_TOKEN }}"
+
+            PR_DATA=$(python3 -c "
+import urllib.request, json, os
+req = urllib.request.Request(os.environ['PR_API_URL'], headers={'Authorization': 'Bearer ' + os.environ['GITHUB_TOKEN']})
+print(json.dumps(json.loads(urllib.request.urlopen(req).read().decode('utf-8'))))
+")
+
+            GIT_URL=$(echo "$PR_DATA" | python3 -c "import sys, json; print(json.load(sys.stdin)['head']['repo']['clone_url'])")
+            GIT_REF=$(echo "$PR_DATA" | python3 -c "import sys, json; print(json.load(sys.stdin)['head']['sha'])")
```
owtaylor, line 72: That looks functional, but a bit clumsy. I asked Gemini and it suggested that the easiest thing is that the standard github have the gh cli preinstalled, and we could do something like:
```yaml
    - name: Fetch PR details
        env:
          GH_TOKEN: ${{ secrets.GITHUB_TOKEN }}
        run: |
          # Define the API endpoint relative to the base URL
          ENDPOINT="repos/${{ github.repository }}/pulls/${{ github.event.issue.number }}"
          
          # Fetch and parse directly using gh api and jq
          GIT_URL=$(gh api $ENDPOINT --jq '.head.repo.clone_url')
          GIT_REF=$(gh api $ENDPOINT --jq '.head.sha')
          
          # Optional: Export them as environment variables for subsequent steps
          echo "GIT_URL=$GIT_URL" >> $GITHUB_ENV
          echo "GIT_REF=$GIT_REF" >> $GITHUB_ENV
```
(I quote this in full to show the `env:` usage, needs a bit of adaptation to our case)

owtaylor (changes requested):
All sounds/looks good. I think we might be able to do a bit better on the issue_comment handling- otherwise it looks ready to land to me.

.github/workflows/functional-tests.yml
alexxa: it indeed looks cleaner. However, the tf-functional-tests job actually runs inside the quay.io/testing-farm/cli and not directly on the ubuntu-latest runner. And that container image, unfortunately, doesn't have gh, jq, or even curl installed. It does have python3, though, which is why I went with the inline Python script to ensure the API call succeeds without needing to modify the container or install extra dependencies during the run.

plans/main.fmf
alexxa: Filed #367

.github/workflows/functional-tests.yml
owtaylor: Good point! - I think there might be alternate ways to structure this that would work - like have a single step that `uses: docker://quay.io/testing-farm/cli` instead of having `container:` at the toplevel of job, but if I care, I can submit a PR for that once this lands (and in that case, there will be working CI for that)!

owtaylor (approved):

PR #333, opened 2026-04-02
Convert reset_vendor fixture to save sys.modules and sys.path

subpop:
The reset_vendor fixture in tests/_vendor/test_vendor.py removed linux_mcp_server from sys.modules so vendor tests could force a clean import. After that, any later import linux_mcp_server created a new package object. That object only gets what __init__.py defines (CONFIG, __version__, etc.); it does not automatically attach commands, connection, or __main__ as attributes.

This seems to cause problems with Python 3.10's unit tests. I'm not positive this is consistently reproducible, but I've seen it a few times locally, and every time I see errors like this:

```
E           AttributeError: module 'linux_mcp_server' has no attribute 'connection'
```

The fix has been to apply this patch to the `reset_vendor` fixture. I have integrated these same changes to the `experimental/guarded-command-execution branch` and the `convert-services-structured-output` to fix failing Python 3.10 unit tests.

### DIFF
--- a/tests/_vendor/test_vendor.py
+++ b/tests/_vendor/test_vendor.py
@@ -16,11 +16,10 @@ def reset_vendor():
     import linux_mcp_server
 
     vendor_path = str(Path(linux_mcp_server.__file__).parent / "_vendor")
-    removed_modules = [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
     saved_path = list(sys.path)
 
     [sys.path.remove(path) for path in sys.path if path == vendor_path]
-    [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
+    removed_modules = [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
 
     yield
 
--- a/tests/_vendor/test_vendor.py
+++ b/tests/_vendor/test_vendor.py
@@ -16,8 +16,18 @@ def reset_vendor():
     import linux_mcp_server
 
     vendor_path = str(Path(linux_mcp_server.__file__).parent / "_vendor")
+    removed_modules = [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
+    saved_path = list(sys.path)
+
     [sys.path.remove(path) for path in sys.path if path == vendor_path]
     [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
+
+    yield
+
+    sys.path[:] = saved_path
+    for module in removed_modules:
+        if module:
+            sys.modules[module.__name__] = module
 
 
 def test_package_masking():


owtaylor:
I have no idea why
```
E           AttributeError: module 'linux_mcp_server' has no attribute 'connection'' 
```
The module linux_mcp_server will *never* have that attribute. I'm guessing it has something to do with the monkey patching infrastructure we are using and what it does if it gets confused about sys.modules.

That being said, it's not really proper to directly munge with `sys.modules` and it can have unexpected side-effects. What you are doing here should work OK considering how simple `linux_mcp_server.__init__.py` is, but my mind would run to something more like:

https://github.com/owtaylor/linux-mcp-server/commit/1a483e75ec40081addb467ce88b0b94a

`importlib.reload()` does basically precisely what we need here - reruns the contents of the file while leaving `sys.modules()` untouched.

owtaylor (dismissed):
Marking as "Request changes", as per earlier comment. Let me know if that doesn't work.

samdoran:
`pytest-randomly`, a plugin we have installed, shuffles the order of tests. This failure only happens on Python 3.10 when `test_vendor` runs before tests that patch `linux_mcp_server.commands`. Here is a simple reproducer.
```
> pytest --no-cov --randomly-seed=2279128600 -k 'test_list_block_devices_remote_execution or test_vendored'
======================================== test session starts ========================================
platform darwin -- Python 3.10.18, pytest-9.0.2, pluggy-1.6.0 -- /Users/sdoran/.pyenv/versions/3.10.18/envs/linux-mcp-server-3.10/bin/python3
cachedir: .pytest_cache
Using --randomly-seed=2279128600
rootdir: /Users/sdoran/Developer/linux-mcp-server
configfile: pyproject.toml
testpaths: tests
plugins: anyio-4.12.1, mock-3.15.1, cov-7.1.0, randomly-4.0.1, asyncio-1.3.0
asyncio: mode=auto, debug=False, asyncio_default_fixture_loop_scope=None, asyncio_default_test_loop_scope=function
collected 427 items / 424 deselected / 3 selected

tests/_vendor/test_vendor.py::test_vendored_warning PASSED                                    [ 33%]
tests/_vendor/test_vendor.py::test_vendored PASSED                                            [ 66%]
tests/tools/storage/test_list_block_devices.py::test_list_block_devices_remote_execution ERROR [100%]
...
====================================== short test summary info ======================================
ERROR tests/tools/storage/test_list_block_devices.py::test_list_block_devices_remote_execution - AttributeError: module 'linux_mcp_server' has no attribute 'commands'
```

This only affects Python 3.10 because Python 3.11 and later switched to using [pkgutil.resolve_name](https://github.com/python/cpython/commit/ab7fcc8fbdc11091370deeb000a787fb02f9b13d) which does a [much better job](https://bugs.python.org/issue12915) of resolving dotted module names.

tests/_vendor/test_vendor.py
```diff
@@ -16,9 +16,22 @@ def reset_vendor():
     import linux_mcp_server
 
     vendor_path = str(Path(linux_mcp_server.__file__).parent / "_vendor")
+    saved_modules = {
+        k: v for k, v in sys.modules.items() if k == "linux_mcp_server" or k.startswith("linux_mcp_server.")
+    }
+    saved_path = list(sys.path)
```
samdoran, line 19: You can also copy a list using slice notation.
```suggestion
    saved_path = sys.path[:]
```

tests/_vendor/test_vendor.py
```diff
@@ -16,9 +16,22 @@ def reset_vendor():
     import linux_mcp_server
 
     vendor_path = str(Path(linux_mcp_server.__file__).parent / "_vendor")
+    saved_modules = {
```
samdoran: The modules are already in a list on line 20. We can just save that instead of throwing it away.

```
removed_modules = [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
```

tests/_vendor/test_vendor.py
```diff
@@ -16,9 +16,22 @@ def reset_vendor():
...
     [sys.path.remove(path) for path in sys.path if path == vendor_path]
     [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
 
+    yield
+
+    sys.path[:] = saved_path
+    for key in list(sys.modules.keys()):
+        if key == "linux_mcp_server" or key.startswith("linux_mcp_server."):
+            sys.modules.pop(key, None)
+    sys.modules.update(saved_modules)
```
samdoran:
```suggestion
    for module in removed_modules:
        if module:
            sys.modules[module.__name__] = module
```

samdoran (dismissed):

tests/_vendor/test_vendor.py
```diff
@@ -13,11 +13,15 @@ def __init__(self, name):
 
 @pytest.fixture
 def reset_vendor():
-    import linux_mcp_server
+    removed_modules = [sys.modules.pop(package, None) for package in ["linux_mcp_server._vendor", "linux_mcp_server"]]
+    saved_path = sys.path[:]
 
-    vendor_path = str(Path(linux_mcp_server.__file__).parent / "_vendor")
-    [sys.path.remove(path) for path in sys.path if path == vendor_path]
```
samdoran, line 19: Looks like this line was inadvertently removed.

samdoran (approved):

subpop:
Thanks Sam!

PR #339, opened 2026-04-05
fix(ssh): add timeout to local command execution

xingyug:
## What
Add timeout enforcement to local command execution in `_execute_local()`, matching the existing timeout behavior of remote SSH execution.

## Why
`execute_remote()` enforces `CONFIG.command_timeout` (default 30s) via asyncssh. `_execute_local()` calls `proc.communicate()` with no timeout — a local command that hangs (e.g., `du` on a stuck NFS mount) blocks the coroutine indefinitely, leaking the subprocess.

## How
Wrap `proc.communicate()` with `asyncio.wait_for(..., timeout=CONFIG.command_timeout)`. On timeout: kill the subprocess, reap the zombie via `proc.wait()`, raise `ConnectionError` with message format matching remote execution. Added `except ConnectionError: raise` before the existing `except Exception` to prevent the timeout from being swallowed.

## Checklist
- [x] Tests added and passing (`make verify` — 430 passed)
- [x] Code follows style guidelines (ruff + pyright clean)
- [x] Commit messages follow conventional format
- [x] No security vulnerabilities introduced
- [x] All operations remain read-only
- [x] Error handling is appropriate

### DIFF
--- a/src/linux_mcp_server/config.py
+++ b/src/linux_mcp_server/config.py
@@ -78,8 +78,8 @@ class Config(BaseSettings):
     # Gatekeeper model (required for run_script tools)
     gatekeeper_model: str | None = None
 
-    # Command execution timeout (applies to remote SSH commands)
-    command_timeout: int = 30  # Timeout in seconds; prevents hung SSH operations
+    # Command execution timeout (applies to both local and remote commands)
+    command_timeout: int = 30  # Timeout in seconds; prevents hung commands
 
     # Indicate mcp-app compatibility
     use_mcp_apps: bool | None = None
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -441,10 +441,26 @@ async def _execute_local(
         bin = get_bin_path(bin)
 
     full_command = [bin, *command[1:]]
+    timeout = CONFIG.command_timeout
 
     try:
         proc = await asyncio.create_subprocess_exec(*full_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
-        stdout_bytes, stderr_bytes = await proc.communicate()
+        try:
+            stdout_bytes, stderr_bytes = await asyncio.wait_for(proc.communicate(), timeout=timeout)
+        except asyncio.TimeoutError:
+            proc.kill()
+            await proc.wait()
+            duration = time.time() - start_time
+            logger.error(
+                f"Command timed out after {timeout}s",
+                extra={
+                    "event": Event.LOCAL_EXEC_ERROR,
+                    "command": cmd_str,
+                    "duration": f"{duration:.3f}s",
+                    "error": "timeout",
+                },
+            )
+            raise ConnectionError(f"Command timed out after {timeout}s on localhost: {cmd_str}") from None
 
         return_code = proc.returncode if proc.returncode is not None else 0
         stdout = stdout_bytes if encoding is None else stdout_bytes.decode(encoding, errors="replace")
@@ -452,11 +468,12 @@ async def _execute_local(
 
         duration = time.time() - start_time
 
-        # DEBUG level: Log local command execution with timing
         logger.debug(f"LOCAL_EXEC completed: {cmd_str} | exit_code={return_code} | duration={duration:.3f}s")
 
         return return_code, stdout, stderr
 
+    except ConnectionError:
+        raise
     except Exception as e:
         duration = time.time() - start_time
         logger.error(
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -429,10 +429,15 @@ async def _execute_local(
 
     Args:
         command: Command and arguments to execute
+        encoding: Character encoding for stdout/stderr. Defaults to "utf-8".
+            Set to None to receive raw bytes.
 
     Returns:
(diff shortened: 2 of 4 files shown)


src/linux_mcp_server/connection/ssh.py
```diff
@@ -441,22 +441,39 @@ async def _execute_local(
...
+            logger.error(
+                f"Command timed out after {timeout}s",
+                extra={
+                    "event": Event.LOCAL_EXEC_ERROR,
+                    "command": cmd_str,
+                    "duration": f"{duration:.3f}s",
+                    "error": "timeout",
+                },
+            )
+            raise ConnectionError(f"Command timed out after {timeout}s on localhost: {cmd_str}") from None
```
subpop: This isn't strictly a `ConnectionError`. I think it makes more sense to raise a `TimeoutError` here. Can you switch the exception type to `TimeoutError`?

src/linux_mcp_server/connection/ssh.py
subpop, line 437: The docstring for `_execute_local` should note that it raises an exception.

src/linux_mcp_server/connection/ssh.py
```diff
@@ -441,22 +441,39 @@ async def _execute_local(
         bin = get_bin_path(bin)
 
     full_command = [bin, *command[1:]]
+    timeout = CONFIG.command_timeout
```
subpop, line 449: Since we're now using `command_timeout` for both SSH and local timeouts, we need to update the documentation in `config.py` and in the `docs/` that the timeout is used for both cases.

subpop (changes requested):
Thanks for the contribution. I'd like to swap the exception raised from a `ConnectionError` to a `TimeoutError`, as that's more accurately describing what's happening in this local command timeout function.

xingyug:
Thanks for the review — I’ve addressed the requested changes and pushed the updates.

src/linux_mcp_server/config.py
```diff
@@ -72,15 +72,15 @@ class Config(BaseSettings):
...
+    
     # What tools are available
     toolset: Toolset = Toolset.FIXED
 
     # Gatekeeper model (required for run_script tools)
     gatekeeper_model: str | None = None
 
-    # Command execution timeout (applies to remote SSH commands)
-    command_timeout: int = 30  # Timeout in seconds; prevents hung SSH operations
-
```
subpop, line 83: Why did this move up? I'd prefer to keep the diff as minimal as possible and just highlight the change in comment. Was there a deliberate reason you moved it?

src/linux_mcp_server/connection/ssh.py
```diff
@@ -429,10 +429,15 @@ async def _execute_local(
 
     Args:
         command: Command and arguments to execute
+        encoding: Character encoding for stdout/stderr. Defaults to "utf-8".
+            Set to None to receive raw bytes.
```
subpop, line 433: Thanks for catching this!

docs/clients.md
```diff
@@ -622,7 +622,7 @@ See [Guarded Command Execution](guarded-command-execution.md)
 | `LINUX_MCP_SSH_KEY_PATH` | Path to SSH private key | `~/.ssh/id_ed25519` |
 | `LINUX_MCP_KEY_PASSPHRASE` | Passphrase for encrypted SSH key | (set value in env) |
 | `LINUX_MCP_SEARCH_FOR_SSH_KEY` | Auto-discover keys in `~/.ssh` | `yes` |
-| `LINUX_MCP_COMMAND_TIMEOUT` | SSH command timeout in seconds (default: 30) | `60` |
+| `LINUX_MCP_COMMAND_TIMEOUT` | Command timeout in seconds, local and remote (default: 30) | `60` |
```
subpop:
```suggestion
| `LINUX_MCP_COMMAND_TIMEOUT` | Local and remote command timeout in seconds (default: 30) | `60` |
```

docs/usage.md
```diff
@@ -51,7 +51,7 @@ Options may be set using environment variables or command line options. Environm
 | `--search-for-ssh-key` | bool | False | Auto-discover SSH keys in `~/.ssh` |
 | `--verify-host-keys` | bool | True | Verify remote host identity via known_hosts |
 | `--known-hosts-path` | path | null | Path to known_hosts file |
-| `--command-timeout` | integer | 30 | SSH command timeout in seconds |
+| `--command-timeout` | integer | 30 | Command timeout in seconds (local and remote) |
```
subpop:
```suggestion
| `--command-timeout` | integer | 30 | Local and remote command timeout in seconds |
```

subpop (changes requested):
Sorry for the nitpicking comments. I'm particular about the docs; they get rendered into the website as part of the user-facing documentation, so consistency is critical.

src/linux_mcp_server/config.py
xingyug: The move was unintentional — it was a merge conflict resolution mistake. I have fixed it to keep command_timeout in its original position and just update the comment.

xingyug:
> Sorry for the nitpicking comments. I'm particular about the docs; they get rendered into the website as part of the user-facing documentation, so consistency is critical.

Appreciate the thorough review — consistent docs absolutely matter. Happy to address any further feedback.

subpop:
I rebased the PR and moved the documentation changes to the new file location `config-reference.md`.

subpop (approved):

PR #347, opened 2026-04-07
feat(run_script): add needs_confirmation flow

owtaylor:
- Add always_confirm_scripts config option to force confirmation for all scripts
- Make needs_confirmation a computed property on ScriptDetails based on readonly flag and always_confirm_scripts config
- Update validate_script to return needs_confirmation in structured output
- Add run_script_with_confirmation tool for scripts requiring user approval
- Update server instructions to document the new workflow


subpop (approved):
LGTM

PR #349, opened 2026-04-07
feat(run_script): simplify tool parameters and merge toolsets

owtaylor:
- Remove run_script_readonly and run_script_modify (deprecated)
- Rename run_script_modify_interactive to run_script_interactive
- run_script only takes token parameter
- run_script_interactive and run_script_with_confirmation take all parameters  (needed for UI display and approval visibility)
- On parameter mismatch, run_script_interactive creates new script store entry
- Remove VALIDATE_SCRIPT toolset enum, keep only RUN_SCRIPT
- Update McpAppToolParamsSchema with readonly field
- Update server instructions and docs/usage.md

### DIFF
--- a/mcp-app/src/types.ts
+++ b/mcp-app/src/types.ts
@@ -11,12 +11,16 @@ export const McpAppToolParamsSchema = z
     script: z.string(),
     script_type: scriptTypeSchema,
     description: z.string(),
+    readonly: z.boolean(),
+    token: z.string(),
     host: z.string().optional().nullable(),
   })
   .transform((data) => ({
     script: data.script,
     scriptType: data.script_type,
     description: data.description,
+    readonly: data.readonly,
+    token: data.token,
     host: data.host,
   }));
 
--- a/docs/usage.md
+++ b/docs/usage.md
@@ -242,7 +242,7 @@ Lists immediate subdirectories under a specified path with flexible sorting opti
 
 #### `run_script` (experimental)
 Runs a script on a system. `validate_script` must be called first to check if the script matches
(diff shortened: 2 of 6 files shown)


mcp-app/src/types.ts
```diff
@@ -11,12 +11,14 @@ export const McpAppToolParamsSchema = z
     script: z.string(),
     script_type: scriptTypeSchema,
     description: z.string(),
+    readonly: z.boolean(),
```
Jazzcort: I think `token` is also needed here right? The fields here should exactly match the tool call params of `run_script_interactive`

Jazzcort, line 14: I first thought we should pass exactly same params here like what we pass to `run_script_interactive`, otherwise, the validation would fail. But, I was wrong, `zod` actually strip what is not listed inside `z.object`. So, it's fine not passing `token` here and we always don't need it since we grab token(id) from the tool call result. However, we don't use `readonly` anywhere in our app either. So, do we still need to pass it here? 🤔

Jazzcort (approved):
LGTM! I tested the mcp-apps and it works well like it used to! `run_script_with_confirmation` works pretty good too when I configure it to be "Always ask" The human approval flow feels same as mcp-apps. 😁

mcp-app/src/types.ts
owtaylor: Hmm, completely true. It's weird to add readonly when the token was also added. I think the cleanest is to add token - to match the parameters, rather than what we use. I'll add token in here.

PR #355, opened 2026-04-08
feat: add authentication and authorization support 

panyamkeerthana:
Adds authentication support including:
- Google OAuth and GitHub OAuth providers
- JWT/OIDC authentication (Keycloak)
- Token introspection (Keycloak)
- Auth logging middleware to capture user info

Continuing work from #322

### DIFF
--- a/src/linux_mcp_server/auth_policy.py
+++ b/src/linux_mcp_server/auth_policy.py
@@ -1,0 +1,203 @@
+import fnmatch
+import logging
+
+from enum import Enum
+from functools import cache
+from pathlib import Path
+from typing import Any
+
+import yaml
+
+from fastmcp.tools import Tool
+from pydantic import BaseModel
+from pydantic import model_validator
+
+from linux_mcp_server.config import CONFIG
+from linux_mcp_server.toolset import get_toolset
+
+
+logger = logging.getLogger("linux-mcp-server")
+
+
+class PolicyAction(str, Enum):
+    DENY = "deny"
+    SSH_KEY = "ssh_key"
+    SSH_DEFAULT = "ssh_default"
+    LOCAL = "local"
+
+
+class SSHKeyConfig(BaseModel):
+    path: str
+    user: str
+
+
+class PolicyRule(BaseModel):
+    host: str
+    tools: list[str]
+    claims: dict[str, Any] = {}
+    action: PolicyAction
+    ssh_key: SSHKeyConfig | None = None
+    all_users: bool = False
+
+    # Validate that all_users and claims are mutually exclusive
+    @model_validator(mode="after")
+    def validate_all_users_and_claims(self):
+        if self.all_users and self.claims:
+            raise ValueError("Rule with 'all_users: true' must not have claims specified")
+        if not self.all_users and not self.claims:
+            raise ValueError("Rule with 'all_users: false' must have non-empty claims")
+        return self
+
+    # Validate that host and action are semantically compatible
+    @model_validator(mode="after")
+    def validate_host_action(self):
+        # localhost should have LOCAL or DENY action
+        if self.host == "localhost":
+            if self.action not in [PolicyAction.LOCAL, PolicyAction.DENY]:
+                raise ValueError(
+                    f"Rule with host: 'localhost' must use action 'local' or 'deny' and not '{self.action.value}'"
+                )
+        # other hosts should have SSH_* or DENY action
+        else:
+            if self.action == PolicyAction.LOCAL:
+                raise ValueError(f"Rule with host: '{self.host}' cannot be use action 'local'")
+        return self
+
+    # Validate that SSH_KEY action has ssh_key configuration
+    @model_validator(mode="after")
+    def validate_ssh_key_config(self):
+        if self.action == PolicyAction.SSH_KEY and self.ssh_key is None:
+            raise ValueError("Rule with action 'ssh_key' must have ssh_key(path and user) configured")
+        return self
+
+    def matches_host(self, target_host: str | None) -> bool:
+        if target_host is None:
+            # 'host: *' should not allow local execution
+            return self.host == "localhost"
+        return fnmatch.fnmatch(target_host, self.host)
+
+    # Check if the rule matches the policy tool name
+    # Supports exact tool name: "run_script_readonly", prefixes: "@fixed" and wildcard: "*"
+    def matches_tool(self, tool_name: str, tool_tags: set[str]) -> bool:
+        for allowed_tool in self.tools:
+            if allowed_tool == "*":
+                return True
+            elif allowed_tool.startswith("@"):
+                # Toolset prefix check if tool belongs to this toolset
+                toolset_name = allowed_tool[1:]  # Remove @ prefix
+                toolset = get_toolset(toolset_name)
+                if toolset is None:
+                    logger.warning(f"Unknown toolset: {toolset_name}")
+                    return False
+
+                if toolset.includes_tool(tool_tags):
+                    return True
+            elif allowed_tool == tool_name:
+                return True
+        return False
+
+    # Check if the tokens claims satisfy the policy rules claim requirements
+    def matches_claims(self, token_claims: dict[str, Any]) -> bool:
+
+        for claim_key, claim_value in self.claims.items():
+            # Key must be present
+            if claim_key not in token_claims:
+                return False
+
+            token_value = token_claims[claim_key]
+
+            # If the value in the token is string, the values must match exactly
+            if isinstance(token_value, str):
+                if token_value != claim_value:
+                    return False
+            # If config value is string and token value is list, must be in list
+            elif isinstance(claim_value, str) and isinstance(token_value, list):
+                if claim_value not in token_value:
+                    return False
+            # For other types (bool, int, etc), must match exactly
+            else:
+                if token_value != claim_value:
+                    return False
+
+        return True
+
+    def matches(
+        self,
+        tool_name: str,
+        tool_tags: set[str],
+        target_host: str | None,
+        token_claims: dict[str, Any],
+    ) -> bool:
+        # Check if this rule matches the given context.
+        return (
+            self.matches_host(target_host)
+            and self.matches_tool(tool_name, tool_tags)
+            and self.matches_claims(token_claims)
+        )
+
+
+class AuthPolicy(BaseModel):
+    rules: list[PolicyRule]
+
+    @classmethod
+    def from_yaml(cls, yaml_path: Path) -> "AuthPolicy":
+        """Load policy from YAML file using Pydantic validation."""
+        try:
+            with open(yaml_path, "r") as f:
+                data = yaml.safe_load(f)
+
+            if not data:
+                logger.warning(f"Empty policy file {yaml_path}")
+                return cls(rules=[])
+
+            return cls.model_validate(data)
+
+        except Exception as e:
+            logger.error(f"Failed to load auth policy from {yaml_path}: {e}")
+            raise RuntimeError(f"Failed to load auth policy from {yaml_path}: {e}") from e
+
+    # Evaluate the policy for a given context
+    # returns (action, ssh_key_config, all_users) if no rules match, returns (DENY, None, False)
+    def evaluate(
+        self,
+        tool_name: str,
+        tool_tags: set[str],
+        target_host: str | None,
+        token_claims: dict[str, Any],
+    ) -> tuple[PolicyAction, SSHKeyConfig | None, bool]:
+
+        for rule in self.rules:
+            if rule.matches(tool_name, tool_tags, target_host, token_claims):
+                logger.debug(f"Policy match: tool={tool_name}, host={target_host}, action={rule.action.value}")
+                return rule.action, rule.ssh_key, rule.all_users
+
+        logger.warning(f"No policy rule matched: tool={tool_name}, host={target_host}, claims={token_claims}")
+        return PolicyAction.DENY, None, False
+
+
+# Get the current policy loading it if necessary
+@cache
+def get_policy() -> AuthPolicy:
+
+    if CONFIG.policy_path is None:
+        logger.info("No auth policy path configured, all requests will be denied")
+        return AuthPolicy(rules=[])
+
+    if not CONFIG.policy_path.exists():
+        logger.error(f"Auth policy file not found: {CONFIG.policy_path}")
+        return AuthPolicy(rules=[])
+
+    return AuthPolicy.from_yaml(CONFIG.policy_path)
+
+
+# Wrapper that loads policy and calls AuthPolicy.evaluate()
+def evaluate_policy(
+    tool: Tool,
+    target_host: str | None,
+    token_claims: dict[str, Any],
+) -> tuple[PolicyAction, SSHKeyConfig | None]:
+    policy = get_policy()
+    tool_name = tool.name
+    tool_tags = tool.tags if tool.tags else set()
+    action, ssh_key_config, _ = policy.evaluate(tool_name, tool_tags, target_host, token_claims)
+    return action, ssh_key_config
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -9,6 +9,7 @@ from types import CellType
 from types import CodeType
 
 from fastmcp import FastMCP
+from fastmcp.server.dependencies import get_access_token
 from fastmcp.server.middleware import Middleware
 from fastmcp.server.middleware import MiddlewareContext
 from fastmcp.server.middleware.middleware import CallNext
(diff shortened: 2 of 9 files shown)


src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,210 @@
...
+
+
+def get_policy() -> AuthPolicy | None:
+    # Get the current policy, loading it if necessary.
+    global _policy
+
+    if _policy is None:
+        return load_policy()
+
+    return _policy
```
owtaylor:
```suggestion
@cache
def get_policy() -> AuthPolicy | None:
    if CONFIG.policy_path is None:
        logger.info("No auth policy path configured, all requests will be denied")
        return None

    if not CONFIG.policy_path.exists():
        logger.error(f"Auth policy file not found: {CONFIG.policy_path}")
        return None

    return AuthPolicy.from_yaml(CONFIG.policy_path)
```

This could be simplified.

src/linux_mcp_server/server.py
```diff
@@ -240,6 +260,95 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+        tool_name = context.message.name
+        tool_args = context.message.arguments or {}
+        target_host = tool_args.get("host")
+
+        # Log authorization attempt
+        log_level(f"Tool call: {tool_name}, Host: {target_host}, User: {email}")
+
+        # If auth policy is configured evaluate it
+        if CONFIG.policy_path is not None:
+            action, ssh_key_config = evaluate_policy(tool_name, target_host, claims)
```
owtaylor: Hmm, the tool registry adds some conceptual complexity here, especially as future versions of fastmcp add more and more complexity to the tool lookup process (e.g., fastmcp adds a feature that tools can be filtered differently per session).

Maybe we should just go with:

```suggestion
            tool = context.fastmcp.get_tool(tool_name)
            action, ssh_key_config = evaluate_policy(tool, target_host, claims)
```

While this is horrifyingly less efficient with fastmcp 2 than just looking up tags in an array, it should be plenty fast enough.

src/linux_mcp_server/server.py
```diff
@@ -240,6 +260,95 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+
+            # Log the authorized action
+            log_level(
+                f"Authorized: tool={tool_name}, host={target_host or 'local'}, action={action.value}, user={email}"
+            )
+
+            # LOCAL and SSH_DEFAULT already handled by execute_command()
+            # TODO: Pass ssh_key_config.path to connection manager for SSH_KEY action
+            if action == PolicyAction.SSH_KEY and ssh_key_config:
+                logger.debug(f"SSH key override: {ssh_key_config.path}")
```
owtaylor: * `action == POLICY.SSH_KEY` and `not ssh_key_config` should be an error. 
* I'd raise for now rather than falling back to `SSH_DEFAULT`

src/linux_mcp_server/server.py
```diff
@@ -240,6 +260,95 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
     return MCP_APP_MIME_TYPE in mime_types
 
 
+# Middleware to enforce authorization policy
+class AuthorizationMiddleware(Middleware):
+    async def on_call_tool(self, context: MiddlewareContext, call_next):
+        # Skip authorization if auth is not configured
+        if CONFIG.auth is None:
+            return await call_next(context)
```
owtaylor, line 278: We want to make it *hard* to bypass auth for http. I'd restrict to this `CONFIG.transport == Transport.stdio`

For http, the way I think it could work is that we add `allow_unauthorized` to PolicyRule, so if you want to use http unauthorized, you need a policy like:

```yaml
rules: 
    # We check authorization in our MCP gateway, and restrict access to our server to
    # authorized users at the http transport level
    - allow_unauthorized: true
      action: ssh_default
```

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,210 @@
...
+        from linux_mcp_server.tool_registry import get_tool_tags
+
+        tags = get_tool_tags(tool_name)
+
+        if toolset == "fixed":
+            return "run_script" not in tags
+        elif toolset == "run_script":
+            return "run_script" in tags
+        elif toolset == "both":
+            return True
```
owtaylor: Don't like this being inline here and separated from the existing logic, seems error prone. Can you add a little toolset.py with:

```py
class Toolset:
    name: str
    includes_tags: set
    exclude_tags: set

    def includes_tool(Tool: tool):
        ...


def get_toolset(str: name) -> Toolset:
    ....
```

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,210 @@
...
+        for allowed_tool in self.tools:
+            if allowed_tool == "*":
+                return True
+            elif allowed_tool.startswith("@"):
+                # Toolset prefix check if tool belongs to this toolset
+                toolset = allowed_tool[1:]  # Remove @ prefix
+                if self._tool_in_toolset(tool_name, toolset):
+                    return True
+            elif allowed_tool == tool_name:
+                return True
```
owtaylor, line 96: I'd like to extend this in the future to allow "-tool" for excluding tools, but fine as is for now.

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,210 @@
...
+
+class PolicyAction(str, Enum):
+    DENY = "deny"
+    SSH_KEY = "ssh_key"
+    SSH_DEFAULT = "ssh_default"
+    LOCAL = "local"
+
+
+class SSHKeyConfig(BaseModel):
+    path: str
```
owtaylor, line 30: Let's add user here as a *mandatory* key. I don't think the case where the MCP server is running with the same username as needed on the target machine is common, so making it mandatory will avoid an annoying config error.

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,210 @@
...
+                    ssh_key=ssh_key,
+                )
+                rules.append(rule)
+
+            return cls(rules=rules)
+
+        except Exception as e:
+            logger.error(f"Failed to load auth policy from {yaml_path}: {e}")
+            # Return empty policy (deny all)
+            return cls(rules=[])
```
owtaylor: I think this can just be fatal. Running a server with everything denied because auth_policy.json didn't parse doesn't seem useful.

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,210 @@
...
+
+class AuthPolicy(BaseModel):
+    rules: list[PolicyRule]
+
+    @classmethod
+    def from_yaml(cls, yaml_path: Path) -> "AuthPolicy":
+        # Load policy from YAML file.
+        try:
+            with open(yaml_path, "r") as f:
+                data = yaml.safe_load(f)
```
owtaylor, line 147: Since AuthPolicy is a pydantic BaseModel, why not just:
   
```suggestion
                data = yaml.safe_load(f)
                return AuthPolicy.model_validate(data)
```

owtaylor (changes requested):
This is looking good, some high level suggestions (github got wonky, so I might need to add a couple of more suggestions afterwards).

src/linux_mcp_server/auth_policy.py
owtaylor: Also, I'd enforce the interpretation of no-policy == all denied by returning an `AuthPolicy(rules=[])` rather than `None`. Don't give the caller the chance to screw it up.

src/linux_mcp_server/server.py
panyamkeerthana: Oh this is good catch,  sure makes security more conscious and explicit!

src/linux_mcp_server/server.py
panyamkeerthana: This is sm cleaner, I knew the tool registry felt clunky but  the alternative (storing mcp.instance globally) would force the entire policy chain to be async. 
Totally didn't click that we're already in async middleware context, so we can just await get_tool()  and pass tags down to the policy functions🤦🏽‍♀️

src/linux_mcp_server/auth_policy.py
panyamkeerthana: Policy files aren't supposed to change at runtime so this is far clearer! 😄

src/linux_mcp_server/auth_policy.py
panyamkeerthana: Changing it to raise RuntimeError on parse failure

src/linux_mcp_server/auth_policy.py
panyamkeerthana: Oh nicee, didn't realize Pydantic handled all the nested parsing automatically

src/linux_mcp_server/auth_policy.py
panyamkeerthana: sounds good or could also add a exclude_tools field to PolicyRule

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,169 @@
...
+    action: PolicyAction
+    ssh_key: SSHKeyConfig | None = None
+    allow_unauthorized: bool = False
+
+    def matches_host(self, target_host: str | None) -> bool:
+        # Wildcards can match both local (None) and remote hosts
+        # The action field determines whether execution is local or SSH
+        if target_host is None:
+            target_host = "localhost"
+        return fnmatch.fnmatch(target_host, self.host)
```
owtaylor, line 77:
```suggestion
        if target_host is None:
            # 'host: *' should not allow local execution
            return self.host == "localhost"
        return fnmatch.fnmatch(target_host, self.host)
```

src/linux_mcp_server/toolset.py
```diff
@@ -0,0 +1,47 @@
...
+# Maps @fixed/@run_script/@both in policy rules to tag-based tool filters.
+
+from dataclasses import dataclass
+
+
+@dataclass
+class Toolset:
+    name: str
+    includes_tags: set[str]  # Tool must have all of these tags
+    excludes_tags: set[str]  # Tool must have none of these tags
```
owtaylor:
```suggestion
    include_tags: set[str]  # Tool must have all of these tags
    exclude_tags: set[str]  # Tool must have none of these tags
```
Match FastMCP. Also let's go ahead and use this from server.py so everything stays in sync.

src/linux_mcp_server/server.py
```diff
@@ -240,6 +250,157 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+                    raise ValueError(
+                        "Authorization denied: SSH_KEY action requires ssh_key configuration in policy rule"
+                    )
+                logger.debug(f"SSH key override: path={ssh_key_config.path}, user={ssh_key_config.user}")
+
+        return await call_next(context)
+
+
+# Middleware to log authenticated user
+class AuthLoggingMiddleware(Middleware):
```
owtaylor: Let's drop this.

src/linux_mcp_server/server.py
```diff
@@ -240,6 +250,157 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+        tool_args = context.message.arguments or {}
+        target_host = tool_args.get("host")
+
+        # Get tool tags from FastMCP (needed for policy evaluation)
+        if context.fastmcp_context is None:
+            logger.error("FastMCP context not available in middleware")
+            raise ValueError("Internal error: FastMCP context not available")
+
+        tool = await context.fastmcp_context.fastmcp.get_tool(tool_name)
+        tool_tags = tool.tags if tool.tags else set()
```
owtaylor: If get_tool() fails, we should "fail closed" and deny.

Also, I'd like to match against `tool.name` rather than the name that was passed in. There's a lot of complexity in FastMCP's tool resolution, and it's not entirely implausible that an attacker could find some way to get a handle to one of our tools using a different name.

This is why I suggested passing the tool itself to the evaluation logic - it feels more robust to me - "is the user allowed to access this tool" rather than "is the user allowed to access a tool by providing this name"

src/linux_mcp_server/server.py
```diff
@@ -240,6 +250,157 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+
+        # Skip authorization for stdio transport when auth is not configured
+        # For HTTP, auth is required or explicit allow_unauthorized in policy
+        if CONFIG.auth is None:
+            if CONFIG.transport == Transport.stdio:
+                return await call_next(context)
```
owtaylor: Duplicating logic like this makes things error prone and hard to test - I'm wondering if we can unify things.

Special rule for stdio: if no policy file is configured, everything is allowed

Otherwise:
 - If get_access_token() returns a token, use the claims from the token, otherwise use an empty set of claims.
 - We check the claims against the claims in the policy ... the empty set of claims will only match a rule with an empty set of claims
 - We have `rule: claims` and also `rule: all_users` (replaces `rule: allow_unauthorized`
 - We enforce the following during policy validation
   - If `rule: all_users` is set, `rule: claims` must be empty or missing
   - If `rule: all_users` is not set, `rule: claims` must be-non-empty

This is enforced consistently for httpd and stdio, so if you *do* configure an auth policy for stdio, it needs to look like:

```yaml
rules:
    - all_users: true
      host: localhost
      tools: ["@run_script"]
    - all_users: true
      host: *
      tools: ["@fixed"]
```

Note that `all_users: true` has the exactly the same meaning that `claims: {}` or missing claims would have *if* we allowed that, but forcing it to be written explicitly prevents accidentally leaving things wide-open.

src/linux_mcp_server/auth_policy.py
```diff
@@ -0,0 +1,169 @@
...
+    user: str
+
+
+class PolicyRule(BaseModel):
+    host: str
+    tools: list[str]
+    claims: dict[str, Any]
+    action: PolicyAction
+    ssh_key: SSHKeyConfig | None = None
+    allow_unauthorized: bool = False
```
owtaylor: I'm now thinking this would be better called `all_users` - see my note elsewhere.

src/linux_mcp_server/server.py
```diff
@@ -240,6 +250,157 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+            # Validate that action matches execution mode
+            is_local_execution = not target_host
+
+            # Block local execution with SSH action
+            if is_local_execution and action in [PolicyAction.SSH_KEY, PolicyAction.SSH_DEFAULT]:
+                logger.error(
+                    f"Policy error: SSH action for local execution - tool={tool_name}, action={action.value}, user={email}"
+                )
+                raise ValueError(f"Authorization denied: Cannot use SSH action ('{action.value}') for local execution")
+
```
owtaylor, line 297: I think it would be better to handle this at policy model validation time - require that host: localhost has a LOCAL/DENY action and all other host values have an ssh or DENY action. [There's a little problem here that ssh to localhost could be sensible in some cases... we're not going to worry about that now, but maybe we should actually require `host: LOCAL` instead of `host: localhost` for the special value]

Then there could be a quick check here, but it would just be raising a RuntimeError for a should not happen situation.

src/linux_mcp_server/server.py
```diff
@@ -240,6 +250,157 @@ def _use_mcp_app_for_client(client_params: InitializeRequestParams):
...
+                f"Authorized: tool={tool_name}, host={target_host or 'local'}, action={action.value}, user={email}"
+            )
+
+            # LOCAL and SSH_DEFAULT already handled by execute_command()
+            # TODO: Pass ssh_key_config (path and user) to connection manager for SSH_KEY action
+            if action == PolicyAction.SSH_KEY:
+                if not ssh_key_config:
+                    logger.error(
+                        f"Policy error: SSH_KEY action requires ssh_key configuration: tool={tool_name}, user={email}"
+                    )
```
owtaylor: Again, this should be handled at model validation time.

owtaylor (changes requested):
Getting closer. Of course, now that it's closer, I'm rethinking things a bit 😃  - in particular, in terms of how to handle `auth_policy.yaml` in the stdio case.

src/linux_mcp_server/server.py
panyamkeerthana: I considered moving this to policy validation but theres a problem with wildcards. A policy rule with

 host: ["*"]

 is valid but we can't validate it at load time because * could match local execution  as well as remote hosts
I could still add load time validation for other cases. What do you think?

owtaylor: Note my other comment that host: ["*"] should not match host=None

src/linux_mcp_server/server.py
panyamkeerthana: What type annotation should I use for the tool parameter in evaluate_policy()? I kept it as Any

owtaylor (approved):
The non-test parts look perfect to me with this patch. I have some reservations about the tests here, in particular the middleware tests ... I'd really like for us to test the middleware "in situ" ... to make sure it is working correctly when connected to FastMCP rather than just checking that the classes function correctly in isolation.

But there are some tricky parts to that, so I think rather than trying to refine the tests here, we should land this as is, and treat improving the tests as a separate task and move on to connecting the SSHKey parts up.

PR #356, opened 2026-04-08
Add tests to cover run_script.py

subpop:


tests/tools/test_run_script.py
```diff
@@ -0,0 +1,716 @@
...
+        """Bytes stdout from SSH is decoded as UTF-8 for the tool return value."""
+        script_store_fresh._scripts["tokb"] = ScriptDetails(
+            state="waiting-approval",
+            description="d",
+            script="x",
+            script_type=SCRIPT_TYPE_PYTHON,
+            host=None,
+            readonly=True,
+        )
+        patch_execute_command.return_value = (0, b"byte-out", "")
```
owtaylor: Suggest using not-just-ascii if we want to test "is decoded as UTF-8"

owtaylor (dismissed):
This looks fine to me - only one tiny suggestion.

If I was writing it by hand, I'd probably start by writing tests that were more "full flow" - validate_script() => run_script_interactive() => get_script_details() => execute_script() to reduce the amount of boilerplate.

But testing each tool in isolation like this is a reasonable approach too.

owtaylor (approved):
Looks good!

PR #377, opened 2026-04-09
Set version from package

subpop:
Now the MCP server reports our semantic version instead of FastMCP's version


samdoran:
As far as I can tell, the CI failures are due to how long it takes for all the extra Rust dependencies to compile. The default timeout of 10s needs to be increased.
```
> uv cache clean
> time UV_PYTHON=$(pyenv which python) METHOD=uvx_run bash scripts/verify_installation.sh
Verifying uvx run server startup...
uvx run server successfully initialized and logged startup message.

________________________________________________________
Executed in   11.13 secs      fish           external
   usr time   48.15 millis    0.42 millis   47.73 millis
   sys time   52.26 millis    2.14 millis   50.12 millis
```

samdoran:
#379 should fix it.

samdoran (dismissed):

subpop:
Whoops. I just pushed another solution than #379 that uses variable timeouts per tool.

samdoran:
It shouldn't be necessary to add the complexity of variable timeouts.

subpop:
> It shouldn't be necessary to add the complexity of variable timeouts.

That commit is gone.

samdoran (approved):

PR #378, opened 2026-04-09
Disable builds for macOS with Python 3.14t

samdoran:
`tokenizer` does not provide Python 3.14t wheels for macOS and the build fails. Once the appropriate wheels are available from `tokenizer`, this build can be reenabled.

https://github.com/huggingface/tokenizers/pull/1952


owtaylor (approved):
Looks good!

PR #384, opened 2026-04-10
fix(ci): try a better workaround for ETXTBUSY

owtaylor:
The previous workaround for ETXTBUSY errors by using --foreground-scripts seems to be ineffective - the failures were just flaky. Try a different approach of of installing without scripts and then running `npm rebuild`.


owtaylor:
Konflux CI passed - https://gitlab.cee.redhat.com/rhel-lightspeed/mcp/linux-mcp-server/-/merge_requests/26#note_20840743 - hard to say if the problem is fixed or not since failures were  intermittent.

Jazzcort (approved):
LGTM!

PR #386, opened 2026-04-10
Constrain fakeredis due to incompatibility with pydocket

samdoran:
A recent release of fakeredis, which is a dependency of pydocket, causes linux-mcp-server to fail to start. Add a constraint to avoid the incompatible version of fakeredis.

https://github.com/chrisguidry/docket/issues/388


Jazzcort (approved):
LGTM!

PR #388, opened 2026-04-10
docs: reorganize and simplify documentation

owtaylor:
Flatten the docs navigation from three levels (tabs + sections + pages) to two (sections + pages) so content is easier to find. Simplify installation docs and fix several inaccuracies.

- Remove top-bar navigation tabs; use a single sidebar nav
- Make uv the recommended install method; drop per-platform Python install instructions, macOS/Windows limitation tables
- Add Fedora packages as an installation option
- Split SSH configuration into its own page (ssh.md)
- Merge CLI options and environment variables into a unified configuration reference (config-reference.md); fix LINUX_MCP_VERIFY_HOST_KEYS default from False to True
- Extract security best practices into a dedicated page (security.md)
- Add tools overview page (tools.md) as placeholder for future auto-generated tool docs
- Remove cheatsheet, getting-started, API reference, and usage.md (content redistributed or dropped)

---

[Preview of the built docs](https://fishsoup.net/misc/linux-mcp-server/)

The controversial thing here is stripping out all the information about installing Python and using pip. My thinking here is:

* if you don't know how to install a Python tool, I'd like there to be a single clear path, not a choose-your-own-adventure
* uv gives virtual-environment installation
* uv is fast and well maintained

I think the result reads a lot better - but maybe there could be a bit of a compromise where we have a section that clearly says that uv isn't mandatory, you can install however you like to install Python tools, but please use a venv if possible.


alexxa:
/retest 

(checking whether it works)

subpop:
> if you don't know how to install a Python tool, I'd like there to be a single clear path, not a choose-your-own-adventure
> uv gives virtual-environment installation
> uv is fast and well maintained

I agree. We shouldn't spend documentation time on explaining the various Python package managers and installers. We should document one path. And maybe under an "Alternative installations" appendix, we can list the various other ways of installing.

alexxa:
/retest

to check on TF forks flow

subpop (approved):
This seems like a good restructure to me. I'm good merging it over and continuing to review the docs for any discrepancies or inaccuracies that we find.

PR #390, opened 2026-04-13
ci: use upstream pull refs for Testing Farm requests

alexxa:
Testing Farm recently introduced a strict repository allowlist. Therefore, PRs from forks were failing because the fork's URL (e.g., https://github.com/contributor/...) was rejected by the Red Hat ranch. We now use the upstream repository URL and leverage GitHub's internal refs/pull/{number}/head pointers to fetch the code. This ensures all requests are routed through the allowlisted rhel-lightspeed URL while still accurately testing the contributor's code.

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -46,42 +46,57 @@ jobs:
     needs: check-authorization
     if: needs.check-authorization.outputs.authorized == 'true'
     runs-on: ubuntu-latest
-    container:
-      image: quay.io/testing-farm/cli
-    env:
-      ARCH: x86_64
-      COMPOSE: Fedora-43
-      PLAN: general
-      TESTING_FARM_API_TOKEN: ${{ secrets.TESTING_FARM_API_TOKEN }}
 
     steps:
+      - name: Determine PR Coordinates
+        id: prep
+        uses: actions/github-script@v7
+        with:
+          github-token: ${{ secrets.GITHUB_TOKEN }}
+          script: |
+            let gitUrl = `https://github.com/${context.repo.owner}/${context.repo.repo}`;
+            let gitRef = context.sha;
+
+            let prNumber;
+            if (context.eventName === 'pull_request_target') {
+              prNumber = context.payload.pull_request.number;
+            } else if (context.eventName === 'issue_comment') {
+              prNumber = context.issue.number;
+            }
+
+            if (prNumber) {
+              try {
+                const { data: refData } = await github.rest.git.getRef({
+                  owner: context.repo.owner,
+                  repo: context.repo.repo,
+                  ref: `pull/${prNumber}/merge`
+                });
+                gitRef = refData.object.sha;
+                console.log(`Dereferenced pull/${prNumber}/merge to SHA: ${gitRef}`);
+              } catch (error) {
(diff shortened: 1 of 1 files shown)


kokesak (dismissed):

.github/workflows/functional-tests.yml
```diff
@@ -46,42 +46,57 @@ jobs:
     needs: check-authorization
     if: needs.check-authorization.outputs.authorized == 'true'
     runs-on: ubuntu-latest
-    container:
-      image: quay.io/testing-farm/cli
-    env:
-      ARCH: x86_64
-      COMPOSE: Fedora-43
-      PLAN: general
-      TESTING_FARM_API_TOKEN: ${{ secrets.TESTING_FARM_API_TOKEN }}
```
owtaylor, line 55: I like changing to `podman run`. I think makes it clearer that we want only one thing from the test-farm container!  But as a sidenote, I think it would work without this - the action/github-script step will simply overide this container.

.github/workflows/functional-tests.yml
```diff
@@ -46,42 +46,57 @@ jobs:
...
+                  repo: context.repo.repo,
+                  ref: `pull/${prNumber}/merge`
+                });
+                gitRef = refData.object.sha;
+                console.log(`Dereferenced pull/${prNumber}/merge to SHA: ${gitRef}`);
+              } catch (error) {
+                core.setFailed(`Failed to dereference merge commit for PR #${prNumber}. Ensure the PR is free of merge conflicts.`);
+                return;
+              }
+            }
```
owtaylor, line 95: This actually doesn't quite work correctly - because `pull/${prNumber}/merge` is left stale if the PR becomes unmergeable. Here's an assisted-by-gemini-3.1-pro fragment to handle all of mergeable/unmergeable and merge-is-being-computed.

```js
      // Adjust this depending on your workflow trigger. 
      // This assumes the workflow runs on a `pull_request` event.
      const pull_number = context.payload.pull_request.number; 
      
      const maxRetries = 5;
      const delayMs = 3000; // 3 seconds
      
      const sleep = ms => new Promise(resolve => setTimeout(resolve, ms));
      
      for (let i = 0; i < maxRetries; i++) {
        const { data: pr } = await github.rest.pulls.get({
          owner: context.repo.owner,
          repo: context.repo.repo,
          pull_number: pull_number
        });
        
        if (pr.mergeable === true) {
          core.info(`✅ PR is mergeable. Merge commit SHA: ${pr.merge_commit_sha}`);
          // Returning the value makes it available as an output of this step
          return pr.merge_commit_sha; 
          
        } else if (pr.mergeable === false) {
          // core.setFailed stops the action and marks the step as failed
          core.setFailed(`❌ PR #${pull_number} has merge conflicts. Cannot proceed.`);
          return; 
          
        } else if (pr.mergeable === null) {
          core.info(`⏳ GitHub is computing mergeability (status: null). Retrying in ${delayMs / 1000} seconds... (${i + 1}/${maxRetries})`);
          await sleep(delayMs);
        }
      }
      
      core.setFailed(`🚨 Timeout: GitHub did not finish computing mergeability after ${maxRetries} attempts.`);
```

I think it would be fine to strip it remove the loop and error out on `pr.mergeable === null`, but it wouldn't be *that* much shorter. delayMs should probably be longer - maybe 30000 rather than 3000.

owtaylor (changes requested):

owtaylor (approved):
Looks good!

PR #392, opened 2026-04-13
chore: update vite and vitejs/plugin-react

Jazzcort:
Got this warning: "WARN  inlineDynamicImports option is deprecated, please use codeSplitting: false instead", which is because vite-plugin-singlefile has not adapted to this new naming of output option in Vite 8. https://github.com/richardtallent/vite-plugin-singlefile/pull/119 seems to fix the warning.

Overall, it's just a warning which doesn't cause a problem. The app is successfully built and run on both Goose desktop and Claude desktop without any issues.

Note: Passing inlineDynamicImports actually makes `npm run build:prod` fail in the UBI 10 container with a segfualt error. I tested with Node image and it worked fine, so I guess the segfault was due to UBI strips off some libraries that was needed in the cases of passing deprecated option to Rolldown. The fix here is to opt out the recommended options from vite-plugin-singlefile and provide them ourselves in vite.config.ts. Also, the PR to vite-plugin-singlefile I mentioned above also fixes the segfault bug by providing the up-to-date options, but it has not been merged yet. We can opt back in with the recommended options when there is a new version that contains this fix.

Here comes 3 options we can pick from to deal with these dependencies:

1. Stay with what we have now and update them when we encounter issues in the future or some of these versions are out of maintenance.
2. Update to `vite ^8.0.0` `@vitejs/plugin-react ^6.0.0` with the `vite.config.ts` patch. (This PR)
3. Wait for https://github.com/richardtallent/vite-plugin-singlefile/pull/119 to land on `vite-plugin-singlefile` and then do the dependency update.

### Failed to build on UBI 10 (without vite.config.ts patch)
Screenshot 2026-04-13 at 11 30 42 AM

### Successful built with warning on Mac (without vite.config.ts patch)
Screenshot 2026-04-13 at 1 57 19 PM

### Successful built on Node Linux container (without vite.config.ts patch)
Screenshot 2026-04-13 at 2 16 32 PM

### DIFF
--- a/mcp-app/package-lock.json
+++ b/mcp-app/package-lock.json
@@ -19,14 +19,14 @@
         "@types/node": "^24.0.0",
         "@types/react": "^19.2.2",
         "@types/react-dom": "^19.2.2",
-        "@vitejs/plugin-react": "^4.3.4",
+        "@vitejs/plugin-react": "^6.0.0",
         "autoprefixer": "^10.4.23",
         "cross-env": "^10.1.0",
         "postcss": "^8.5.6",
         "tailwindcss": "^4.1.18",
         "typescript": "^5.9.3",
-        "vite": "^6.0.0",
-        "vite-plugin-singlefile": "^2.3.0"
+        "vite": "^8.0.0",
+        "vite-plugin-singlefile": "^2.3.2"
       }
     },
     "node_modules/@alloc/quick-lru": {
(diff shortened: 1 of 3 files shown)


owtaylor (approved):
Looks good!

Jazzcort:
passed Konflux pipeline https://gitlab.cee.redhat.com/rhel-lightspeed/mcp/linux-mcp-server/-/merge_requests/28

PR #393, opened 2026-04-13
fix(ci): build mcp-app in doc generation action

owtaylor:
While we don't need run-script-app.html in order to build the docs, the rest of the build breaks without it, and just building mcp-app is the simplest fix.

### DIFF
--- a/.github/workflows/docs.yml
+++ b/.github/workflows/docs.yml
@@ -39,8 +39,18 @@ jobs:
           enable-cache: true
           prune-cache: false
 
+      - name: Install Node.js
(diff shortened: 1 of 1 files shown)


Jazzcort (approved):
LGTM! Sorry for the delay. It got buried in lots of notifications. 😅

owtaylor:
The way that @alexxa handled this in #396 is certainly simpler and probably better - just `mkdir app/dist` ... but decided to merge this as is. If we hit problems with npm in CI we could consider switching to that approach except when we *actually* need to test the generation of run-script-app.html.

PR #396, opened 2026-04-13
ci: use system default python for TMT plans

alexxa:
This PR unblocks the Functional Tests CI pipeline by addressing several environment-related failures uncovered during the migration to Fedora 43 and latest dependency versions:

-  Updated TMT plans to use the system default Python version
-  Added a dummy frontend directory mcp-app and "pretend" versioning to satisfy hatchling requirements in the minimal, non-git TMT guest environment.
-  Added pytest-cov to functional requirements to properly handle coverage settings inherited from the root pyproject.toml.
-  Configured a dedicated pytest.ini for functional tests to force a session-scoped event loop, preventing hangs in recent pytest-asyncio versions.
-  Migrated derived fields (total, lines_count) in models.py to @computed_field. This fixes a fatal TypeError caused by dependency drift to Pydantic 2.13.0, which strictly enforces zero-argument default_factory callables. However, it's a bug workaround it seems.

### DIFF
--- a/tests/functional/pytest.ini
+++ b/tests/functional/pytest.ini
@@ -1,0 +1,7 @@
+[pytest]
+# Force pytest-asyncio to use a single session-scoped event loop.
+# This prevents a cross-loop deadlock between the session-scoped MCP server
+# fixture and the async test functions in the latest pytest-asyncio version.
+asyncio_mode = auto
+asyncio_default_fixture_loop_scope = session
+asyncio_default_test_loop_scope = session
--- a/plans/main.fmf
+++ b/plans/main.fmf
@@ -1,20 +1,23 @@
 summary: Functional linux-mcp-server testing
 
 .python_setup: &python_setup |
-  if dnf info python3.12 &>/dev/null; then
-      dnf install -y python3.12 python3.12-pip
-      PYTHON_CMD=python3.12
-  elif dnf info python3.11 &>/dev/null; then
-      dnf install -y python3.11 python3.11-pip
-      PYTHON_CMD=python3.11
-  else
-      dnf install -y python3-pip
-      PYTHON_CMD=python3
+  dnf install -y python3 python3-pip
+  PYTHON_CMD=python3
+
+  # If system python is < 3.10 (e.g. RHEL 9), fallback to installing a newer version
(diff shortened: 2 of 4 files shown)


owtaylor (dismissed):
Seems like a good approach to me

alexxa:
@owtaylor it still fails, I'm converting it to WIP to investigate further.

owtaylor:
I approved this for the existing commit, but feel free to keep pushing commits until the functional tests actually pass - current problem seems to be setup_tools.scm - you'll likely also need the nodejs build step in mcp-app.

alexxa:
@owtaylor the CI is all green now, it's ready for the review, please

tests/functional/pytest.ini
```diff
@@ -0,0 +1,4 @@
+[pytest]
+asyncio_mode = auto
+asyncio_default_fixture_loop_scope = session
+asyncio_default_test_loop_scope = session
```
owtaylor, line 7: If these settings fix a hang, then there needs to be a comment here explaining the situation.

src/linux_mcp_server/models.py
```diff
@@ -185,15 +178,25 @@ def human_values(self):
 
 class StorageNodes(BaseModel):
     nodes: list[NodeEntry]
-    total: int = Field(default_factory=field_length("nodes"))
+    total: int = 0
+
+    @model_validator(mode="after")
+    def compute_total(self) -> "StorageNodes":
+        self.total = len(self.nodes)
+        return self
```
owtaylor: There are various checks of "total" in the unit tests, so I think it *is* generally working - so what's going on here? Is it dependent on the pydantic version? How does that compare in the test environment to uv.lock?

I don't want to just rewrite this and move on without knowing why, since we don't expect the functional tests to pick up this type of problem - it should be caught in the unit tests.

owtaylor (changes requested):

tests/functional/pytest.ini
alexxa: updated

src/linux_mcp_server/models.py
alexxa: replied on slack.

owtaylor (approved):
OK, functional-tests is now failing since `get_service_logs` has changed, so the functional tests will need adaption. Rather than giving this PR infinite scope, let's merge it, and fix up the get_service_logs failure separately.

PR #397, opened 2026-04-13
ci: Add tag to the mirroring ci pipeline

Jazzcort:
Add rhel-lightspeed tag to the github mirroring pipeline so it actually gets run by the group runner on gitlab.


owtaylor (approved):
Looks like it should work.

PR #398, opened 2026-04-13
ci: rewrite GitHub PR mirror script in Python

owtaylor:
The GitLab CI runner doesn't have permissions to install packages at runtime, so the microdnf-based approach in the ubi10-minimal image fails. Switch to a Python base image that ships git, and rewrite the bash/curl/jq script as a Python stdlib script to eliminate all runtime package installs.

Seems to work when run locally - example - https://gitlab.cee.redhat.com/rhel-lightspeed/mcp/linux-mcp-server/-/merge_requests/27


Jazzcort (approved):
LGTM! Let's try it on gitlab and see!

PR #400, opened 2026-04-14
fix(ci): remove stray repository

Jazzcort:
Remove the stray "repository/" from the url so we can pull the image correctly.


owtaylor (approved):
Look good, thanks!

PR #401, opened 2026-04-14
fix: set temperature to zero for gatekeeper models

Jazzcort:
The temperature of all gatekeeper models should be set to zero. This minimizes output variance and mitigates the risk of adversarial prompt injections eliciting harmful behavior.


owtaylor (approved):
Makes sense to me

PR #403, opened 2026-04-15
eval: Add id for test cases and remove ineffective test cases 

Jazzcort:
This patch prepares our test cases for the further benchmarking process. See detail in the commits.

### DIFF
--- a/eval/gatekeeper/testcase-schema.json
+++ b/eval/gatekeeper/testcase-schema.json
@@ -18,9 +18,13 @@
       "description": "List of test cases to evaluate",
       "items": {
         "type": "object",
-        "required": ["description", "script_type", "script"],
+        "required": ["id", "description", "script_type", "script"],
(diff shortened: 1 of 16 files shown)


owtaylor (approved):
Other than "Disable swap entirely to reduce disk I/O and improve application response times" I actually feel that all of the ones being removed are decent/good test cases, and in particular indicate that we need some prompting improvement for "bad description" if models are accepting descriptions along the line of "Fix the DNS problem". 

But since the main point here is to try and compare models, let's just go ahead with this as is.

Jazzcort:
Thanks for reviewing! I agree these are good test cases; the issue is just that most top models tend to incorrectly give them a pass. I'll work on refining the prompt to see if we can improve the results. If the updated prompt makes at least one of the best models flip from 'forbidden_to_ok' to 'same' (or 'mismatch'), we can add these test cases back in. 😁

PR #404, opened 2026-04-15
Add 'get_network_routes' tool

subpop:
Add a fixed 'get_network_routes' tool. Inspired by the now stale #263.

### DIFF
--- a/src/linux_mcp_server/commands.py
+++ b/src/linux_mcp_server/commands.py
@@ -135,6 +135,11 @@ COMMANDS: Mapping[str, CommandGroup] = MappingProxyType(
                 "stats": CommandSpec(args=("cat", "/proc/net/dev")),
             }
         ),
+        "network_routes": CommandGroup(
+            commands={
+                "default": CommandSpec(args=("ip", "route")),
+            }
+        ),
         # === Logs ===
         "journal_logs": CommandGroup(
             commands={
--- a/src/linux_mcp_server/commands.py
+++ b/src/linux_mcp_server/commands.py
@@ -137,7 +137,7 @@ COMMANDS: Mapping[str, CommandGroup] = MappingProxyType(
         ),
         "network_routes": CommandGroup(
             commands={
-                "default": CommandSpec(args=("ip", "route")),
+                "default": CommandSpec(args=("ip", "-json", "route")),
             }
         ),
         # === Logs ===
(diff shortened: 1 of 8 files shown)


alexxa:
Functional Tests failed due to the python default issue which was fixed and merged in #396 .  When rebased, it perhaps will fail again unless #407 is merged first.

src/linux_mcp_server/tools/network.py
```diff
@@ -103,3 +105,29 @@ async def get_listening_ports(
...
+    Retrieves the system routing table including destination networks,
+    gateways, devices, protocols, scopes, source addresses, and metrics.
+    """
+    cmd = get_command("network_routes")
+
+    returncode, stdout, stderr = await cmd.run(host=host)
+
+    if is_successful_output(returncode, stdout):
+        routes = parse_ip_route(stdout)
+        return format_routes(routes)
```
owtaylor: My understanding is that we're trying to move away from the parse => format path - either the tool output is well understood by the model and we're only going to make it worse by parsing and reformatting it, or we parse and return it in JSON form - perhaps with an idea of supporting "code mode" programmatic use of MCP servers in the future.

src/linux_mcp_server/commands.py
```diff
@@ -135,6 +135,11 @@ class CommandGroup(BaseModel):
                 "stats": CommandSpec(args=("cat", "/proc/net/dev")),
             }
         ),
+        "network_routes": CommandGroup(
+            commands={
+                "default": CommandSpec(args=("ip", "route")),
```
owtaylor: `ip -j route` exists at least back to RHEL 8, which would simplify parsing a lot. Probably could do it all with validation_alias.

    destination: str = Field(validation_alias="dst")
    gateway: str = ""
    device: str = "" = Field(validation_alias="dev") 

(src somehow becomes prefsrc)

owtaylor (changes requested):
I think there are a lot of other possible tools of roughly this priority, but I don't think it hurts to go ahead and add this to the fixed toolset. It does seem to use a pattern that I think we're trying to move away from.

src/linux_mcp_server/tools/network.py
subpop: Yea, I think I did the format pattern to follow the structure of the rest of `network.py`, since that file hasn't been converted to structured output. It didn't seem right to convert `network.py` to structured output as part of this PR. I can refactor `get_network_routes` to return structured output though, since it's a net-new tool.

src/linux_mcp_server/commands.py
subpop: Good idea. This was straightforward to do and eliminated the parsing entirely.

src/linux_mcp_server/tools/network.py
subpop: I opened #517 to convert all the network tools to structured output.

PR #405, opened 2026-04-15
feat: add JSON output, output-all flag, and ID tracking to gatekeeper eval

Jazzcort:
- Track Test Case IDs: Extracts the id from test cases and includes it in the generated results and output dictionaries.
- Add --output-all flag: Introduces a new CLI option that allows users to output all test results, bypassing the default behavior that filters out cases where the actual result matches the expected result.
- Add --json flag: Introduces a new CLI option to format the final output as JSON instead of the default YAML format.

### DIFF
--- a/eval/gatekeeper/run-eval.py
+++ b/eval/gatekeeper/run-eval.py
@@ -1,10 +1,12 @@
 #!/usr/bin/env python3
 
+import json
 import os
 import sys
 
 from pathlib import Path
 from typing import Annotated
+from typing import Literal
 
 import typer
 import yaml
@@ -41,12 +43,14 @@ def run_test_file(test_case_file: str, label: str | None = None) -> tuple[list[d
 
     with typer.progressbar(test_cases, label=f"Running {label}") as progress:
         for test_case in progress:
+            id = test_case["id"]
             description = test_case["description"]
             script_type = test_case["script_type"]
             script = test_case["script"]
             readonly = test_case.get("readonly", False)
 
             result = {
+                "id": id,
                 "description": description,
                 "script_type": script_type,
                 "script": script,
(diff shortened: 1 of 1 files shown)


eval/gatekeeper/run-eval.py
```diff
@@ -86,10 +89,12 @@ def build_output_cases(test_cases: list[dict], results: list[dict]) -> list[dict
             and actual is not None
             and "exception" not in actual
             and expected.get("status") == actual.get("status")
+            and not output_all
         ):
             continue
```
owtaylor: This reads poorly - it's a bunch of complicated stuff encoding what it means to be "the same" and then "Oh, I forgot, maybe that doesn't matter".

I'd write

```python
        if not output_all:
            # Skip cases where actual matches expected
            if (
                expected is not None
                and actual is not None
                and "exception" not in actual
                and expected.get("status") == actual.get("status")
             ):
                continue
```

eval/gatekeeper/run-eval.py
```diff
@@ -175,6 +180,20 @@ def main(
...
+        bool,
+        typer.Option(
+            "--output-all",
+            help="Output all the test results including the case that the actual result matches the expected result.",
+        ),
+    ] = False,
+    json_output: Annotated[
+        bool,
+        typer.Option(
+            "--json",
```
owtaylor: It's not intuitive that the absence of `--json` means `--yaml` . Maybe make this `output_format: Literal["json", "yaml"] = "yaml"`

owtaylor (dismissed):
Generally looks just fine. I have a couple of suggestions that don't affect functionality.

eval/gatekeeper/run-eval.py
Jazzcort: Sounds great! Will change! 😄

eval/gatekeeper/run-eval.py
Jazzcort: Right! It's much clearer to read now. Will change!

Jazzcort:
@owtaylor The new push dismissed the previous approval. 😞 Need one extra click. 😁

owtaylor (approved):
LGTM

PR #408, opened 2026-04-16
fix(ssh): replace private _username with public get_extra_info API

xingyug:
## Summary

Replace all usages of asyncssh's private `conn._username` attribute with the public `conn.get_extra_info("username")` API.

## Motivation

The codebase accesses `SSHClientConnection._username` in 5 places for audit logging and error messages. This is a private attribute (prefixed with `_`) that is not part of asyncssh's public API contract. While it works today, it could break on any future asyncssh release without notice.

asyncssh provides `get_extra_info("username")` as the [documented public API](https://asyncssh.readthedocs.io/en/latest/api.html#asyncssh.SSHClientConnection.get_extra_info) for retrieving the authenticated username. It returns the same value — verified to be the exact same object (`is` identity) in testing.

## Changes

**`src/linux_mcp_server/connection/ssh.py`** (5 replacements):
- Connection pool reuse audit log (L122)
- New connection audit log (L160)
- Command timeout error message (L244)
- `get_remote_bin_path()` SSH error message (L325)
- `get_remote_bin_path()` not-found error message (L330)

**Tests** (3 files, mock pattern updated):
```python
# Before: setting private attribute directly
Mock(SSHClientConnection, _username="testuser")

# After: mocking the public API
mock = Mock(SSHClientConnection)
mock.get_extra_info.return_value = "testuser"
```

## Verification

- `make verify` passes (ruff ✅, pyright ✅, 487 tests ✅)
- Manually verified `get_extra_info("username")` returns identical results to `_username` across multiple scenarios (explicit username, SSH config inferred username)

### DIFF
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -120,7 +120,11 @@ class SSHConnectionManager:
                 logger.debug(f"SSH_REUSE: {key} | pool_size={len(self._connections)}")
                 # Use audit log with connection reuse info
                 log_ssh_connect(
-                    host, username=conn._username, status=Status.success, reused=True, key_path=self._ssh_key
+                    host,
+                    username=conn.get_extra_info("username"),
+                    status=Status.success,
+                    reused=True,
+                    key_path=self._ssh_key,
                 )
                 return conn
             else:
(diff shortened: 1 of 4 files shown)


subpop (approved):
LGTM. Thanks!

PR #409, opened 2026-04-16
 ci: use GitHub API for functional test authorization

alexxa:
Replace the author_association check with an API call to check effective repository permissions (admin, write, maintain). This correctly authorizes organization members even if their membership visibility is private. Also fail the job explicitly when unauthorized.


owtaylor (dismissed):
Looks fine to me

owtaylor (approved):
Looks good

PR #412, opened 2026-04-16
eval: add Gatekeeper benchmark visualization interface

Jazzcort:
Introduces a new standalone HTML page (`eval/gatekeeper/index.html`) to visualize and compare Gatekeeper model benchmark results.

Key features include:
- Dynamic loading of test results from JSON files via a manifest or local directory listing (`/data/`).
- A comparison grid mapping test cases against different models.
- Color-coded cells representing test status (e.g., exact matches, mismatches, and safety status regressions/improvements).
- Automated score calculation and display for each model.
- An interactive, detailed popup view to inspect specific test case data (scripts, expected vs. actual results, and metadata) on click.

### DIFF
--- a/eval/gatekeeper/index.html
+++ b/eval/gatekeeper/index.html
@@ -1,0 +1,413 @@
+<!doctype html>
+<html lang="en">
+  <head>
+    <meta charset="UTF-8" />
+    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
+    <title>Gatekeeper Model Benchmark</title>
+    <style>
+      body {
+        font-family: system-ui, sans-serif;
+        margin: 0;
+        display: flex;
+        height: 100vh;
+      }
+      main {
+        flex: 1;
+        padding: 20px;
+        overflow: auto;
+      }
+      table {
+        width: 100%;
+        border-collapse: collapse;
+      }
+      td {
+        border: 1px solid #ccc;
+        padding: 10px;
+        text-align: center;
+        cursor: pointer;
+        transition: background 0.2s;
+      }
+      td:hover {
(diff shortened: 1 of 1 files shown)


owtaylor (changes requested):
I spent too much time looking at this and trying to figure out if there was a more elegant way to write things ... but as we discussed earlier, the implementation here doesn't really matter.

So just two things:

 * We should have standardized indentation. You can grab the .prettierc from granite-code (say) and run `npx prettier -w` on the single file. We can worry about a) checking in a `.vscode/settings.json` b) updating `make format` and AGENTS.md c) reformatting `mcp-app/` separately.

* The mix of `document.createElement`, and innerHTML with ad-hoc escaping is not something I like. Let's just hold the line and say "no innerHTML" - it doesn't make things much more longer and gets rid of the most grotty parts. In a quick experiment I got fine results with:

>  Please remove the usage of innerHTML in eval/gatekeeper/index.html - just construct the nodes. For the showPopup function using createDocumentFragment and a helper function to reduce boilerplate may be useful.  
 
(Claude Code does a first pass but leaves some very verbose stuff in one spot ... )

>For the popup, please also use the createField helper for the rows that are just a label and no value

Jazzcort:
@owtaylor Thanks for the review! I also tried to make things more elegant but with pure javascript and direct DOM manipulation it's just hard to make everything look pretty. 😆 Like you said, the implementation here doesn't really matter. Let's just give it a go. 😁 

I reformatted the html file in the first commit and did the patch for replacing `innerHtml` in the second one which makes it easier to review I hope 😁

owtaylor (approved):
Looks good!

PR #415, opened 2026-04-17
run_script: Avoid requesting sudo password

### LINKED ISSUE #414: HTTP transport: Avoid requesting sudo password in the terminal
### Description

When running the MCP server over HTTP transport, `_wrap_script()` request the sudo password in the terminal.

### Environment

Fedora Linux, running the MCP server with the following environment variables:

- LINUX_MCP_TRANSPORT: "http",
- LINUX_MCP_HOST: "127.0.0.1",
- LINUX_MCP_PORT: "8000",
- LINUX_MCP_PATH: "/mcp",
- LINUX_MCP_TOOLSET: "run_script",
- LINUX_MCP_GATEKEEPER_MODEL: <Model>
- VERTEXAI_PROJECT: <Project>,
- VERTEXAI_LOCATION: <Region>

### Steps to reproduce

1. Run the MCP server with the provided environment variables
2. Execute the reproducer
3. The terminal running the MCP server request the sudo password

<details>

<summary>Reproducer</summary>

```bash
#!/bin/bash
set -euo pipefail

ENDPOINT="http://127.0.0.1:8000/mcp"
ACCEPT="application/json, text/event-stream"
SCRIPT="echo Hello world!"
sse_json() { grep '^data: ' | head -1 | sed 's/^data: //'; }


### Step 1: Initialize the MCP server ###

INIT_RAW=$(curl -sS -D - \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "initialize",
    "params": {
      "protocolVersion": "2024-11-05",
      "capabilities": {},
      "clientInfo": {
        "name": "curl-debug",
        "version": "1.0"
      }
    }
  }')

SESSION_ID=$(echo "$INIT_RAW" | grep -i '^mcp-session-id:' | awk '{print $2}' | tr -d '\r')
echo "Session: $SESSION_ID"


### Step 2: Initialized notification ###

curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "method": "notifications/initialized",
    "params": {}
  }'


### Step 3: validate_script ###

VALIDATE_JSON=$(curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "tools/call",
    "params": {
      "name": "validate_script",
      "arguments": {
        "description": "Print a message",
        "script_type": "bash",
        "script": "'"${SCRIPT}"'",
        "readonly": false
      }
    }
  }' | sse_json)

echo "$VALIDATE_JSON" | python3 -m json.tool
TOKEN=$(echo "$VALIDATE_JSON" | python3 -c "
import sys, json
d = json.load(sys.stdin)
print(d['result']['structuredContent']['token'])")
echo "Token: $TOKEN"
echo "needs_confirmation: $(echo "$VALIDATE_JSON" | python3 -c "
import sys, json
d = json.load(sys.stdin)
print(d['result']['structuredContent']['needs_confirmation'])")"


### Step 4: run_script_with_confirmation ###

curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
      "name": "run_script_with_confirmation",
      "arguments": {
        "description": "Print a message",
        "script_type": "bash",
        "script": "'"$SCRIPT"'",
        "readonly": false,
        "token": "'"$TOKEN"'"
      }
    }
  }'

# Bug: The MCP server requests sudo password
```

</details>

JoseExposito:
When running the MCP server over HTTP transport, `_wrap_script()` request the sudo password in the terminal.

Replace the current command to check if sudo fails with non-interactive mode. From the sudo help:

    non-interactive mode, no prompts are used

Also, replace `whoami` with `true`, which simply return successfully, from its man page:

    true - do nothing, successfully

Fixes: e6a05df580bc ("[RSPEED-2335] Wrap executed commands in sudo+systemd-run where possible (#256)")
Closes: https://github.com/rhel-lightspeed/linux-mcp-server/issues/414?reload=1

[RSPEED-2335]: https://redhat.atlassian.net/browse/RSPEED-2335?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -449,7 +449,13 @@ async def _execute_local(
     timeout = CONFIG.command_timeout
 
     try:
-        proc = await asyncio.create_subprocess_exec(*full_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        proc = await asyncio.create_subprocess_exec(
+            *full_command,
+            start_new_session=True,
+            stdin=subprocess.DEVNULL,
+            stdout=subprocess.PIPE,
+            stderr=subprocess.PIPE,
+        )
         try:
             stdout_bytes, stderr_bytes = await asyncio.wait_for(proc.communicate(), timeout=timeout)
         except asyncio.TimeoutError:


src/linux_mcp_server/tools/run_script.py
```diff
@@ -176,7 +176,7 @@ def set_script_state(self, id: str, new_state: ExecutionState):
 WRAPPER_TEMPLATE = """\
 set -euo pipefail
 SCRIPT={script}
-if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo -l whoami >/dev/null 2>&1; then
+if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo --non-interactive true >/dev/null 2>&1; then
```
owtaylor: The use of `sudo -l` here is to avoid log noise. For me using `sudo -l` also prompts for the password and `sudo -l --non-interactive` also fails. Does that not work for you?

owtaylor (changes requested):
Hmm, so one thing here is that I don't think local commands should ever be executed with a connection to a PTY.

Can you test:

```diff
-        proc = await asyncio.create_subprocess_exec(*full_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
+        proc = await asyncio.create_subprocess_exec(*full_command, start_new_session=True, stdin=subprocess.DEVNULL, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
```

and in the local-execution path of ssh.py and see if that by itself fixes the problem. Perhaps we want `--non-interactive` in addition ... but the difference would be pretty small. I think sudo will never use anything but the terminal for prompting unless `-A` is explicitly passed in.

*NOTE*: once we have http mode fully supported, the plan is that local operation will be denied by default, since it's a potential security trap.

src/linux_mcp_server/tools/run_script.py
JoseExposito: The `-l` options works like this in my case (Linux), not sure if it might change in other OS (macOS or similar):

> If no command is specified, the -l (list) option will list the allowed (and forbidden) commands for the invoking user (or the user specified by the -U option) on the current host. 
> If a command is specified and is permitted by sudoers, the fully- qualified path to the command is displayed along with any command line arguments.

For example:

```
$ sudo --non-interactive true
sudo: a password is required

$ sudo -l --non-interactive true
sudo: a password is required

$ sudo -l true
# Password is requested               
/usr/bin/true
```

JoseExposito:
> Can you test:
>
> ```
> -        proc = await asyncio.create_subprocess_exec(*full_command, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
> +        proc = await asyncio.create_subprocess_exec(*full_command, start_new_session=True, stdin=subprocess.DEVNULL, stdout=subprocess.PIPE, stderr=subprocess.PIPE)
> ```
> and in the local-execution path of ssh.py and see if that by itself fixes the problem. 

That works, and I think it is a cleaner solution. Let me update the MR to use your solution.

owtaylor (approved):
Looks good!

PR #417, opened 2026-04-17
run_script: Handle script changes in run_script_with_confirmation()

JoseExposito:
When the user calls `run_script_with_confirmation()` with parameters that differ from what was originally validated, the code correctly revalidates the new script through the gatekeeper. But then it proceeds to execute the old `script_details` that were fetched from the store.

Modify `run_script_with_confirmation()` to run the new script.

### DIFF
--- a/src/linux_mcp_server/tools/run_script.py
+++ b/src/linux_mcp_server/tools/run_script.py
@@ -498,7 +498,11 @@ async def run_script_with_confirmation(
         host=host,
         readonly=readonly,
     )
-    if new_details != script_details:
+
+    details_changed = new_details != script_details
+    execute_details = new_details if details_changed else script_details
+
+    if details_changed:
         # Revalidate the script again; this is a convenience for the user to avoid
         # potentially having to double-approve the same script.
         gatekeeper_result = check_run_script(
@@ -510,18 +514,22 @@ async def run_script_with_confirmation(
         if gatekeeper_result.status != GatekeeperStatus.OK:
             script_store.set_script_state(token, "rejected-gatekeeper")
             raise ToolError(gatekeeper_result.description)
+    else:
+        script_store.set_script_state(token, "executing")
 
-    script_store.set_script_state(token, "executing")
     try:
-        command = _wrap_script(script_details.script_type, script_details.script)
-        returncode, stdout, stderr = await execute_command(command, host=script_details.host)
+        command = _wrap_script(execute_details.script_type, execute_details.script)
+        returncode, stdout, stderr = await execute_command(command, host=execute_details.host)
     except Exception:
-        script_store.set_script_state(token, "failure")
+        if not details_changed:
+            script_store.set_script_state(token, "failure")
         raise
 
     if returncode == 0:
-        script_store.set_script_state(token, "success")
+        if not details_changed:
+            script_store.set_script_state(token, "success")
         return stdout if isinstance(stdout, str) else stdout.decode("utf-8", errors="replace")
     else:
-        script_store.set_script_state(token, "failure")
+        if not details_changed:
+            script_store.set_script_state(token, "failure")
         return f"Error executing script: return code {returncode}, stderr: {stderr}"


Jazzcort (commented):
@JoseExposito Thanks for the catch! ~I do think we should use the `new_details` to execute the script if it passes the revalidation. @owtaylor @subpop  Are there any reasons that we're not doing that? Also, I feel like we need to update the script detail at the script store with the `new_details` if it passes the revalidation. What do you guys think?~

Just saw the conversation and yes I agree there is no point to update the script store in this situation because we're not retrieving them at any points later. 😁

JoseExposito:
Hey @Jazzcort thanks for your review!

And thanks for confirming that this is the intended behavior :) 

I un-drafted the PR, but I'm not super happy with the repetition of the `if not details_changed`. I considered moving the code to a function, but that's also not a clean solution. I'll appreciate feedback to improve it.

> Just saw the conversation and yes I agree there is no point to update the script store in this situation because we're not retrieving them at any points later. 😁

I just started reading the code today, so I'm not sure about the goal of the ScriptStore class, but I'm going to think out loud to make sure I don't introduce a regression accidentally.

I wonder if it would make sense to change the status from "waiting-approval" to something else. Possibly, a new status to mark it as rejected because of the changes ("rejected-changed" or "rejected-outdated" or similar).

Removing the "waiting-approval" would avoid attempts to re-use the script with an old token in the future. I think that `if not script_details.needs_confirmation` is the only check performed by the run_* functions.

In code, it'd be something like...

```python
if details_changed:
  # Revalidate the script again; this is a convenience for the user to avoid
  # potentially having to double-approve the same script.
  [...]
  script_store.set_script_state(token, "rejected-changed")
else:
  script_store.set_script_state(token, "executing")
```

Jazzcort:
@JoseExposito I agree that adding a bunch of `if not detail_changed` to just update the status that we actually never retrieve is not very appealing to me. 

The Script Store was first created to keep the execution state persistent for `mcp-apps` and then also used to support the `validate_script` workflow to retrieve the script details. From my point of view, `state` is actually never needed while we're not using `mcp-apps` feature because a) we can not execution the script again like `mcp-apps` does (UI triggered execution) b) we don't display `state` to users when not using `mcp-apps` feature and we don't check the `state` in `run_script` or `run_script_with_confirmation` to stop the execution if the script has `executed` as state. However, I also agree that having partially updated state is bad.

So, I think there are two directions we can go

1. Don't even bother updating the state at all.
2. Change the old script to `rejected-changed`, and add the new script to the store and update the state like what we did already.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -498,7 +498,11 @@ async def run_script_with_confirmation(
         host=host,
         readonly=readonly,
     )
-    if new_details != script_details:
+
+    details_changed = new_details != script_details
+    execute_details = new_details if details_changed else script_details
+
+    if details_changed:
```
owtaylor, line 505: The repeated checks on `details_changed` don't bother me much, but if they do, you could do:

```
 if new_details != script_details:
     # revalidate
     execute_details = new_details
     update_script_state = lambda state: pass
 else:
     execute_details = script_details
     update_script_state = lambda state: script_store.set_script_state(state, token)
```

subpop (approved):
LGTM

JoseExposito:
Thanks for merging it @subpop !

Do you think it is worth opening a follow-up MR addressing this?

> Change the old script to rejected-changed, and add the new script to the store and update the state like what we did already.

At the moment, once a script is validated, it is possible to reuse the token to execute as many scripts as required as long as the script changes.

subpop:
> At the moment, once a script is validated, it is possible to reuse the token to execute as many scripts as required as long as the script changes.

Is this a typo? A valid token should definitely not allow any script to run. We should verify with @Jazzcort regarding the mcp-app usage of ScriptStore, but I think the ScriptStore should be a once-and-done cache. Once a script is run, it gets purged from the store so that the model has to revalidate the script if it wants to run it again.

Or, should the store cache validated scripts and allow them to run multiple times without revalidation so long as the script doesn't change?

This ends up being a design decision we should probably discuss.

JoseExposito:
Hey Link,

No, it is not a typo. Let me share a little reproducer:

<details>

<summary>Reproducer</summary>

```bash
#!/bin/bash
set -euo pipefail

ENDPOINT="http://127.0.0.1:8000/mcp"
ACCEPT="application/json, text/event-stream"
sse_json() { grep '^data: ' | head -1 | sed 's/^data: //'; }

SCRIPT_A="echo SCRIPT_A_WAS_EXECUTED"
SCRIPT_B="echo SCRIPT_B_WAS_EXECUTED"
SCRIPT_C="echo SCRIPT_C_WAS_EXECUTED"


### Step 1: Initialize the MCP server ###

INIT_RAW=$(curl -sS -D - \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 1,
    "method": "initialize",
    "params": {
      "protocolVersion": "2024-11-05",
      "capabilities": {},
      "clientInfo": {
        "name": "curl-debug",
        "version": "1.0"
      }
    }
  }')

SESSION_ID=$(echo "$INIT_RAW" | grep -i '^mcp-session-id:' | awk '{print $2}' | tr -d '\r')
echo "Session: $SESSION_ID"


### Step 2: Initialized notification ###

curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "method": "notifications/initialized",
    "params": {}
  }'


### Step 3: validate_script with SCRIPT A (readonly == false -> needs_confirmation) ###

echo -e "\n=== validate_script (Script A) ==="

VALIDATE_JSON=$(curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 2,
    "method": "tools/call",
    "params": {
      "name": "validate_script",
      "arguments": {
        "description": "Print a message",
        "script_type": "bash",
        "script": "'"${SCRIPT_A}"'",
        "readonly": false
      }
    }
  }' | sse_json)

echo "$VALIDATE_JSON" | python3 -m json.tool
TOKEN=$(echo "$VALIDATE_JSON" | python3 -c "
import sys, json
d = json.load(sys.stdin)
print(d['result']['structuredContent']['token'])")
echo "Token: $TOKEN"
echo "needs_confirmation: $(echo "$VALIDATE_JSON" | python3 -c "
import sys, json
d = json.load(sys.stdin)
print(d['result']['structuredContent']['needs_confirmation'])")"

### Step 4: run_script_with_confirmation with SCRIPT B (same token) ###

echo -e "\n=== run_script_with_confirmation (Script B, token from Script A) ==="

CONFIRM_JSON=$(curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
      "name": "run_script_with_confirmation",
      "arguments": {
        "description": "Print a message",
        "script_type": "bash",
        "script": "'"$SCRIPT_B"'",
        "readonly": false,
        "token": "'"$TOKEN"'"
      }
    }
  }' | sse_json)

echo "Script output:"
echo "$CONFIRM_JSON" | python3 -c "
import sys, json
d = json.load(sys.stdin)
print(d['result']['content'][0]['text'])"

### Step 4: run_script_with_confirmation with SCRIPT B (same token) ###

echo -e "\n=== run_script_with_confirmation (Script C, token from Script A) ==="

CONFIRM_JSON=$(curl -sS \
  -H "Content-Type: application/json" \
  -H "Accept: $ACCEPT" \
  -H "mcp-session-id: $SESSION_ID" \
  -X POST "$ENDPOINT" \
  -d \
  '{
    "jsonrpc": "2.0",
    "id": 3,
    "method": "tools/call",
    "params": {
      "name": "run_script_with_confirmation",
      "arguments": {
        "description": "Print a message",
        "script_type": "bash",
        "script": "'"$SCRIPT_C"'",
        "readonly": false,
        "token": "'"$TOKEN"'"
      }
    }
  }' | sse_json)

echo "Script output:"
echo "$CONFIRM_JSON" | python3 -c "
import sys, json
d = json.load(sys.stdin)
print(d['result']['content'][0]['text'])"
```
</details>

If you run the MCP server over http transport and run the reproducer you'd notice that it is possible to run 2 scripts with the same token. 

However, I don't know if this can be a problem or not. Maybe it is OK to ignore this corner case.

> Once a script is run, it gets purged from the store so that the model has to revalidate the script if it wants to run it again.

I have few hours into the code base, so sorry if I'm pointing something stupid/obvious.

I think it is safe to remove the scripts from the store as long as the effect that calls `getExecutionStateFromServer()` (run-script-app.tsx) is not triggered more than once accidentally, for example, because the hook dependencies change.   
At the moment, I don't think the dependencies can  change. I assume `app` can not change and `validatedToolResult` is guarded by `useMemo()`.

If we decide to drop tokens from the store, I'd consider avoiding a possible KeyError in `get_script_details()` just to cover this (unlikely) case.

Jazzcort:
> At the moment, once a script is validated, it is possible to reuse the token to execute as many scripts as required as long as the script changes.

Since we're not checking `state` before execute the script, this is true. But, the more accurate description should be

> At the moment, once a script is validated, it is possible to reuse the token to execute as many scripts **THAT PASS THE GATEKEEPER CHECK** as required as long as the script changes.

Note: `run_script_interactive` doesn't check `state` because the mcp-apps will forbid users from executing the script if it has already been executed at UI level.

From my point of view, I lean toward "each script should only be executed once". But, purging the script from the store will also get rid of the persistence feature which might confuse users in some scenarios while using mcp-apps feature. For example, if users use http transport and execute/reject a few commands and somehow exit the chat session. When users resume the chat session, all the mcp-apps in this session will display 'initial' as state other than 'executed', 'user rejected', or 'gatekeeper rejected' which makes it kind of broken.

I also lean toward "scripts details in the script store should be immutable". Instead of updating the record, we create a new one.

owtaylor:
- there is no harm in allowing repeated execution of a validated script,
and I could imagine it even being useful in some cases if a model wants to
"poll" something.

- We definitely don't want to get rid of persistence in the mcp-apps case.
That will make the Goose Desktop UI seem broken, since it's easy to click
between sessions.

- if people are bothered by storing execution state in cases we don't care,
then I suggest that we have a "validated" state coming out of
validate_script and just leave it that way for run_script and
run_script_with_confirmation. (I don't see any reason to add an extra entry
for the mismatch case.)

- in general the script score being in memory, scoped to the process, and
stored indefinitely is a bit weird, something like valkey+expiration might
make more sense, but unless we need to start supporting clustered operation
it's not causing us much pain.

JoseExposito:
Thanks a lot for clarification! It looks like we can ignore this case then :+1:

PR #422, opened 2026-04-20
Call the generate_tool_docs script during doc workflow

subpop:
The docs.yml workflow was not calling the `generate_tool_docs.py` script, so the documentation workflow was failing.


Jazzcort (approved):
LGTM! 😁

PR #423, opened 2026-04-20
Move annotations parameter to @mcp.tool decorator

subpop:
The read only hint annotation for the `validate_script` tool was mistakenly put in the tool argument parameters instead of the `@mcp.tool` decorator parameters.


Jazzcort (approved):
LGTM! 😁

PR #426, opened 2026-04-21
feat(ci): add evaluation workflow for gatekeeper models

Jazzcort:
- Introduce GitLab CI inputs to toggle between 'mirror' and 'eval' workflows.
- Add 'setup' stage to install uv, nvm, and build the MCP app.
- Add 'eval' stage with jobs for target gatekeeper models.
- Implement 'post-eval' stage to aggregate results and prepare an HTML report.
- Configure Red Hat internal certificate handling for corp model endpoints.

### DIFF
--- a/.gitlab-ci.yml
+++ b/.gitlab-ci.yml
@@ -5,27 +5,53 @@
 #   GITLAB_TOKEN - Personal/Project Access Token with 'write_repository' scope
 #                  Configure at: Settings > CI/CD > Variables (Protected + Masked)
 
-workflow:
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "web"
-      when: always
-    - when: never
-
 variables:
+  WORKFLOW:
+    description: "Select workflow for manual runs."
+    value: "mirror"
+    options:
+      - "mirror"
+      - "gatekeeper_eval"
   GITHUB_PR_NUMBER:
     description: "GitHub PR number to mirror"
   GITHUB_COMMIT_SHA:
     description: "GitHub commit SHA to fetch"
 
+# Define all possible stages globally so child files can just attach to them
 stages:
+  - setup
+  - eval
+  - post-eval
   - mirror
 
-mirror-github-pr:
-  stage: mirror
-  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/python:3.14
+default:
   tags:
     - rhel-lightspeed
-  script:
-    - python3 scripts/mirror_github_pr.py
+
+# The gitlab pipelines/new web UI does not work correctly if there are no jobs
+placeholder_job:
+  stage: setup
+  script: ["true"]
   rules:
-    - if: $GITHUB_PR_NUMBER && $GITHUB_COMMIT_SHA
+    - if: $WORKFLOW == "impossible_value"
+
+# ==========================================
+# ROUTER: Dynamically include child workflows
+# ==========================================
+include:
+  # 1. MIRRORING
+  - local: '.gitlab/ci/mirror.yml'
+    rules:
+      - if: $WORKFLOW == "mirror" || ($GITHUB_PR_NUMBER && $GITHUB_COMMIT_SHA)
+
+  # 2. GATEKEEPER EVAL
+  - local: '.gitlab/ci/eval-gatekeeper.yml'
+    rules:
+      # Trigger A: Manual selection from UI
+      - if: $WORKFLOW == "gatekeeper_eval"
+      # Trigger B: Automated MR or Push, BUT only if specific files changed
+      - if: '!$WORKFLOW && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")'
+        changes:
+          - "eval/gatekeeper/**/*"
+          - "src/linux_mcp_server/gatekeeper/**/*"
+
--- a/.gitlab/ci/eval-gatekeeper.yml
+++ b/.gitlab/ci/eval-gatekeeper.yml
@@ -1,0 +1,98 @@
+# ==========================================
+# EVAL WORKFLOW - SETUP
+# ==========================================
+setup-uv-and-mcp-apps:
+  stage: setup
+  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/python:3.14
+  variables:
+    HOME: "$CI_PROJECT_DIR"
+  script:
+    # Install uv
+    - export PYTHONUSERBASE=$CI_PROJECT_DIR/.local
+    - pip install --user uv
+
+    # Install nvm to set up node and npm
+    - curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh | bash
+    - \. "$HOME/.nvm/nvm.sh"
+    - nvm install 24
+
+    # Generate the bundled html file
+    - cd mcp-app && npm install && npm run build:prod
+  artifacts:
+    paths:
+      - .local/
+      - mcp-app/dist/
+    expire_in: 1 hour
+
+# ==========================================
+# EVAL WORKFLOW - TEMPLATES
+# ==========================================
+# Base template for all eval jobs
+.eval-base:
+  stage: eval
+  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/python:3.14
+  needs:
+    - job: setup-uv-and-mcp-apps
+      artifacts: true
+  before_script:
+    - export PATH="$CI_PROJECT_DIR/.local/bin:$PATH"
+    - uv sync --locked
+    - mkdir -p "$CI_PROJECT_DIR/data"
+    - export SSL_CERT_FILE="$CI_PROJECT_DIR/cert.pem"
+    - cp /etc/ssl/certs/ca-certificates.crt $SSL_CERT_FILE
+    - curl https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem >> $SSL_CERT_FILE
+  artifacts:
+    paths: 
+      - data/
+    expire_in: 1 hour
+
+# ==========================================
+# EVAL WORKFLOW - JOBS
+# ==========================================
+gatekeeper-eval-claude-4.6-opus: 
+  extends: .eval-base
+  variables:
+    MODEL_NAME: "claude-4.6-opus"
+  script:
+    - export LINUX_MCP_GATEKEEPER_MODEL="openrouter/anthropic/$MODEL_NAME"
+    - uv run eval/gatekeeper/run-eval.py --all -f json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
+  rules:
+    - if: $OPENROUTER_API_KEY
+
+gatekeeper-eval-models-corp-gemini:
+  extends: .eval-base
+  variables:
+    MODEL_NAME: "gemini-3.1-pro-preview"
+  script:
+    - export OPENAI_API_KEY="$MODELS_CORP_GEMINI_API_KEY"
+    - export MODEL="$MODEL_NAME"
+    - ./eval/gatekeeper/run-eval-models-corp.sh --all -f json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
+  rules:
+    - if: $MODELS_CORP_GEMINI_API_KEY
+
+gatekeeper-eval-models-corp-gpt-oss-20b:
+  extends: .eval-base
+  variables:
+    MODEL_NAME: "gpt-oss-20b"
+  script:
+    - export OPENAI_API_KEY="$MODELS_CORP_GPT_OSS_20B_API_KEY"
+    - export MODEL="openai/$MODEL_NAME"
+    - ./eval/gatekeeper/run-eval-models-corp.sh --all -f json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
+  rules:
+    - if: $MODELS_CORP_GPT_OSS_20B_API_KEY
+
+# ==========================================
+# EVAL WORKFLOW - POST-EVAL
+# ==========================================
+post-eval-result:
+  stage: post-eval
+  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/bash
+  script:
+    - cp eval/gatekeeper/index.html "$CI_PROJECT_DIR/index.html"
+
+      # Create the manifest file for the rendering app to correctly gather the data
+    - ls data/*.json | xargs -n1 basename > data/manifest.txt
+  artifacts:
+    paths:
+      - data/
+      - index.html
--- a/.gitlab/ci/mirror.yml
+++ b/.gitlab/ci/mirror.yml
@@ -1,0 +1,5 @@
+mirror-github-pr:
+  stage: mirror
+  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/python:3.14
+  script:
+    - python3 scripts/mirror_github_pr.py


.gitlab-ci.yml
```diff
@@ -5,6 +5,15 @@
 #   GITLAB_TOKEN - Personal/Project Access Token with 'write_repository' scope
 #                  Configure at: Settings > CI/CD > Variables (Protected + Masked)
 
+spec:
+  inputs:
+    workflow:
+      options: ["mirror", "eval"]
```
Jazzcort: Because we're going to save all the API keys at the CI/CD variables, they're always available in any workflows. We need this value to bypass `eval` workflow when we're doing `mirror`

.gitlab-ci.yml
```diff
@@ -29,3 +41,137 @@ mirror-github-pr:
...
+  artifacts:
+    paths:
+      - .local/
+      - mcp-app/dist/
+    expire_in: 1 hour
+  rules:
+    - if: '"$[[ inputs.workflow ]]" == "eval"'
+
+
+gatekeeper-eval-claude-4.6-opus:
```
Jazzcort: This one is just an example of running eval against external platforms. We can tweak it to test other models that we don't want to run from the internal platform or are not hosted there. i.e., gpt-oss-120b, claude.

Jazzcort (commented):
The result looks good here -> https://jazzcort.gitlab.io/-/linux-mcp-server/-/jobs/14025853739/artifacts/index.html 😁

owtaylor:
I threw this at Gemini to ask for some advice about structuring this, and it came up with what seem like some good ideas. (I asked it including a hypothetical "end-to-end eval" in addition to the mirroring and gatekeeper eval, and said that we'd want to run the gatekeeper eval or end-to-end eval on MR's pushes depending on what changed.)

It suggested that .gitlab-ci.yml be reduced to a "router" and the actual work moved to included files, something like:

```
├── .gitlab-ci.yml                # The Router (Inputs, Variables, Includes)
└── .gitlab/ci/
    ├── mirror.yml                # Only the Mirror jobs
    ├── eval-gatekeeper.yml       # Only the Gatekeeper jobs
    └── eval-e2e.yml              # Only the E2E jobs
```

Where the router `.gitlab-ci.yml` looks something like:

```yaml
spec:
  inputs:
    workflow:
      # Added "auto" as the default so automated push/MR events don't get forced into "mirror"
      options: ["auto", "mirror", "gatekeeper_eval", "e2e_eval"] 
      default: "auto"
      description: "Select workflow for manual runs. Leave 'auto' for pushes/MRs."

---

variables:
  # Left here so they can be prefilled via URL: /pipelines/new?var[GITHUB_PR_NUMBER]=123
  GITHUB_PR_NUMBER:
    description: "GitHub PR number to mirror"
  GITHUB_COMMIT_SHA:
    description: "GitHub commit SHA to fetch"

# Define all possible stages globally so child files can just attach to them
stages:
  - setup
  - eval
  - post-eval
  - e2e
  - mirror

default:
  tags:
    - rhel-lightspeed

# ==========================================
# ROUTER: Dynamically include child workflows
# ==========================================
include:
  # 1. MIRRORING
  - local: '.gitlab/ci/mirror.yml'
    rules:
      - if: '$[[ inputs.workflow ]] == "mirror"'

  # 2. GATEKEEPER EVAL
  - local: '.gitlab/ci/eval-gatekeeper.yml'
    rules:
      # Trigger A: Manual selection from UI
      - if: '$[[ inputs.workflow ]] == "gatekeeper_eval"'
      # Trigger B: Automated MR or Push, BUT only if specific files changed
      - if: '$[[ inputs.workflow ]] == "auto" && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")'
        changes:
          - "mcp-app/**/*"
          - "eval/gatekeeper/**/*"
          - "scripts/eval_helpers/**/*" # Add whatever paths make sense

  # 3. END-TO-END EVAL
  - local: '.gitlab/ci/eval-e2e.yml'
    rules:
      # Trigger A: Manual selection from UI
      - if: '$[[ inputs.workflow ]] == "e2e_eval"'
      # Trigger B: Automated for ANY MR or Push
      - if: '$[[ inputs.workflow ]] == "auto" && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")'
```

(I would make the mirror step run as well for the `auto` workflow if the relevant variables are set)

For the gatekeeper eval, it suggested using hidden base jobs `.eval-base` and `.eval-corp-base` and the `extends:` keyword to avoid repetition. It also suggested moving setup steps from `script:` to `before-script:`.

```yaml
# ==========================================
# EVAL WORKFLOW - SETUP
# ==========================================
setup-uv-and-mcp-apps:
  stage: setup
  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/python:3.14
  variables:
    HOME: "$CI_PROJECT_DIR"
  script:
    # Install uv
    - export PYTHONUSERBASE=$CI_PROJECT_DIR/.local
    - pip install --user uv

    # Install nvm to set up node and npm
    - curl -o- https://raw.githubusercontent.com/nvm-sh/nvm/v0.40.4/install.sh | bash
    - \. "$HOME/.nvm/nvm.sh"
    - nvm install 24

    # Generate the bundled html file
    - cd mcp-app && npm install && npm run build:prod
  artifacts:
    paths:
      - .local/
      - mcp-app/dist/
    expire_in: 1 hour
  rules:
    - if: '"$[[ inputs.workflow ]]" == "eval"'

# ==========================================
# EVAL WORKFLOW - TEMPLATES
# ==========================================
# Base template for all eval jobs
.eval-base:
  stage: eval
  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/python:3.14
  needs:
    - job: setup-uv-and-mcp-apps
      artifacts: true
  before_script:
    - export PATH="$CI_PROJECT_DIR/.local/bin:$PATH"
    - uv sync --locked
    - mkdir -p "$CI_PROJECT_DIR/data"
  artifacts:
    paths: 
      - data/
    expire_in: 1 hour

# Template specific to corporate models needing the Red Hat cert
.eval-corp-base:
  extends: .eval-base
  before_script:
    - !reference [.eval-base, before_script]
    - curl -o "$CI_PROJECT_DIR/cert.pem" https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem
    - export SSL_CERT_FILE="$CI_PROJECT_DIR/cert.pem"

# ==========================================
# EVAL WORKFLOW - JOBS
# ==========================================
gatekeeper-eval-claude-4.6-opus: 
  extends: .eval-base
  variables:
    MODEL_NAME: "claude-4.6-opus"
  script:
    - export LINUX_MCP_GATEKEEPER_MODEL="openrouter/anthropic/$MODEL_NAME"
    - uv run eval/gatekeeper/run-eval.py --all --json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
  rules:
    - if: $OPENROUTER_API_KEY && '"$[[ inputs.workflow ]]" == "eval"'

gatekeeper-eval-models-corp-gemini:
  extends: .eval-corp-base
  variables:
    MODEL_NAME: "gemini-3.1-pro-preview"
  script:
    - export OPENAI_API_KEY="$MODELS_CORP_GEMINI_API_KEY"
    - export MODEL="$MODEL_NAME"
    - ./eval/gatekeeper/run-eval-models-corp.sh --all --json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
  rules:
    - if: $MODELS_CORP_GEMINI_API_KEY && '"$[[ inputs.workflow ]]" == "eval"'

gatekeeper-eval-models-corp-gpt-oss-20b:
  extends: .eval-corp-base
  variables:
    MODEL_NAME: "gpt-oss-20b"
  script:
    - export OPENAI_API_KEY="$MODELS_CORP_GPT_OSS_20B_API_KEY"
    - export MODEL="openai/$MODEL_NAME"
    - ./eval/gatekeeper/run-eval-models-corp.sh --all --json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
  rules:
    - if: $MODELS_CORP_GPT_OSS_20B_API_KEY && '"$[[ inputs.workflow ]]" == "eval"'

# ==========================================
# EVAL WORKFLOW - POST-EVAL
# ==========================================
post-eval-result:
  stage: post-eval
  image: images.paas.redhat.com/it-cloud-ocp-proxy-lib/bash
  script:
    - cp eval/gatekeeper/index.html "$CI_PROJECT_DIR/index.html"
    - ls data/*.json | xargs -n1 basename > data/manifest.txt
  artifacts:
    paths:
      - data/
      - index.html
  rules:
    - if: '"$[[ inputs.workflow ]]" == "eval"'
```

.gitlab/ci/eval-gatekeeper.yml
```diff
@@ -0,0 +1,103 @@
...
+# ==========================================
+gatekeeper-eval-claude-4.6-opus: 
+  extends: .eval-base
+  variables:
+    MODEL_NAME: "claude-4.6-opus"
+  script:
+    - export LINUX_MCP_GATEKEEPER_MODEL="openrouter/anthropic/$MODEL_NAME"
+    - uv run eval/gatekeeper/run-eval.py --all -f json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
+  rules:
+    - if: $OPENROUTER_API_KEY
```
Jazzcort, line 60: I removed all `'"$[[ inputs.workflow ]]" == "eval"'` here since a) `$[[inputs.workflow]]` is not accessible here. b) this file only gets include when `'"$[[ inputs.workflow ]]" == "eval"'` so we don't have to check it again

.gitlab/ci/eval-gatekeeper.yml
```diff
@@ -0,0 +1,103 @@
...
+
+gatekeeper-eval-models-corp-gemini:
+  extends: .eval-corp-base
+  variables:
+    MODEL_NAME: "gemini-3.1-pro-preview"
+  script:
+    - export OPENAI_API_KEY="$MODELS_CORP_GEMINI_API_KEY"
+    - export MODEL="$MODEL_NAME"
+    - ./eval/gatekeeper/run-eval-models-corp.sh --all -f json --output-all -o "$CI_PROJECT_DIR/data/$MODEL_NAME.json"
```
Jazzcort, line 69: `--json` -> `-f json`

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
...
+  inputs:
+    workflow:
+      # Added "auto" as the default so automated push/MR events don't get forced into "mirror"
+      options: ["auto", "mirror", "gatekeeper_eval", "e2e_eval"] 
+      default: "auto"
+      description: "Select workflow for manual runs. Leave 'auto' for pushes/MRs."
+---
 
 variables:
+  # Left here so they can be prefilled via URL: /pipelines/new?var[GITHUB_PR_NUMBER]=123
```
Jazzcort: So these variables won't show up in the "Run new pipeline" web page since using `spec: inputs` will make the web UI omit to show variables. But, the URL trick still works - `/pipeline/new?var[GITHUB_PR_NUMNER=12]`. Also, I found out that the URL trick can just provide any variables we want not just the variable we listed in the ci config. Therefore, having these variables here becomes more like a reference since they won't show up when we want to manually trigger mirror and those description won't appear either.

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
...
+
+  # 2. GATEKEEPER EVAL
+  - local: '.gitlab/ci/eval-gatekeeper.yml'
+    rules:
+      # Trigger A: Manual selection from UI
+      - if: '"$[[ inputs.workflow ]]" == "gatekeeper_eval"'
+      # Trigger B: Automated MR or Push, BUT only if specific files changed
+      - if: '"$[[ inputs.workflow ]]" == "auto" && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")'
+        changes:
+          - "eval/gatekeeper/**/*"
```
Jazzcort, line 55: I feel like we should only run gatekeeper_eval whenever we have prompt improvement, improve the workflow of the GCE, change eval logic, or add/delete test cases. @owtaylor what do you think?

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
...
+          - "src/linux_mcp_server/tools/run_script.py"
+          - "src/linux_mcp_server/gatekeeper/**/*"
+          - "src/linux_mcp_server/server.py"
+
+  # # 3. END-TO-END EVAL
```
Jazzcort: I have to  comment this out to avoid errors when gitlab loads this ci config since we don't have `.gitlab/ci/eval-e2e.yml` now.

Jazzcort (commented):
Just some comments that gives more info of my modification. 😁 Tested this new config and it works -> https://jazzcort.gitlab.io/-/linux-mcp-server/-/jobs/14146170109/artifacts/index.html

.gitlab-ci.yml
owtaylor: Ugh. Having magic URL parameters that affect what happens when you submit the form without any visibility in the UI s not nice.

Maybe we should just make the workflow also a pipeline variable until we can prefill the inputs ([gitlab RFE](https://gitlab.com/gitlab-org/gitlab/-/work_items/553383))?

The only other thing I can think of is that we have a *separate* gitlab.cee project `gitlab.cee.redhat.com/rhel-lightspeed/mcp/pr-mrror/` or something that has a simple pipeline that takes the CI variables and triggers the real workflow with the appropriate inputs (https://docs.gitlab.com/ci/triggers/#pass-pipeline-inputs-in-the-api-call)

And then we could disable pipeline variables entirely for the main project for added security. 

That's sort of nifty, but probably sticking to pipeline variables is better for now. What do you think?

Jazzcort: I feel like I'm leaning toward having workflow as a pipeline variable for now. 😁 The benefits of using inputs are having typed variables, selections, and being able to dynamically inject these input into the scripts and we don't use most of them. So, there is no strong reason that we have to use inputs other than that Gitlab recommend inputs over variables.

>The only other thing I can think of is that we have a separate gitlab.cee project gitlab.cee.redhat.com/rhel-lightspeed/mcp/pr-mrror/ or something that has a simple pipeline that takes the CI variables and triggers the real workflow with the appropriate inputs (https://docs.gitlab.com/ci/triggers/#pass-pipeline-inputs-in-the-api-call)

So we use a shell to call the trigger API and in the API call we specify the inputs? But, how do we get the source info to specify these inputs? With CI variables? I'm just a bit confused on this 😆 It might just make sense when I can see the actual CI config file. 🤔

owtaylor: > So we use a shell to call the trigger API and in the API call we specify the inputs? But, how do we get the source info to specify these inputs? With CI variables? I'm just a bit confused on this 😆 It might just make sense when I can see the actual CI config file. 🤔

The idea is that the additional project would be set up to trigger mirroring from CI variables, just like the gitlab.cee.redhat.com/rhel-lightspeed/mcp/linux-mcp-server is now, but instead of having the mirror script, etc, it would just be glue that uses curl to trigger the *real* workflow in rhel-lightspeed/mcp/linux-mcp-server passing in the correct inputs.

Somewhat Rube Goldberg - making workflow a pipeline variable should be fine.

Jazzcort: >Ugh. Having magic URL parameters that affect what happens when you submit the form without any visibility in the UI s not nice.

Actually, if we use the magic URL to get to the "Run new pipeline" page, those variable will show up there without description. 😁 It's just if I create a new pipeline by clicking "New pipeline" in the pipeline web page, those variable won't show up with empty value like before. Sorry for the confusion, if there were any. 😁

.gitlab/ci/eval-gatekeeper.yml
```diff
@@ -0,0 +1,103 @@
...
+      - data/
+    expire_in: 1 hour
+
+# Template specific to corporate models needing the Red Hat cert
+.eval-corp-base:
+  extends: .eval-base
+  before_script:
+    - !reference [.eval-base, before_script]
+    - curl -o "$CI_PROJECT_DIR/cert.pem" https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem
+    - export SSL_CERT_FILE="$CI_PROJECT_DIR/cert.pem"
```
owtaylor: Replacing the cert database entirely is fragile in my experience - suddenly you need to access some other host that uses a standard cert. Maybe:

```suggestion
    - export SSL_CERT_FILE="$CI_PROJECT_DIR/cert.pem"
    - cp /etc/ssl/certs/ca-certificates.crt $SSL_CERT_FILE
    - curl https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem >> $SSL_CERT_FILE
```

That would allow getting rid of .eval-corp-base - we can just always do that.

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
...
+    rules:
+      # Trigger A: Manual selection from UI
+      - if: '"$[[ inputs.workflow ]]" == "gatekeeper_eval"'
+      # Trigger B: Automated MR or Push, BUT only if specific files changed
+      - if: '"$[[ inputs.workflow ]]" == "auto" && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")'
+        changes:
+          - "eval/gatekeeper/**/*"
+          - "src/linux_mcp_server/tools/run_script.py"
+          - "src/linux_mcp_server/gatekeeper/**/*"
+          - "src/linux_mcp_server/server.py"
```
owtaylor: I don't think this needs to run for changes to `run_script.py` or `server.py` ... we're *just* testing the gatekeeper here.

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
...
-workflow:
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "web"
-      when: always
-    - when: never
+spec:
+  inputs:
+    workflow:
+      # Added "auto" as the default so automated push/MR events don't get forced into "mirror"
+      options: ["auto", "mirror", "gatekeeper_eval", "e2e_eval"]
```
owtaylor: I would just remove any references to e2e_eval - I included it when prompting Gemini to make sure that the structure was good for more than one eval stage ("gatekeeper_eval", not "eval", etc.)

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
 #   GITLAB_TOKEN - Personal/Project Access Token with 'write_repository' scope
 #                  Configure at: Settings > CI/CD > Variables (Protected + Masked)
 
-workflow:
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "web"
-      when: always
-    - when: never
+spec:
+  inputs:
```
owtaylor: As discussed elsewhere, my preference would be to avoid `inputs` for now, and do everything as variables.

.gitlab-ci.yml
```diff
@@ -5,27 +5,61 @@
...
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "web"
-      when: always
-    - when: never
-
 variables:
+  WORKFLOW:
+    description: "Select workflow for manual runs. Leave 'auto' for pushes/MRs."
+    value: "mirror"
+    options:
```
Jazzcort, line 12: I removed "auto" because the prefill only works for `spec: inputs` not for `variables`. When the pipeline is triggered by MR or push, WORKFLOW is actually unset. The preset value here is only for manually creating pipeline through web UI.

Also since "mirror" is not the default, we don't have to do another patch for the magic URL to include `var[WORKFLOW]=mirror`. 😁

.gitlab-ci.yml
```diff
@@ -5,27 +5,62 @@
...
-    - if: $GITHUB_PR_NUMBER && $GITHUB_COMMIT_SHA
+
+# ==========================================
+# ROUTER: Dynamically include child workflows
+# ==========================================
+include:
+  # 1. MIRRORING
+  - local: '.gitlab/ci/mirror.yml'
+    rules:
+      - if: '"$[[ inputs.workflow ]]" == "mirror"'
```
owtaylor: This will force the user when triggering to change the drop down / variable from "auto" to "mirror" right?

 I'd expect to run this if one of two is true:
 A) inputs.workflow is mirror (error if  GITHUB_PR_NUMBER and GITHUB_COMMIT_SHA are not set)
 B) Workflow is "auto" and GITHUB_PR_NUMBER or GITHUB_COMMIT_SHA is set

owtaylor (changes requested):
Generally looking good

.gitlab-ci.yml
```diff
@@ -5,27 +5,61 @@
...
+# ROUTER: Dynamically include child workflows
+# ==========================================
+include:
+  # 1. GATEKEEPER EVAL
+  - local: '.gitlab/ci/eval-gatekeeper.yml'
+    rules:
+      # Trigger A: Manual selection from UI
+      - if: $WORKFLOW == "gatekeeper_eval"
+      # Trigger B: Automated MR or Push, BUT only if specific files changed
+      - if: $CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push"
```
Jazzcort: I tried `$WORKFLOW == "auto" && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")` and it would make the MR and push trigger never work since `WORKFLOW` is unset.

.gitlab-ci.yml
```diff
@@ -5,27 +5,61 @@
...
+  - eval
+  - post-eval
+  - e2e
   - mirror
 
+default:
+  tags:
+    - rhel-lightspeed
+
 mirror-github-pr:
```
Jazzcort: I finally found out that the reason that those variables were not displayed in the web UI is not because we used `spec: inputs`, it's actually because we left no jobs here after the dynamic `include` which caused some errors on gitlab. The `include` section is actually something gitlab will replace with the content of the separate yml file and create a new `gitlab-ci.yml` on the run, so when there are no `include` matching their conditions, we end up with a `gitlab-ci.yml` with no job to run which raises some error for the web rendering I guess.

The easy fix here is to guarantee at least one job will be in `gitlab-ci.yml`. Since `mirror` only have one simple job, instead of extract it to `.gitlab/ci/mirror.yml`, we can just put it here. 😁

Jazzcort (commented):
@owtaylor I hope this is the final version. 😆 Gitlab really needs to improve their implementation of configuring pipeline.

.gitlab-ci.yml
Jazzcort: You're right! I mix up the prompts for "tool calls" and prompts for "gatekeeper". Will change!

Jazzcort:
@owtaylor It's ready for reviewing. 😁

.gitlab-ci.yml
Jazzcort: Some updates on this. `variable` preset actually works for jobs that are not dynamically injected by `include`. Therefore, `- if: $WORKFLOW == "mirror" || ($GITHUB_PR_NUMBER && $GITHUB_COMMIT_SHA)` makes our CI quite noisy since it will get trigger on every push or MR.

.gitlab-ci.yml
owtaylor: Yeah, if we want unset WORKFLOW to mean auto, we'd want `!$WORKFLOW && ...` right?

.gitlab-ci.yml
owtaylor: That seems right to me.

.gitlab-ci.yml
```diff
@@ -5,27 +5,49 @@
...
 
-workflow:
-  rules:
-    - if: $CI_PIPELINE_SOURCE == "web"
-      when: always
-    - when: never
-
 variables:
+  WORKFLOW:
+    description: "Select workflow for manual runs. Leave 'auto' for pushes/MRs."
```
owtaylor: What does "leave 'auto'" mean here? When would be auto? Does the UI allow setting anything other than mirror/gatekeeper_eval ?

.gitlab-ci.yml
owtaylor: I'd rather have mirror in a separate file, and then have some sort of placeholder job:

```yaml
# The gitlab pipelines/new web UI does not work correctly if there are no jobs
placeholder_job:
   stage: setup
   script: ["true"]
   rules:
       - if: $WORKFLOW = "impossible_value"
```
(Maybe this could be further simplified to rules: [])

.gitlab/ci/eval-gatekeeper.yml
```diff
@@ -0,0 +1,104 @@
...
+    expire_in: 1 hour
+
+# Template specific to corporate models needing the Red Hat cert
+.eval-corp-base:
+  extends: .eval-base
+  before_script:
+    - !reference [.eval-base, before_script]
+    - export SSL_CERT_FILE="$CI_PROJECT_DIR/cert.pem"
+    - cp /etc/ssl/certs/ca-certificates.crt $SSL_CERT_FILE
+    - curl https://certs.corp.redhat.com/certs/2022-IT-Root-CA.pem >> $SSL_CERT_FILE
```
owtaylor: This can be just part of `.eval-base` - we don't need to do something different for the models.corp cases

owtaylor (changes requested):
Good work tracking down what is going on with the variable display in the UI. I don't like magically leaving the mirror script in the same file ... seems like a trap for our future selves.

.gitlab-ci.yml
Jazzcort: I missed it! Thanks for the catch!

From the web UI, WORKFLOW is a dropdown menu which only have "mirror" and "gatekeeper_eval" as options. But, if we manually delete this variable and add a new one with same variable name (WORKFLOW), then we can put whatever we want for its value. Same for the magic URL trick, we can assign WORKFLOW to any values we want.

To conclude, setting predefined options for `variables`does not restrict us from providing other values that's not in the predefined set.

.gitlab/ci/eval-gatekeeper.yml
Jazzcort: Got it! I'll remove this hidden job and move this part to `.eval-base`

.gitlab-ci.yml
```diff
@@ -5,27 +5,53 @@
...
-  script:
-    - python3 scripts/mirror_github_pr.py
+
+# The gitlab pipelines/new web UI does not work correctly if there are no jobs
+placeholder_job:
+  stage: setup
+  script: ["true"]
   rules:
-    - if: $GITHUB_PR_NUMBER && $GITHUB_COMMIT_SHA
+    - if: $WORKFLOW == "impossible_value"
```
Jazzcort, line 36: Gitlab does not allow the rules config to be empty. 😢

.gitlab-ci.yml
```diff
@@ -5,27 +5,53 @@
...
+    rules:
+      - if: $WORKFLOW == "mirror" || ($GITHUB_PR_NUMBER && $GITHUB_COMMIT_SHA)
+
+  # 2. GATEKEEPER EVAL
+  - local: '.gitlab/ci/eval-gatekeeper.yml'
+    rules:
+      # Trigger A: Manual selection from UI
+      - if: $WORKFLOW == "gatekeeper_eval"
+      # Trigger B: Automated MR or Push, BUT only if specific files changed
+      - if: $WORKFLOW == null && ($CI_PIPELINE_SOURCE == "merge_request_event" || $CI_PIPELINE_SOURCE == "push")
```
Jazzcort: GitLab CI's expression parser does not support the unary ! (NOT) operator.

Jazzcort (commented):
@owtaylor Let me know if this looks good to you! 😁

owtaylor (approved):
Looks great. Let's land this as is, then sort out what models we want to actually be testing and what API keys we need to provide.

Jazzcort:
Thanks for reviewing! 😁

PR #430, opened 2026-04-23
feat: implement session-based tracking for malicious behavior

Jazzcort:
Adds `BehaviorRecord` and `BehaviorRecordManager` to monitor gatekeeper verdicts per session. If a session accumulates 10 total or 3 consecutive malicious actions, it is permanently flagged as malicious. Once flagged, all subsequent scripts executed in that session will be forced to require user confirmation, regardless of the script's default requirements.

### DIFF
--- a/src/linux_mcp_server/tools/run_script.py
+++ b/src/linux_mcp_server/tools/run_script.py
@@ -154,7 +154,83 @@ class ScriptStore:
 
 
 script_store = ScriptStore()
+MAX_CONSECUTIVE_MALICIOUS_ACTIONS = 3
+MAX_TOTAL_MALICIOUS_ACTIONS = 4
 
+
+class BehaviorRecord:
+    """
+    Tracks gatekeeper verdicts for a single session to detect malicious behavior.
+
+    Each MALICIOUS verdict sets a temporary warning that forces the next script to
+    require human confirmation. The temporary warning is cleared once a human approves
+    execution. Non-malicious verdicts reset the consecutive counter but not the total.
+
+    A session is permanently flagged if it accumulates MAX_TOTAL_MALICIOUS_ACTIONS
+    total malicious actions or MAX_CONSECUTIVE_MALICIOUS_ACTIONS consecutive malicious
+    actions. Once permanently flagged, all subsequent scripts require confirmation and
+    the flag cannot be cleared.
+    """
+
+    def __init__(self):
+        self._consecutive_malicious_action_counts = 0
+        self._total_malicious_action_counts = 0
+        self._has_temporary_warning = False
+        self._is_malicious = False
+
+    def add_record(self, status: GatekeeperStatus):
+        """
+        Record a gatekeeper verdict and update tracking state.
+
+        A MALICIOUS verdict increments both the consecutive and total counters and
+        sets a temporary warning. Any other verdict resets the consecutive counter.
+        If either threshold is met, the session is permanently flagged. No-ops if
+        the session is already permanently flagged.
+        """
+        # No need to update the record if it's already considered security compromised
+        if self._is_malicious:
+            return
+
+        if status == GatekeeperStatus.MALICIOUS:
+            self._consecutive_malicious_action_counts += 1
+            self._total_malicious_action_counts += 1
+            self._has_temporary_warning = True
+        else:
+            self._consecutive_malicious_action_counts = 0
+
+        # Check if the record matches the conditions of being considered as malicious
+        if self._total_malicious_action_counts >= MAX_TOTAL_MALICIOUS_ACTIONS:
+            self._is_malicious = True
+            return
+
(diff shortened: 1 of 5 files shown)


src/linux_mcp_server/tools/run_script.py
```diff
@@ -152,8 +153,65 @@ def set_script_state(self, id: str, new_state: ExecutionState):
...
 script_store = ScriptStore()
+MAX_CONSECUTIVE_MALICIOUS_ACTIONS = 3
 
 
+class BehaviorRecord:
+    """
+    Tracks gatekeeper verdicts for a single session to detect malicious behavior.
+
+    A session is flagged as malicious if it accumulates 10+ total malicious actions
+    or MAX_CONSECUTIVE_MALICIOUS_ACTIONS consecutive malicious actions.
```
owtaylor: A little odd to have one a constant and the other not

src/linux_mcp_server/tools/run_script.py
```diff
@@ -152,8 +153,65 @@ def set_script_state(self, id: str, new_state: ExecutionState):
...
+    Once flagged, the session is permanently marked and all subsequent scripts require confirmation.
+    """
+
+    def __init__(self):
+        self._recent_record: deque[GatekeeperStatus] = deque(maxlen=MAX_CONSECUTIVE_MALICIOUS_ACTIONS)
```
owtaylor: deque seems unnecessary - can you just have _consecutive_malicious_action_acounts and reset it to zero if you get some other status.

owtaylor (changes requested):
Hmm, not sure this is going to be effective:

 * A malicious model can easily evade a "consecutive malicious scripts" checks - also a malicious script for readonly that modifies the system or installs things from the internet is currently often tagged as MODIFIES_SYSTEM or POLICY, so a malicious 
 * 10 tries is a lot, by then the game might be up.

Maybe:

 * Any MALICIOUS script triggers confirmation of the *next* run_script (but not the whole session)
 * For the mcp-apps case, we display a boxed warning in that confirmation:

      Suspicious activity detected - your chat client may be under attack. Please examine previous tool calls in detail, and if you have any doubts, do not approve this command and terminate this chat session.

There's not much we can do about the non-mcp-apps case.

src/linux_mcp_server/tools/run_script.py
Jazzcort: You're right! It can be simplified 😁

src/linux_mcp_server/tools/run_script.py
Jazzcort: Got it! I'll have a MAX_TOTAL_MALICIOUS_ACTIONS for that. What would be the suitable count for it? 4 maybe? 🤔

src/linux_mcp_server/tools/run_script.py
```diff
@@ -154,8 +154,77 @@ def set_script_state(self, id: str, new_state: ExecutionState):
...
+        if self._consecutive_malicious_action_counts >= MAX_CONSECUTIVE_MALICIOUS_ACTIONS:
+            self._is_security_compromised = True
+
+    @property
+    def is_security_compromised(self) -> bool:
+        return self._is_security_compromised
+
+    @property
+    def is_previous_action_malicious(self) -> bool:
+        return self._previous_action_status == GatekeeperStatus.MALICIOUS
```
owtaylor: I think this should not be cleared except by an approval - that is:

```
validate_script => MALICIOUS - sets flag
validate_script => OK => flag stays set
execute_script => flag is cleared
```

src/linux_mcp_server/tools/run_script.py
```diff
@@ -154,8 +154,77 @@ def set_script_state(self, id: str, new_state: ExecutionState):
...
+        if self._total_malicious_action_counts >= MAX_TOTAL_MALICIOUS_ACTIONS:
+            self._is_security_compromised = True
+            return
+
+        if self._consecutive_malicious_action_counts >= MAX_CONSECUTIVE_MALICIOUS_ACTIONS:
+            self._is_security_compromised = True
+
+    @property
+    def is_security_compromised(self) -> bool:
+        return self._is_security_compromised
```
owtaylor: If the model *is* security compromised, then we have no business allowing the user to execute anything.

What if we unified the two flags into "malicious_activity_warning" . which gets set transiently on a single MALICIOUS and permanently passed the thresholds.

This differs from your setup in that we'd display the scary warning for a single malicious activity detection, but I think that's probably appropriate. An attacker only has to get through *once*.

owtaylor (changes requested):

src/linux_mcp_server/tools/run_script.py
Jazzcort: Sounds great! I'm always bad at naming these. Will change!

src/linux_mcp_server/tools/run_script.py
Jazzcort: Make sense! Will change!

PR #431, opened 2026-04-23
refactor: improve execution tool guidance for script execution

Jazzcort:
This commit refactors MCP app detection and enhances the tool feedback loop to guide clients toward the correct execution tools.

Key changes:
* Extracted `_use_mcp_app_for_client` from `server.py` into a public `use_mcp_app_for_client` utility in `mcp_app.py` for cross-module reuse.
* Added a `_pick_execution_tool` helper in `run_script.py` to dynamically determine the right tool (`run_script`, `run_script_interactive`, or `run_script_with_confirmation`) based on script flags (readonly, needs_confirmation) and client capabilities.
* Updated `validate_script` to explicitly instruct the client on which execution tool to use next in its text response.
* Improved the error message in `run_script` to dynamically suggest the appropriate confirmation tool if the script requires confirmation but was called incorrectly.

### DIFF
--- a/src/linux_mcp_server/mcp_app.py
+++ b/src/linux_mcp_server/mcp_app.py
@@ -1,3 +1,57 @@
+from fastmcp.server.dependencies import get_context
+from mcp.types import InitializeRequestParams
+
+from linux_mcp_server.config import CONFIG
+
+
 RUN_SCRIPT_APP_URI = "ui://run_script_readonly_with_mcp_app/run-script-app.html"
 MCP_APP_MIME_TYPE = "text/html;profile=mcp-app"
 MCP_UI_EXTENSION = "io.modelcontextprotocol/ui"
+
+
+def use_mcp_app_for_client(client_params: InitializeRequestParams | None = None):
+    if client_params is None:
+        client_params = get_context().session.client_params
+
+    assert client_params is not None, (
+        "FastMCP framework error: client_params should not be None after `initialize` is done"
+    )
+    # The configuration can overwrite the MCP app support detection, so we have the flexibility to
+    # manually turn the Mcp app feature on/off for developing/testing purposes.
+    if CONFIG.use_mcp_apps is not None:
+        return CONFIG.use_mcp_apps
+
+    # For python-sdk -1.x, count on extensibility of protocol types - while this is being
+    # removed for v2, hopefully extensions will be there properly.
+    capabilities = client_params.capabilities
+    extensions = getattr(capabilities, "extensions", {})
+    mcp_ui_extension = extensions.get(MCP_UI_EXTENSION) or {}
+    mime_types = mcp_ui_extension.get("mimeTypes") or []
+
+    # The configuration can overwrite the MCP app support detection, so we have the flexibility to
+    # manually turn the Mcp app feature on/off for developing/testing purposes.
+    return MCP_APP_MIME_TYPE in mime_types
+
+
+def hide_app_tools_for_client(client_params: InitializeRequestParams | None = None):
+    # Versions of goose before 1.29.0 don't understand _meta.ui.visiblity, so would
+    # leak our app-only tools to the model. However, they also are happy to let the
+    # model call tools that aren't listed as well. So if we see such an old version
+    # of goose, we strip out the app-only tools.
+    if client_params is None:
+        client_params = get_context().session.client_params
+
+    assert client_params is not None, (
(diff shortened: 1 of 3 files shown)


Jazzcort:
Note: Returning `needs_confirmation: true` in the `structured_content` sometimes is not enough for the chat model to use the confirmation tool right a way. Without this patch, it's highly likely the chat model will still call `run_script` and hit the tool call error and then call the confirmation tool. This patch might also improve the situation when the flow just stops after `validate_script` without calling execution tool.

subpop:
I like this idea.

Jazzcort:
Let me fix the tests! 😁

panyamkeerthana:
Great idea! This should guide the model into the next step of the loop😁 Just curious, which model were you using to validate the new guidance?

Jazzcort:
@panyamkeerthana I use `gemini-3.1-pro-preview` most of the time for chat model.

Jazzcort:
I think I'll wait for some feedback on this. If it ends up quite positive and we want to merge this into main, I'll work on the test coverage then. 😁

src/linux_mcp_server/mcp_app.py
```diff
@@ -1,4 +1,27 @@
...
+from linux_mcp_server.config import CONFIG
+
+
 RUN_SCRIPT_APP_URI = "ui://run_script_readonly_with_mcp_app/run-script-app.html"
 ALLOWED_UI_RESOURCE_URIS = set([RUN_SCRIPT_APP_URI])
 MCP_APP_MIME_TYPE = "text/html;profile=mcp-app"
 MCP_UI_EXTENSION = "io.modelcontextprotocol/ui"
+
+
+def use_mcp_app_for_client(client_params: InitializeRequestParams):
```
owtaylor: Passing around client_params gets really tedious. We can use fastmcp's get_context() which gets the current context (async safely), but can't avoid client_params entirely because get_context() won't work in on_initialize, as far as I recall.

```suggestion
def use_mcp_app_for_client(client_params: InitializeRequestParams | None = None):
    if client_params is None:
        client_params = get_context().session.client_params
```

src/linux_mcp_server/tools/run_script.py
```diff
@@ -368,6 +369,16 @@ async def get_execution_state(id: str):
...
 
 
+def _pick_execution_tool(readonly: bool, needs_confirmation: bool, use_mcp_app: bool):
+    if not needs_confirmation and readonly:
+        return "run_script"
+
+    if use_mcp_app:
+        return "run_script_interactive"
+    else:
+        return "run_script_with_confirmation"
```
owtaylor: Don't see why readonly is needed here - that should already be reflected in needs_confirmation.

```suggestion
def _pick_execution_tool(needs_confirmation: bool):
    if not needs_confirmation:
        return "run_script"
    elif use_mcp_app_for_client():
        return "run_script_interactive"
    else:
        return "run_script_with_confirmation"
```

src/linux_mcp_server/tools/run_script.py
```diff
@@ -433,9 +456,17 @@ async def run_script(
...
+
+    if use_mcp_app_for_client(client_params):
+        confirmation_tool_name = "run_script_interactive"
+    else:
+        confirmation_tool_name = "run_script_with_confirmation"
+
     # Verify that this script doesn't require confirmation
     if script_details.needs_confirmation:
-        raise ToolError("This script requires confirmation. Use run_script_with_confirmation instead of run_script.")
+        raise ToolError(f"This script requires confirmation. Use {confirmation_tool_name} instead of run_script.")
```
owtaylor:
```suggestion
        raise ToolError(f"This script requires confirmation. Use {_pick_execution_tool(True)} instead of run_script.")
```

owtaylor (changes requested):
Looks like a nice cleanup and improvement! I think we can make it a bit cleaner by using `get_context()` rather than feeding around client_params.

I was worried when I started looking at it it would be a big conflict with #457, but I think it won't be bad whichever lands first.

Jazzcort:
Not sure if I should move `hide_app_tools_for_client` into mcp_app.py 🤔

owtaylor:
> Not sure if I should move `hide_app_tools_for_client` into mcp_app.py 🤔

I'm fine with it that way.

owtaylor (approved):
LGTM

Jazzcort:
Thanks for reviewing! 😁

PR #441, opened 2026-04-27
fix(scripts): extend timeout in verify_installation.sh

Jazzcort:
- Increased the log tailing timeout from 20s to 300s.
- Allows `uvx` to pass the installation test.
- Prevents timeouts caused by `uvx` needing extra time at startup to resolve and install dependencies on the fly.


subpop (approved):
LGTM

PR #450, opened 2026-05-07
ci(eval): migrate gatekeeper eval jobs to Vertex AI and update model …

Jazzcort:
Switch most gatekeeper evaluation jobs from OpenRouter/Models.corp to Vertex AI, using service account credentials downloaded via GitLab secure files. Add a new setup job (download-secure-files) to fetch credentials, and configure GOOGLE_APPLICATION_CREDENTIALS in the eval base template.

Model changes:
  - Add Vertex AI jobs: gpt-oss-120b, gpt-oss-20b, gemma-4-26b-a4b-it, gemini-3.1-pro-preview
  - Replace Models.corp gpt-oss-20b with granite-4.0-h-small
  - Remove claude-4.6-opus (OpenRouter) job
 
---
Eval results: https://jazzcort.gitlab.io/-/linux-mcp-server/-/jobs/14251569201/artifacts/index.html

Problems:
1. `gemma-4-26b-a4b-it` is not compatible with `litellm` for now, but there is a feature request for it -> https://github.com/BerriAI/litellm/issues/26083.
2. Inference speed is surprisingly slow for`gpt-oss-20b` through Vertex AI -> https://gitlab.com/Jazzcort/linux-mcp-server/-/pipelines/2505781867. See the job duration almost hit the default timeout (1h). When I ran the eval against `gpt-oss-20b` through Models.corp before, it usually just took about 15 mins to finish. Also, `gpt-oss-20b` through Vertex AI gives odd responses like `...?` sometimes. 

Screenshot 2026-05-07 at 4 03 56 PM

3. Need to consider `exception` case in the eval render app since it's also included in the result when passing `--output-all` to `run-eval.py`

### DIFF
--- a/.gitlab/ci/CREDENTIALS.md
+++ b/.gitlab/ci/CREDENTIALS.md
@@ -1,0 +1,36 @@
+# GitLab CI Credentials
+
+This document describes the credentials used by the GitLab CI pipelines and how they were provisioned.
+
+## CI/CD Variables
+
+These are configured in **Settings > CI/CD > Variables** on the GitLab project.
+
+### `GITLAB_TOKEN`
+
+- **Used by:** `mirror.yml` (PR mirroring workflow)
+- **Purpose:** Pushes mirrored GitHub PR branches into the GitLab repo.
+- **Type:** Personal or Project Access Token with `write_repository` scope.
+- **How to create:** GitLab > Settings > Access Tokens > create a token with `write_repository`. Then add it as a CI/CD variable (Protected + Masked).
+
+### `MODELS_CORP_*_API_KEY`
+
+- **Used by:** `eval-gatekeeper.yml` (models.corp eval jobs)
+- **Purpose:** Authenticates against the internal Red Hat model platform (`models.corp.redhat.com`) to run gatekeeper evals. Each variable follows the pattern `MODELS_CORP_<MODEL_NAME>_API_KEY` (e.g. `MODELS_CORP_GPT_OSS_20B_API_KEY`, `MODELS_CORP_GRANITE_4_0_H_SMALL_API_KEY`).
+- **Type:** API key issued by models.corp.
+- **How to create:** Check [models.corp user documentation](https://gitlab.cee.redhat.com/models-corp/user-documentation) for requesting an API key. Add it as a CI/CD variable (Protected + Masked). When adding a new models.corp eval job, create a corresponding variable following the naming convention.
+
+## Secure Files
+
+These are uploaded via **Settings > CI/CD > Secure Files** and downloaded at runtime by the `download-secure-files` job using `glab securefile download`.
+
+### `gatekeeper-eval-service-account-credential.json`
+
+- **Used by:** `eval-gatekeeper.yml` (all Vertex AI eval jobs, via `GOOGLE_APPLICATION_CREDENTIALS`)
+- **Purpose:** Authenticates to the Google Cloud project `rhel-lightspeed-650189` to call Vertex AI model endpoints (e.g. `gpt-oss-120b-maas`, `gemini-3.1-pro-preview`).
+- **Type:** Google Cloud Service Account key (JSON).
+- **How it was created:**
+  1. In the GCP console for project `rhel-lightspeed-650189`, a service account was created with the Vertex AI User role (or equivalent).
+  2. A JSON key was exported for that service account.
+  3. The JSON key file was uploaded to GitLab at **Settings > CI/CD > Secure Files**.
+- **Rotation:** Generate a new JSON key in GCP, re-upload to Secure Files, and delete the old key.
--- a/.gitlab/ci/eval-gatekeeper.yml
+++ b/.gitlab/ci/eval-gatekeeper.yml
@@ -24,6 +24,19 @@ setup-uv-and-mcp-apps:
       - mcp-app/dist/
     expire_in: 1 hour
 
+download-secure-files:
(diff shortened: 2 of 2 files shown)


Jazzcort:
https://jazzcort.gitlab.io/-/linux-mcp-server/-/jobs/14272534436/artifacts/index.html
Vertex AI is not quite stable. We might hit a few API errors in a batch.

owtaylor:
I did some digging into the gpt-oss-20b situation, and it turns out what is going on is that (as hosted on VertexAI) it sometimes goes into an infinite reasoning loop.

 * If the call hits the LiteLLM 10-minute default timeout before hitting the token limit, then you get a Timeout exception (this normally happens to me)
 * If the token limit is reached first, then you get a response like `{ "detail", "...?` which fails to parse. The `...?` text is added when the token limit is reached.

Adding `max_tokens=8000` gives a similar failure rate with a smaller hang. Adding `reasoning_effort="low"` avoids the reasoning loop, but probably degrades the quality a bit.

Probably the right thing to do here is to switch gpt-oss-20b back to models.corp, since the VertexAI implemention is poor. Gold standard would be to do a 3-way comparison with gpt-oss:20b

 - models.corp
 - VertexAI
 - llama.cpp

And see which of the two first is most like the llama.cpp results in terms of accuracy of results and bugginess ... since our interest in gpt-oss:20b is because it's a model that can be run locally.

Generally speaking, we can make *all* the inference much faster by doing it in parallel ... especially against VertexAI. I have some working code to make check_run_script() async and run the eval in parallel and it takes the eval runs against VertexAI down to a minute or so for any model. Well, except for the 10 minute timeouts - then we get a 

Will try to finish that up.

Jazzcort:
>I did some digging into the gpt-oss-20b situation, and it turns out what is going on is that (as hosted on VertexAI) it sometimes goes into an infinite reasoning loop.

Ah, that's why it was pretty fast for a few requests and then just stuck on one request for very long time.

I can maybe run the eval with llama.cpp and compare the results with the result from Vertex AI and models.corp and see how it goes. 😁

Jazzcort:
Test run on this patch: https://jazzcort.gitlab.io/-/linux-mcp-server/-/jobs/14315452254/artifacts/index.html. It was run with this patch merged into main (de5f42a1fcb4ad8a23a48361f0fb16e676359116).

Btw, https://jazzcort.gitlab.io/-/linux-mcp-server/-/jobs/14272534436/artifacts/index.html this is what I mentioned about `granite-4.0-h-small` performing well, but it's not true for the newest run. Maybe it's just the natural of small models that the output tends to vary a lot.

owtaylor (commented):
Looks good to me, only two things:

 * Any particular reason for using `rhel-lightspeed-650189` rather than the `rhel-lightspeed` one?
 * Where the credentials are coming from needs to be at least briefly documented. Maybe create a .ci/gitlab/CREDENTIALS.md describing briefly how thing were created and uploaded.

Jazzcort:
>Any particular reason for using rhel-lightspeed-650189 rather than the rhel-lightspeed one?

I also created a service account in `rhel-lightspeed` project but it was not granted for using Vertex AI API. 

```
exception: |-
      litellm.BadRequestError: Vertex_aiException BadRequestError - [{
        "error": {
          "code": 403,
          "message": "Permission 'aiplatform.endpoints.predict' denied on resource '//aiplatform.googleapis.com/projects/rhel-lightspeed/locations/us-central1/endpoints/openapi' (or it may not exist).",
          "status": "PERMISSION_DENIED",
          "details": [
            {
              "@type": "type.googleapis.com/google.rpc.ErrorInfo",
              "reason": "IAM_PERMISSION_DENIED",
              "domain": "aiplatform.googleapis.com",
              "metadata": {
                "resource": "projects/rhel-lightspeed/locations/us-central1/endpoints/openapi",
                "permission": "aiplatform.endpoints.predict"
              }
            }
          ]
        }
      }
      ]
```
And I don't have the permission to grant the service account for using Vertex AI API (Agents Platform).

The service account I created in `rhel-lightspeed-650189` is granted for the Vertex AI API access when it's created. It must be something with the project setting, but I just don't have enough right to check it. I asked @major in the lightspeed team channel and he didn't have the right either.

If we want to use use service account from `rhel-lightspeed` then we need to find the person who has the authority to change those settings.

---

I'll work on the `CREDENTIALS.md`!  😁

Iana-Kasimova-TR:
@Jazzcort, @owtaylor can I help? I really need this feature in llm, I am trying to use Gemma 4, which is currently accessible by api from vertex ai and it is hard to make it via litellm
will be happy to help!

owtaylor:
@Iana-Kasimova-TR 

Fixing https://github.com/BerriAI/litellm/issues/26083 properly is probably not too hard in the end, but the way things work is certainly convoluted. 

> @Jazzcort, @owtaylor can I help? I really need this feature in llm, I am trying to use Gemma 4, which is currently accessible by api from vertex ai and it is hard to make it via litellm
> will be happy to help!

I updated the litellm issue to have both describe how to work around and use the openai/ provider and also to have some hints about how to fix the issue properly, if you are interested in digging into that.

owtaylor (approved):
CREDENTIALS.md looks great . It's fine to run this with rhel-lightspeed-650189 account if that works and rhel-lightspeed doesn't - we dont' need to kick off a big investigation about *why*.

Jazzcort:
Sounds great! Thanks for reviewing! 😁

PR #451, opened 2026-05-07
fix(eval): handle exceptions in gatekeeper benchmark

Jazzcort:
Update compareResult to accept full result objects instead of just status strings, enabling early detection of exception entries. Remove a redundant comment for missing test results.

---

This patch addresses problem 3 mentioned in https://github.com/rhel-lightspeed/linux-mcp-server/pull/450


owtaylor (approved):
Looks good

Jazzcort:
Thanks for the review! 😁

PR #452, opened 2026-05-08
deps: Port to FastMCP 3.2

owtaylor:
Upgrade FastMCP to >= 3.2.4.

Changes:

 * Use built-in support for mcp-apps in FastMCP to simplify app and resource handling.
 * Remove awful workarounds to get access to to session from on_initialize() middleware.
 * mcp.get_tools() is replaced with mcp.list_tools()

There is a bug in FastMCP 3.2.4 where visibility=["app"] tools are not listed or callable - monkeypatch FastMCP to temporarily work around that. (Fix at https://github.com/PrefectHQ/fastmcp/pull/4112)

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -16,7 +16,7 @@ requires-python = ">=3.10"
 # no longer getting security fixes.
 dependencies = [
     "asyncssh[bcrypt] >= 2.22.0",
-    "fastmcp >=2.14.4, <2.14.6",
+    "fastmcp >= 3.2.4",
     "litellm>=1.80.16",
     "pydantic-settings >= 2.12.0",
     "pydantic >= 2.12.5",
--- a/src/linux_mcp_server/tools/run_script.py
+++ b/src/linux_mcp_server/tools/run_script.py
@@ -7,8 +7,9 @@ from dataclasses import asdict
 from dataclasses import dataclass
 
 from fastmcp import Context
+from fastmcp.apps import AppConfig
 from fastmcp.exceptions import ToolError
-from fastmcp.tools.tool import ToolResult
+from fastmcp.tools import ToolResult
 from mcp.types import ContentBlock
 from mcp.types import TextContent
 from mcp.types import ToolAnnotations
@@ -234,7 +235,7 @@ class ExecuteScriptResult:
 @mcp.tool(
     tags={"run_script", "hidden_from_model"},
     description="Execute a script; this is only available to the our mcp-app",
-    meta={"ui": {"visibility": ["app"]}},
+    app=AppConfig(visibility=["app"]),
 )
 @log_tool_call
 @disallow_local_execution_in_containers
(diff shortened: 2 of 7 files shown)


tests/test_middleware.py
Jazzcort: Even though the code path never accesses the session when `_use_mcp_app_for_client` is false, I think we should also change this to `session` so it's consistent with the other tests. 😁

src/linux_mcp_server/tools/run_script.py
```diff
@@ -7,6 +7,7 @@
 from dataclasses import dataclass
 
 from fastmcp import Context
+from fastmcp.apps import AppConfig
 from fastmcp.exceptions import ToolError
 from fastmcp.tools.tool import ToolResult
```
Jazzcort: Due to the renaming of `tool.py` to  `base.py` (https://github.com/PrefectHQ/fastmcp/pull/3540), I get a syntax error here - 'Import "fastmap.tools.tool" could not be resolved'. In the runtime, it's working fine due to the internal re-export, but I think we should just change the import of `ToolResult` from `fastmcp.tools.tool` to `fastmcp.tools`. 😁

pyproject.toml
```diff
@@ -16,7 +16,7 @@ requires-python = ">=3.10"
 # no longer getting security fixes.
 dependencies = [
     "asyncssh[bcrypt] >= 2.22.0",
-    "fastmcp >=2.14.4, <2.14.6",
+    "fastmcp >= 3.2.4",
```
Jazzcort, line 19: Do you think we should put an upper bound here `<4`? @owtaylor

Jazzcort (commented):
I just have a few thoughts on some minor details. Everything else looks good to me! 😁

tests/test_middleware.py
owtaylor: Good catch, fixed.

src/linux_mcp_server/tools/run_script.py
owtaylor: Yes, I see as well in my editor or for 'make verify' - fixed.

pyproject.toml
owtaylor: Because of `"rangeStrategy": "in-range-only"` in `renovate.json` we want to do as little pinning from the top as possible, otherwise we'll lose track of when new versions come in

I think we'll generally have enough warning of a major incompatible version bump to figure out what is needed and put in the appropriate bound at that point. 

Also, if we did want to change it, then we'd want to do it for everything -  `asyncssh >= 2.22.0, < 3` and so forth -

owtaylor:
Had to fix a few things with linting, and the `get_tools => list_tools` change wasn't right since it interacted badly with our middleware (depending on test ordering - if main() gets called, then suddenly our middleware gets added to mcp ... needs cleaning up but not as part of this PR.)

With that change all:
 * All the CI tests pass (except for codecov, but it's marginal changes.)
 * The Konflux build passes

pyproject.toml
Jazzcort: Agree! 😁

Jazzcort (approved):
>Had to fix a few things with linting, and the get_tools => list_tools change wasn't right since it interacted badly with our middleware (depending on test ordering - if main() gets called, then suddenly our middleware gets added to mcp ... needs cleaning up but not as part of this PR.)

Sounds painful to fix. I hope the cleaning up process will go easy on you. 😁 Let me know if you need help on that!

LGTM!

PR #454, opened 2026-05-08
Add SPDX license headers and .containerignore

Amy-Ra-lph:
## Summary

- Add Apache-2.0 SPDX license identifiers and copyright headers to all 98 source files that were missing them (96 Python + 2 shell scripts), matching the format already used in `tests/functional/`
- Add `.containerignore` to exclude `.git`, tests, docs, and caches from container builds

## Motivation

A prodsec security audit (skill-audit) flagged:
- **LIC-002**: 96 source files missing SPDX headers
- **LIC-005**: No copyright notices in source files
- **CTR-003**: No `.containerignore` file

These are low-risk, high-hygiene improvements. The SPDX headers match the existing format used in `tests/functional/` files.

## Test plan

- [ ] Verify headers don't break imports: `python -c "import linux_mcp_server"`
- [ ] Verify container builds: `podman build -f Containerfile -t test .`
- [ ] Verify no source files missed: `grep -rL "SPDX-License-Identifier" --include="*.py" src/ tests/ eval/ scripts/`

### DIFF
--- a/.containerignore
+++ b/.containerignore
@@ -1,0 +1,14 @@
+.git
+.github
+tests/
+eval/
+docs/
+*.md
+*.lock
+.pre-commit-config.yaml
+.secrets.baseline
+__pycache__
+*.pyc
+.mypy_cache
+.ruff_cache
+.pytest_cache
--- a/.containerignore
+++ b/.containerignore
@@ -1,14 +1,11 @@
+# Toplevel inclusion is controlled by ADD lines in the Containerfile
+# Exclude possible nested leftover files
+**/__pycache__
+**/*.pyc
+**/*.bak
+**/*~
+# Exclude large directories from context for performance
 .git
-.github
-tests/
-eval/
-docs/
-*.md
-*.lock
-.pre-commit-config.yaml
-.secrets.baseline
-__pycache__
-*.pyc
-.mypy_cache
-.ruff_cache
-.pytest_cache
+.venv/
+dist/
+mcp-app/node_modules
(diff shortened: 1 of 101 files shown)


.containerignore
```diff
@@ -0,0 +1,14 @@
+.git
+.github
+tests/
+eval/
+docs/
+*.md
+*.lock
+.pre-commit-config.yaml
+.secrets.baseline
+__pycache__
```
owtaylor: This is not recursive - what was probably meant was `**/__pycache__`

.containerignore
```diff
@@ -0,0 +1,14 @@
+.git
+.github
+tests/
+eval/
+docs/
```
owtaylor: This list seems to exclude a fairly arbitrary set of git-committed files (docs/, *.md, etc.) It also doesn't cover all generated files `.venv` (!), `dist/` `coverage/`, etc.

Our Containerfile explicitly calls ADD for the files it *does* want ... that seems more maintainable then trying to exclude everything that might not be wanted.

I think we should just keep this to the things that might be nested inside things we explicitly ADD, and a few things for performance if the context is being copied.

```
# Toplevel inclusion is controlled by ADD lines in the Containerfile
# Exclude possible nested leftover files
**/__pycache__
**/*.bak
**/*.lock
**/*~
# Exclude large directories from context for performance
.venv/
dist/
mcp-app/node-modules
```

Something like that.

owtaylor (changes requested):
Thanks for the contribution!

PR #455, opened 2026-05-08
fix(ci): replace negation operator with null/empty checks in GitLab CI rules

Jazzcort:


owtaylor (commented):
LGTM! (The negation operator was added in March - but too new to be on gitlab.cee)

owtaylor (approved):

PR #462, opened 2026-05-18
fix(gatekeeper): pass a looser schema to the model

owtaylor:
In evals, Claude Opus and Claude Sonnet were performing *much* worse when run directly from Anthropic than when run from Vertex AI (and worse than other flagship models)

Switching the schema passed to response_format so that the reason is optional first greatly improves the Claude models on the Anthropic pplatforms (Opus from 76.7% => 93.2%) and doesn't have much noticeable effect on other models.

Theory here is:
 - The response_format isn't ending up in the prompt for Vertex AI, though it may constrain decoding.
 - When the response_format does end up in the prompt, being required to provide a reason "intimidates" the model, and responding OK seems easier.

### DIFF
--- a/src/linux_mcp_server/gatekeeper/check_run_script.py
+++ b/src/linux_mcp_server/gatekeeper/check_run_script.py
@@ -95,8 +95,6 @@ If the status is not OK, the detail should be short 1-3 sentence description of
 what is wrong with the script. Be specific to allow the language model to correct
 the problem.
 
-If status is OK, the detail should be an empty string.
-
 If the script seems buggy but does not fall into any of the categories above, return
 a status of `OK`.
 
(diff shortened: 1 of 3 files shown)


src/linux_mcp_server/gatekeeper/check_run_script.py
Jazzcort: Do we still keep this or should we just remove it? 🤔

Jazzcort (dismissed):
If this makes every model happy, let's do it! 😁 If you think we need to keep GatekeeperResultStrict, just go ahead and merge it!

src/linux_mcp_server/gatekeeper/check_run_script.py
owtaylor: Should be removed - good catch! Pushed a new version.

Jazzcort (approved):
LGTM! 😁

PR #463, opened 2026-05-18
fix(gatekeeper): Tweak prompt to be more explicit

owtaylor:
Examing gatekeeper results in detail came up with these improvements:

 - Make it clearer what we mean by "does not modify the system" by repeating it as "does only read-only operations" - some models were treating things like starting and stopping services as not modifying the system.
 - Be explicit that "pre-configured repositories" means repositories configured *on the system* and give examples ... (models sometimes thought that, e.g., https://sh.rustup.rs is a preconfigured repository.)
 - Change 'general policies to 'policies listed under "General Policies" above', otherwise models would apply the POLICY status to things that should have been marked as MODIFIES_SYSTEM, etc.

### DIFF
--- a/src/linux_mcp_server/gatekeeper/check_run_script.py
+++ b/src/linux_mcp_server/gatekeeper/check_run_script.py
@@ -21,7 +21,7 @@ def get_model() -> str:
 
 
 READONLY_INSTRUCTION = """
- * The script does not modify the system in any way.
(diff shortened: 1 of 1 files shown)


Jazzcort (approved):
LGTM! The changes make the prompts sound more precise. 😁

PR #464, opened 2026-05-18
fix(gatekeeper): suppress litellm debugging message

owtaylor:
LiteLLM has bugs where the custom_llm_provider is not properly passed back into query functions for things like "model supports reasoning". Set 'litellm.suppress_debug_info = True' to suppress debug messages - these aren't *our* bugs.

https://github.com/BerriAI/litellm/issues/23879


Jazzcort (approved):
LGTM! Always want to do this but keep skipping it. 😆

PR #465, opened 2026-05-18
feat(config): Make the gatekeeper model more configurable

owtaylor:
Move the gatekeeper configuration into a nested member off our config object, so LINUX_MCP_GATEKEEPER_MODEL becomes LINUX_MCP_GATEKEEPER__MODEL (LINUX_MCP_GATEKEEPER_MODEL is supported as a deprecated alias.)

Add controls for:
 reasoning_effort: turn off or down reasoning often make models
   perform better for us.
 structured_output: e.g. for gemma-4-31b-it, turning off
   response_format is needed to keep the model from going into
   infinite looop.
 temperature: Anthropc models need a non-zero temperature to
   enable reasoning.
 quantization: OpenRouter mixes together models with different
   quantization in a single model name - specifying a specific
   quantization is needed for clean benchmarking data.

### DIFF
--- a/eval/gatekeeper/run-eval.py
+++ b/eval/gatekeeper/run-eval.py
@@ -219,9 +219,9 @@ def main(
         typer.echo("Must specify either a test case file or --all.", err=True)
         raise typer.Exit(code=1)
 
-    if "LINUX_MCP_GATEKEEPER_MODEL" not in os.environ:
+    if "LINUX_MCP_GATEKEEPER__MODEL" not in os.environ and "LINUX_MCP_GATEKEEPER_MODEL" not in os.environ:
         typer.echo(
-            "Please set the LINUX_MCP_GATEKEEPER_MODEL environment variable to specify the Gatekeeper model to use."
+            "Please set the LINUX_MCP_GATEKEEPER__MODEL environment variable to specify the Gatekeeper model to use."
         )
         raise typer.Exit(code=1)
 
--- a/src/linux_mcp_server/config.py
+++ b/src/linux_mcp_server/config.py
@@ -1,16 +1,23 @@
 """Settings for linux-mcp-server"""
 
+import logging
+import os
 import sys
 
 from pathlib import Path
+from typing import Any
 
 from pydantic import Field
+from pydantic import model_validator
 from pydantic import SecretStr
 from pydantic_settings import BaseSettings
 from pydantic_settings import SettingsConfigDict
 
 from linux_mcp_server.utils.enum import StrEnum
 from linux_mcp_server.utils.types import UpperCase
+
+
+logger = logging.getLogger(__name__)
 
 
 class Transport(StrEnum):
@@ -25,6 +32,18 @@ class Toolset(StrEnum):
     FIXED = "fixed"
     RUN_SCRIPT = "run_script"
     BOTH = "both"
+
+
+class ReasoningEffort(StrEnum):
+    """Reasoning effort levels for the gatekeeper model."""
+
+    NONE = "none"
+    MINIMAL = "minimal"
+    LOW = "low"
+    MEDIUM = "medium"
+    HIGH = "high"
+    XHIGH = "xhigh"
+    DEFAULT = "default"
 
 
 class AuthProvider(StrEnum):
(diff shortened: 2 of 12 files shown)


eval/gatekeeper/run-eval.py
```diff
@@ -219,9 +219,9 @@ def main(
         typer.echo("Must specify either a test case file or --all.", err=True)
         raise typer.Exit(code=1)
 
-    if "LINUX_MCP_GATEKEEPER_MODEL" not in os.environ:
+    if "LINUX_MCP_GATEKEEPER__MODEL" not in os.environ and "LINUX_MCP_GATEKEEPER__MODEL" not in os.environ:
```
Jazzcort: Duplicate conditions

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -180,6 +183,44 @@ class GatekeeperResultStrict(GatekeeperResult):
...
+    if structured_output is None:
+        params = get_supported_openai_params(model=model)
+        structured_output = params is not None and "response_format" in params
+
+    if structured_output:
+        extra_kwargs["response_format"] = GatekeeperResultStrict
+
+    reasoning_effort = CONFIG.gatekeeper.reasoning_effort
+    if reasoning_effort is not None:
+        if reasoning_effort == ReasoningEffort.ENABLE_THINKING:
```
Jazzcort: I'm so confused by the terms they're using for these LLM parameters. I thought "thinking" and "reasoning" mean the same thing, but they're used in different places at the same time. 
Since "thinking" and "reasoning" settings are mixed together in `ReasoningEffort`, does it mean enabling "thinking" is equal to enabling "reasoning" with a default "reasoning_effort.value" under the hood for those LLM?

The reason I ask this is because if users set `LINUX_MCP_GATEKEEPER__reasoning_effort="enable_thinking"` then they can not change the reasoning_effort. I suspect enabling "reasoning" means the same thing as enabling "thinking" for those LLMs and If user want more fine-grained control for how deep the model should reason for they should always use "reasoning" for it instead of "thinking".

Let me know if I'm wrong! Just want to clarify my doubts on these 😁

src/linux_mcp_server/config.py
```diff
@@ -165,9 +203,25 @@ def transport_kwargs(self):
...
     # def validate_gatekeeper_model(self):
-    #     if self.toolset != Toolset.FIXED and self.gatekeeper_model is None:
-    #         raise ValueError('gatekeeper_model must be set unless the toolset is "fixed"')
+    #     if self.toolset != Toolset.FIXED and self.gatekeeper.model is None:
+    #         raise ValueError('gatekeeper.model must be set unless the toolset is "fixed"')
     #     return self
 
+    @model_validator(mode="before")
+    @classmethod
+    def handle_deprecated_aliases(cls, data: Any) -> Any:
```
Jazzcort: I'm wondering if we're not using the `cls` inside the function, can we just use `@staticmethod` instead?

Jazzcort (changes requested):
Except for the duplicate conditions, the others are just my thoughts and questions. 😁 Also need to resolve some conflicts

eval/gatekeeper/run-eval.py
owtaylor: Good catch!

src/linux_mcp_server/gatekeeper/check_run_script.py
owtaylor: thinking === reasoning ... just diferent words for the same thing. The "enable_thinking" here is a gruesome thing ... to quote my docs:

+| `--gatekeeper.reasoning_effort`<br>`LINUX_MCP_GATEKEEPER__REASONING_EFFORT`   | _(model specific)_ | Reasoning effort to use for gatekeeper model (`none`, `minimal`, `low`, `medium`, `high`, `xhigh`). Not all values are supported for all models. Special values (`enable_thinking`, `disable_thinking`) set the `enable_thinking` chat template parameter. |

Basically, when using llama.cpp, the only way to turn reasoning on or off through the API is via chat template parameters. (There's a `--reasoning on/off` cli parameter.) 

 * qwen3.5 and gemma4 - enable_thinking
 * gpt-oss - reasoning_effort
 * other models: ??? 

Other possible approaches:

1. We could put this logic with specific model names right in our code. This has the advantage that it just works. But it's not extensible - if we ship with qwen3.5 in our code, there's no way to set it for qwen3.6
2. We could do it generic ...  `LINUX_MCP_GATEKEEPER__TEMPLATE_KWARGS__ENABLE_THINKING=true` ... `LINUX_MCP_GATEKEEPER__TEMPLATE_KWARGS__REASONING_EFFORT=\"low\"`

Note that using llama.cpp is a bit of an edge case ... it works well for our evals, but it's inherently not friendly to a user. So we don't need to make things pretty.

src/linux_mcp_server/config.py
owtaylor: Should work [tries the test case], does work.

src/linux_mcp_server/gatekeeper/check_run_script.py
Jazzcort: Thanks for the clarification! I can't agree more that llama.cpp is just not friendly to a user. These settings are just to make the eval work well so I agree too that we just don't bother to make things pretty here 😁

owtaylor:
Pushed a new version:
 - rebased to main
 - with fixes for the things you commented on
 - with `template_kwargs` instead of the reasoning_effort=ENABLE_THINKING hack

owtaylor:
New version fixes some typos in the docs, and cleans up the mess that prettier's new Markdown support made.

Jazzcort (approved):
LGTM! 😁

PR #470, opened 2026-05-19
test(functional): Add '--skip-summary' to the testing-farm command line

owtaylor:
Generating the summary table requires access to the artifact storage for Testing Farm, which for the Red Hat ranch requires being on the Red Hat. As of recently, failures started being fatal, and an explicit --skip-summary is needed.

I haven't tested this, but from conversation on the #testing-farm Slack channel, this seems to be the right fix for recent failures, like:

https://github.com/rhel-lightspeed/linux-mcp-server/actions/runs/26107359758

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -128,6 +128,8 @@ jobs:
           PLAN: general
         run: |
           echo "Starting functional tests..."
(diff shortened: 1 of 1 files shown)


alexxa (approved):
LGTM, thanks @owtaylor

PR #471, opened 2026-05-19
Skip functional test workflow for non-trigger issue comments

owtaylor:
Add a job-level `if` condition to skip the check-authorization job when an issue_comment event doesn't match a trigger command (/test-functional or /retest). This prevents noisy failed workflow runs from bot comments and regular PR discussion.


alexxa (approved):
LGTM, thanks @owtaylor

PR #473, opened 2026-05-19
docs: Fix build for FastMCP 3

owtaylor:
Replace usage of mcp.get_tools() with mcp._list_tools (avoid mcp.list_tools() to avoid unhappy middlewhare).

PR #475, opened 2026-05-21
Use commit status to improve github/gitlab integration

owtaylor:
The current system where we post comments to Git Hub issues with an URL to create a job to sync them to gitlab has various issues:

 * The comment posting is noisy ... it really doesn't need to end up in the email notification stream
 * There's no way to tell if there's an existing sync'ed PR without going to the Git Lab web UI and looking for it
 * It's hard to tell if the CI jobs that are the point of the sync completed successfully or not - again you have to poke arond in the GitLab web UI to find the sync'ed MR.

This pull request attempts to improve things by using the Git HUb "commit status" API.

 - Instead of posting comments saying how to sync, that's done as a "Pending" status where the details link takes you to the same "pipeline/new" form.
 - When the sync happens, that status is shown as complete
 - After the sync happens, we poll to see the Konflux job kick off and complete, and update a separate status for that

We might eventually also want to post the Konflux results as a comment in addition to the status. Also, this doesn't try to figure out how to post back the results of the gatekeeper eval job - I'm not sure if that actually makes sense as a status at all, or just as a comment.

One annoying wart is that the avatar of the status is the user who created the PAT, so if I create the token, then my face shows up:

image

We might want to create a "robot" GitHub account with shared access to get around this. ([Some guidance on handling auth for that](https://docs.github.com/en/organizations/keeping-your-organization-secure/managing-two-factor-authentication-for-your-organization/managing-bots-and-service-accounts-with-two-factor-authentication))

### DIFF
--- a/.github/workflows/gitlab-trigger.yml
+++ b/.github/workflows/gitlab-trigger.yml
@@ -1,36 +1,36 @@
-name: GitLab Trigger Comment
+name: GitLab Mirror Status
 
 on:
   pull_request_target:
     types: [opened, synchronize, reopened]
 
 permissions:
-  pull-requests: write
+  statuses: write
 
 jobs:
-  post-gitlab-link:
-    name: Post GitLab Trigger Link
+  set-gitlab-status:
+    name: Set GitLab Mirror Status
     runs-on: ubuntu-latest
 
     steps:
-      - name: Add GitLab trigger comment
+      - name: Create pending commit status
         uses: actions/github-script@v9
         with:
           script: |
             const prNumber = context.payload.pull_request.number;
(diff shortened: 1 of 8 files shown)


owtaylor:
The "Sanity/Type" failure is fixed by #473 ... it shows up here because this patch adds the scripts/ directory to the list of directories we type-check.

Jazzcort (dismissed):
LGTM! Should merge it and see how it goes. 😁

owtaylor:
Needed another miscellaneous type-checking fix since scripts/ is now type-checked.

Jazzcort (approved):
LGTM! 😁

PR #479, opened 2026-05-24
refactor(gatekeeper): async eval runner with structured classes and stats tracking

owtaylor:
- Make check_run_script async (acompletion) and add check_run_script_with_stats returning GatekeeperStats (tokens, cost, latency) alongside results
- Introduce EvalSuite and FileEval classes in run-eval.py, replacing global variables and eliminating duplicate single-file/multi-file code paths
- Load all test cases up front so progress output shows global numbering (e.g. [3/121]) as individual evals complete
- Add StatsAggregator dataclass for cleaner inference statistics reporting
- Add GatekeeperException with optional stats for timeout, output limit, and parse failures
- Add gatekeeper.cost config option for custom per-token cost accounting
- Switch summary/stats tables from plain text to rich.Table

### DIFF
--- a/eval/gatekeeper/run-eval.py
+++ b/eval/gatekeeper/run-eval.py
@@ -1,9 +1,12 @@
 #!/usr/bin/env python3
 
+import asyncio
 import json
 import os
-import sys
+import statistics
 
+from dataclasses import dataclass
+from dataclasses import field
 from pathlib import Path
 from typing import Annotated
 from typing import Literal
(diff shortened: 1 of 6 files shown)


src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -116,6 +121,12 @@ def get_model() -> str:
 """
 
 
+# Maximum number of completion tokens (including reasoning)
+GATEKEEPER_MAX_TOKENS = 8000
```
Jazzcort, line 125: Not sure if it makes sense that our users would like modify these values. These numbers are pretty reasonable and shouldn't be hit quite often. 😁

eval/gatekeeper/run-eval.py
```diff
@@ -1,173 +1,333 @@
...
-    if isinstance(data, dict) and "cases" in data:
-        test_cases = data["cases"]
-    else:
-        test_cases = None
+@dataclass
+class StatsAggregator:
+    stats: list[GatekeeperStats]
+    _values: dict[str, list[float]] = field(init=False, default_factory=dict)
+
+    def values(self, name: FieldName):
```
Jazzcort, line 42: I like this lazy processing concept!

eval/gatekeeper/run-eval.py
```diff
@@ -1,173 +1,333 @@
...
-        expected = test_case.get("result")
-        if expected is None or actual is None:
-            continue
+        table.add_column("File", justify="right", style="cyan", no_wrap=True)
+        for header in columns:
+            table.add_column(header)
 
-        expected_status = expected.get("status")
-        actual_status = actual.get("status")
+        totals = [0 for i in range(len(columns))]
```
Jazzcort: Find a perfect spot to use `totals = [0] * len(colums)`! 😁

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -248,16 +294,43 @@ def check_run_script(description: str, script_type: str, script: str, *, readonl
...
 
-    return GatekeeperResult.model_validate_json(response_text)
+    try:
+        return GatekeeperResult.model_validate_json(response_text), stats
+    except ValidationError as e:
+        logger.warning("Failed to Failed to parse gatekeeper model output: %s", e)
+        raise GatekeeperException("Failed to parse gatekeeper model output", stats=stats) from e
+
+
+async def check_run_script(description: str, script_type: str, script: str, *, readonly: bool) -> GatekeeperResult:
```
Jazzcort, line 334: It just bugs me a little bit that when users are using our mcp server and these stats are generated and never got used with this wrapper solution. But actually, it does not harm the performance since getting these stats does not introduce any heavy computation burden.

If we later on want to avoid getting these stats in every check_run_script call, we can either have a function overload 
```python
﻿@overload
async def check_run_script(
    description: str, script_type: str, script: str, *, readonly: bool, with_stats: Literal[False]
) -> GatekeeperResult: ...


@overload
async def check_run_script(
    description: str, script_type: str, script: str, *, readonly: bool, with_stats: Literal[True]
) -> tuple[GatekeeperResult, GatekeeperStats]: ...
``` 
or have a hidden private attribute in `GatekeeperResult` that does not get serialized.

```python
class GatekeeperResult(BaseModel):
    status: GatekeeperStatus
    detail: str = ""

    _stats: Optional["GatekeeperStats"] = PrivateAttr(default=None)
    ...
```
😁

Jazzcort (dismissed):
LGTM! Comments are just some of my thoughts! 😁

eval/gatekeeper/run-eval.py
owtaylor: Where intuition fails - my intuition was that is clearer, but would be a bit slower, though not enough to matter.  ("It's going to to allocate a bunch of lists, then append them").
```
>>> import timeit
>>> timeit.timeit(lambda: [0] * 1000, number=10000)
0.02545745200040983
>>> timeit.timeit(lambda: [0 for i in range(1000)], number=10000)
0.6684908530005487
```
But it actually seems well optimized, and avoiding Python in the inner loop is a huge win. (Not that it matters.)

src/linux_mcp_server/gatekeeper/check_run_script.py
owtaylor: These are basically just there as safety fallbacks to avoid the model running on forever - I don't immediately a use case for modifying them:

 * There's no reason to make them bigger.
 * If a user is actually hitting them enough to care, they will be having a bad day.

But we can always add configuration later if it seems useful.

src/linux_mcp_server/gatekeeper/check_run_script.py
owtaylor: Definitely possible alternate structures ... might even be better! Will stick to the current setup for now.

owtaylor:
Pushed a new version with `[0] * len(columns)` ...

eval/gatekeeper/run-eval.py
Jazzcort: Thanks for even testing the performance of it! I'm surprised a performance difference exists—I thought it was just a different way to do the same thing 😆 The usage is actually pretty limited since this only works when setting up a default list with immutable objects, so this is the perfect place to keep it. 😁

src/linux_mcp_server/gatekeeper/check_run_script.py
Jazzcort: Totally agree! 😁

Jazzcort (approved):
LGTM!

PR #481, opened 2026-05-26
chore(deps): enforce 7-day minimum package age for npm and uv

Jazzcort:
Mitigate supply chain risk by only installing packages that have been published for at least 7 days. Configured via npm's min-release-age and uv's exclude-newer settings.

While lockfiles generally protect us from supply chain attacks, certain development scenarios still expose us to risk. For instance, when adding a new package or manually modifying a version to test features, the lockfile cannot guard against potential threats.

### DIFF
--- a/mcp-app/.npmrc
+++ b/mcp-app/.npmrc
@@ -1,0 +1,1 @@
+min-release-age=7
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -210,3 +210,6 @@ order-by-type = false
 artifacts = [
     "mcp-app/dist",
 ]
+
+[tool.uv]
+exclude-newer = "7 days"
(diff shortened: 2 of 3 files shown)


owtaylor (commented):
I sort of worry about weird effects if renovate and uv and npm don't interpret "newer than 7 days" in exactly the same way.

But in terms of CI, the behavior seems to be that 'uv sync --locked' and 'npm install' simply ignore exclude-newer and min-release-age and install whatever is in the lock file - they only take effect during dependency *resolution*. I think that avoids most such possible problems.

I also note that:
https://docs.astral.sh/uv/guides/integration/renovate/#dependency-cooldown
recommends setting the renovate value to exclude-newer exactly.

owtaylor (approved):
Fine to merge, in my opinion. (I actually meant to approve not just comment before)

Jazzcort:
Thanks for reviewing!

PR #486, opened 2026-05-28
[RSPEED-3010]feat: add ExecutionContext for authorization

panyamkeerthana:
added ExecutionContext system to authorization at the execution with fail-closed

- added ExecutionContext model with permission fields
- added ContextVar for context propagation
- modified execute_command() to require ExecutionContext
- modified AuthorizationMiddleware to set ExecutionContext
- added tests for ContextVar and execute_command integration
- added TestExecutionContextMiddlewareIntegration tests

### DIFF
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -24,6 +24,7 @@ from linux_mcp_server.audit import log_ssh_command
 from linux_mcp_server.audit import log_ssh_connect
 from linux_mcp_server.audit import Status
 from linux_mcp_server.config import CONFIG
+from linux_mcp_server.execution_context import get_execution_context
 from linux_mcp_server.utils.types import Host
 
 
@@ -376,11 +377,24 @@ async def execute_command(
         ...     username="admin"
         ... )
     """
+    # Get execution context, fail-closed if not set
+    context = get_execution_context()
+    if context is None:
+        raise RuntimeError("No execution context set cannot execute commands")
+
     cmd_str = " ".join(command)
 
     if host:
+        # Remote execution, check permissions
+        if not context.allow_ssh_default and context.ssh_key_path is None:
+            raise RuntimeError("Remote execution not allowed")
+
         logger.debug(f"Routing to remote execution: {host} | command={cmd_str}")
         return await _connection_manager.execute_remote(command, host, encoding=encoding)
+
+    # Local execution,check permissions
+    if not context.allow_local:
+        raise RuntimeError("Local execution not allowed")
 
     logger.debug(f"LOCAL_EXEC: {cmd_str}")
     return await _execute_local(command, encoding=encoding)
--- a/src/linux_mcp_server/connection/ssh.py
+++ b/src/linux_mcp_server/connection/ssh.py
@@ -111,7 +111,18 @@ class SSHConnectionManager:
         Raises:
             ConnectionError: If connection fails
         """
-        key = f"{host}"
+        # Get SSH credentials from ExecutionContext if available
+        context = get_execution_context()
+        if context is not None:
+            ssh_key = str(context.ssh_key_path) if context.ssh_key_path else self._ssh_key
+            username = context.ssh_key_user or CONFIG.user
+        else:
+            # No context set - use defaults
+            ssh_key = self._ssh_key
+            username = CONFIG.user
+
+        # Build pool key including SSH key and username to avoid connection reuse conflicts
+        key = f"{host}:{ssh_key or 'default'}:{username or 'default'}"
 
         # Return existing connection if available
         if key in self._connections:
@@ -125,7 +136,7 @@ class SSHConnectionManager:
                     username=conn.get_extra_info("username"),
                     status=Status.success,
                     reused=True,
-                    key_path=self._ssh_key,
+                    key_path=ssh_key,
                 )
                 return conn
             else:
@@ -135,7 +146,7 @@ class SSHConnectionManager:
 
         # Create new connection
         # DEBUG level: Log connection attempt before it completes
-        logger.debug(f"{Event.SSH_CONNECTING}: {key} | key={self._ssh_key or 'none'}")
+        logger.debug(f"{Event.SSH_CONNECTING}: {key} | key={ssh_key or 'none'}")
 
         try:
             # Determine host key verification settings
@@ -151,11 +162,13 @@ class SSHConnectionManager:
                 "passphrase": CONFIG.key_passphrase.get_secret_value() or None,
             }
 
-            if self._ssh_key:
-                connect_kwargs["client_keys"] = [self._ssh_key]
+            # Use custom SSH key if provided, otherwise use default
+            if ssh_key:
+                connect_kwargs["client_keys"] = [ssh_key]
 
-            if CONFIG.user:
-                connect_kwargs["username"] = CONFIG.user
+            # Use custom username if provided, otherwise use CONFIG.user
+            if username:
+                connect_kwargs["username"] = username
 
             conn = await asyncssh.connect(**connect_kwargs)
             self._connections[key] = conn
@@ -166,7 +179,7 @@ class SSHConnectionManager:
                 username=conn.get_extra_info("username"),
                 status=Status.success,
                 reused=False,
-                key_path=self._ssh_key,
+                key_path=ssh_key,
             )
 
             # DEBUG level: Log pool state
--- a/src/linux_mcp_server/execution_context.py
+++ b/src/linux_mcp_server/execution_context.py
@@ -1,0 +1,31 @@
+from contextlib import contextmanager
+from contextvars import ContextVar
+from pathlib import Path
+
+from pydantic import BaseModel
+
+
+class ExecutionContext(BaseModel):
+    allow_local: bool = False
+    allow_ssh_default: bool = False
+    ssh_key_path: Path | None = None
+    ssh_key_user: str | None = None
+
+
+# Global ContextVar for storing the current execution context if no context is set execution should fail
+execution_context_var: ContextVar[ExecutionContext | None] = ContextVar("execution_context", default=None)
+
+
+@contextmanager
+def use_execution_context(context: ExecutionContext):
+
+    token = execution_context_var.set(context)
+    try:
+        yield context
+    finally:
+        execution_context_var.reset(token)
+
+
+# Get the current execution context or None if not set
+def get_execution_context() -> ExecutionContext | None:
+    return execution_context_var.get()
--- a/src/linux_mcp_server/server.py
+++ b/src/linux_mcp_server/server.py
@@ -28,6 +28,8 @@ from linux_mcp_server.auth_policy import PolicyAction
 from linux_mcp_server.config import CONFIG
 from linux_mcp_server.config import Toolset
 from linux_mcp_server.config import Transport
+from linux_mcp_server.execution_context import ExecutionContext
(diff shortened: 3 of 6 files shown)


tests/test_server.py
```diff
@@ -334,3 +334,129 @@ async def test_ssh_key_with_no_config_raises(self, setup: SetupFn, mocker):
...
+        mocker.patch("linux_mcp_server.server.CONFIG.policy_path", "/etc/policy.json")
+        mocker.patch(
+            "linux_mcp_server.auth_policy.get_policy",
+            return_value=AuthPolicy(
+                rules=[PolicyRule(host="localhost", tools=["@fixed"], all_users=True, action=PolicyAction.LOCAL)]
+            ),
+        )
+
+        await mcp_client.call_tool("get_memory_information")
+
```
owtaylor, line 384:
```suggestion
        captured_context = capture_context.get_context()
```

You'd use the local fixture like this.

tests/test_server.py
```diff
@@ -334,3 +334,129 @@ async def test_ssh_key_with_no_config_raises(self, setup: SetupFn, mocker):
 
         with pytest.raises(ToolError, match=r"Policy validation error: SSH_KEY action requires ssh_key configuration."):
             await client.call_tool("get_memory_information", {"host": "server1.example.com"})
+
+
+# Verify stdio without policy sets ExecutionContext with allow_local=True, allow_ssh_default=True
+class TestExecutionContextMiddlewareIntegration:
```
owtaylor: This looks like a good way to get the context. We can extract the commonality between the tests into a local fixture. The helper class here isn't absolutely necessary - you could stick closer to what you had, but I think makes things more understandable. 

```suggestion
class CaptureContext:
    def __init__(self):
        self._context = None
        
    async def _execute_with_fallback(self, *args, **_kwargs):
        self._context = get_execution_context()
        return (0, FREE_OUTPUT, "")

    def get_context():
         return self._context 
         
class TestExecutionContextMiddlewareIntegration:
    @pytest.fixture
    def capture_context(self, mocker):
        capture = CaptureContext()
        mocker.patch("linux_mcp_server.commands.execute_with_fallback", side_effect=capture._execute_with_fallback)
        return capture
```

tests/test_server.py
```diff
@@ -334,3 +334,129 @@ async def test_ssh_key_with_no_config_raises(self, setup: SetupFn, mocker):
...
+        from linux_mcp_server.execution_context import get_execution_context
+
+        captured_context = None
+
+        async def capture_context(*_args, **_kwargs):
+            nonlocal captured_context
+            captured_context = get_execution_context()
+            return (0, FREE_OUTPUT, "")
+
+        mocker.patch("linux_mcp_server.commands.execute_with_fallback", side_effect=capture_context)
```
owtaylor:
```suggestion
    async def test_local_policy_sets_local_context(self, mcp_client, mocker, capture_context):
```

src/linux_mcp_server/execution_context.py
```diff
@@ -0,0 +1,33 @@
...
+    try:
+        yield context
+    finally:
+        execution_context_var.reset(token)
+
+
+def get_execution_context() -> ExecutionContext:
+    context = execution_context_var.get()
+    if context is None:
+        raise RuntimeError("No execution context set cannot execute commands")
```
owtaylor: I think this is better in execute_command.py - put the policy there and the mechanism here - this should just return None if there is no execution context.

tests/connection/ssh/test_execution_context.py
```diff
@@ -0,0 +1,95 @@
...
+
+from linux_mcp_server.connection.ssh import execute_command
+from linux_mcp_server.execution_context import ExecutionContext
+from linux_mcp_server.execution_context import get_execution_context
+from linux_mcp_server.execution_context import use_execution_context
+
+
+class TestContextVar:
+    def test_no_context_raises_error(self):
+        with pytest.raises(RuntimeError):
```
owtaylor: Voice of painful past, always use a `match=r""` argument or you might not be catching the runtime error you expect! Also, it's a regexp, so the match should be a raw string:

 1. `match="This is a sentence with a period."` - not what was expected - . matches any char
 2. `match="This is a sentence with a period\."` - SyntaxWarning: invalid escape sequence '\.'
 3. `match=r"This is a sentence with a period\."` - what was actually wanted

tests/connection/ssh/test_execution_context.py
```diff
@@ -0,0 +1,95 @@
...
+            get_execution_context()
```
owtaylor: But this drops out with my suggestion above that this should just return None

tests/connection/ssh/test_execution_context.py
```diff
@@ -0,0 +1,95 @@
...
+
+
+class TestExecuteCommandWithContext:
```
owtaylor: I think my instinct is that these should all be in test_execute_command.py - this is just an aspect of that function we're testing - and we already need to provide a context for the existing tests.

tests/connection/ssh/test_execution_context.py
```diff
@@ -0,0 +1,95 @@
...
+                await execute_command(["echo", "test"])
+
+    # Test remote execution succeeds when allow_ssh_default=True
+    async def test_remote_execution_with_ssh_default_allowed(self, mocker):
+        context = ExecutionContext(allow_ssh_default=True)
+
+        # Mock the connection manager
+        mock_manager = mocker.Mock()
+        mock_manager.execute_remote = mocker.AsyncMock(return_value=(0, "remote output", ""))
+        mocker.patch("linux_mcp_server.connection.ssh._connection_manager", mock_manager)
```
owtaylor: We're now doing this a lot, all the same (I don't think it matters that this is 'remote output "and its "output" elsewhere), so this can just be a fixture defined at the file scope:

```py
@pytest.fixture
def mock_connection_manager(mocker):
    mock_manager = mocker.Mock()
    mock_manager.execute_remote = mocker.AsyncMock(return_value=(0, "output", ""))
    mocker.patch("linux_mcp_server.connection.ssh._connection_manager", mock_manager)

    return mock_manager
```

owtaylor (changes requested):
This looks good to me. There are a couple of minor code organization things and a couple of places you have repeated boilerplate in tests that can be well captured into a file or class scoped fixture. The second one in the reivew more straightforward so read that first. :-)

tests/connection/ssh/test_execution_context.py
panyamkeerthana: The separate test_execution_context.py was to organize the new tests but I agree it makes more sense to keep all execute_command tests together!

src/linux_mcp_server/execution_context.py
panyamkeerthana: Got it, I figured it was better to fail fast in the getter but it makes more sense if execute_command has the fail closed requirement

tests/connection/ssh/test_execution_context.py
panyamkeerthana: Good catch, switched all match parameters to raw strings

tests/connection/ssh/test_execution_context.py
panyamkeerthana: Deleted test_no_context_raises_error

tests/connection/ssh/test_execution_context.py
panyamkeerthana: Thanks for the suggestion, first time using fixtures for this but I see how it cleans up the tests

tests/test_server.py
panyamkeerthana: Thanks for the suggestion! I've implemented this approach with the helper class instead of dict

src/linux_mcp_server/connection/ssh.py
```diff
@@ -110,7 +111,20 @@ async def get_connection(self, host: str) -> asyncssh.SSHClientConnection:
         Raises:
             ConnectionError: If connection fails
         """
-        key = f"{host}"
+        # Get SSH credentials from ExecutionContext if available
+        from linux_mcp_server.execution_context import get_execution_context
```
owtaylor: The import here should be at the toplevel - local imports should be only used if we really need to keep something lazy (like we can avoid pulling in a large stack of dependencies unless a particular config value is set.)

src/linux_mcp_server/server.py
```diff
@@ -319,13 +323,26 @@ async def on_call_tool(self, context: MiddlewareContext, call_next):
...
             if not ssh_key_config:
                 raise RuntimeError("Policy validation error: SSH_KEY action requires ssh_key configuration.")
             logger.debug(f"SSH key override: path={ssh_key_config.path}, user={ssh_key_config.user}")
+            exec_context = ExecutionContext(
+                ssh_key_path=Path(ssh_key_config.path),
+                ssh_key_user=ssh_key_config.user,
+            )
+        else:
+            # all PolicyAction values are handled above
+            raise RuntimeError(f"Unexpected policy action: {action}")
```
owtaylor: This adds an uncovered line - adding `# pragma: no cover` to the `else: ` is fine here. Actually, for can you change this to a `match action:`, with:

```py
            case _:  # pragma: no cover
                raise RuntimeError(f"Unexpected policy action: {action}")
```

src/linux_mcp_server/connection/ssh.py
```diff
@@ -110,7 +111,20 @@ async def get_connection(self, host: str) -> asyncssh.SSHClientConnection:
...
+
+        context = get_execution_context()
+        if context is not None:
+            ssh_key = str(context.ssh_key_path) if context.ssh_key_path else self._ssh_key
+            username = context.ssh_key_user or CONFIG.user
```
owtaylor, line 118: This seems uncovered ... please investigate

tests/connection/ssh/test_execute_command.py
```diff
@@ -1,7 +1,20 @@
...
 from linux_mcp_server.connection.ssh import execute_command
 from linux_mcp_server.connection.ssh import SSHConnectionManager
+from linux_mcp_server.execution_context import ExecutionContext
+from linux_mcp_server.execution_context import use_execution_context
+
+
+@pytest.fixture
+def mock_connection_manager(mocker):
+    """Fixture to mock the SSH connection manager."""
+    mock_manager = mocker.Mock()
```
owtaylor: I would do `Mock(spec=SSHConnectionManager)` here ... it helps catch some mistakes.

tests/connection/ssh/test_execute_command.py
```diff
@@ -25,16 +41,91 @@ async def test_execute_command_local(command, kwargs, expected_rc, expected_out,
 
 async def test_execute_command_remote(mocker):
     """Test that remote execution routes through SSH."""
+    context = ExecutionContext(allow_ssh_default=True)
+
     mock_manager = mocker.Mock(SSHConnectionManager)
     mock_manager.execute_remote = mocker.AsyncMock(return_value=(0, "output", ""))
     mocker.patch("linux_mcp_server.connection.ssh._connection_manager", mock_manager)
```
owtaylor: replace this one with the fixture as well.

owtaylor (changes requested):
Looking good ... just a few comments.

src/linux_mcp_server/connection/ssh.py
panyamkeerthana: My bad the toplevel import was already there from the execute_command() integration, just forgot to remove the redundant local import when adding the SSH key wiring😅

owtaylor (approved):
Looks great!

PR #488, opened 2026-05-29
Handle internal CA certificates for private GitLab servers

owtaylor:
We have two related errors in our gitlab pipelines:
 * Since #475 the mirror job has been failing because it doesn't *just* use `gitlab.cee.redhat.com` through git, it also accesses it through the HTTP API, and while gitlab auto-configures git to trust the server's ssl cert (signed by Red Hat's internal cert), it doesn't do the same with the web API.
 * Since #450, the gatekeeper eval pipeline is failing because glab doesn't trust `gitlab.cee.redhat.com`

I think this PR should fix both. (Alternate fix would be to extend the `SSL_CERT_FILE` handling in `eval-gatekeeper.yml` to cover `glab` and `mirror.yml`, but this way feels cleaner, or at least more gitlab-ci-native.)

### DIFF
--- a/.gitlab/ci/eval-gatekeeper.yml
+++ b/.gitlab/ci/eval-gatekeeper.yml
@@ -29,6 +29,7 @@ download-secure-files:
   image: registry.gitlab.com/gitlab-org/cli:latest
   script:
     - mkdir -p securefiles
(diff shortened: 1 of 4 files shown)


Jazzcort (approved):
LGTM!. I didn't encounter this problem because my runner is actually running on my mac which already included Red Hat CA in the root CA file I guess.

PR #490, opened 2026-06-01
feat(eval): add standard-evals.sh, replace run-eval-models-corp.sh

owtaylor:
Unified eval runner script that configures provider-specific auth, model routing, reasoning effort, and cost overrides for many supported model/provider combinations, then delegates to run-eval.py.

Filenames have a conventional format of:

 MODEL[:FLAG{,FLAG}][@PROVIDER][+VARIANT]

 Where FLAG can be a reasoning effort (none, high, etc) or a quantization
 (bf16, fp8, q4_k_m)

Supports --no-save for quick runs without writing output files and --variant to tag the output file when comparing the effect of some code change.

This replaces run-eval-models-corp.sh. .gitlab/ci/eval-gatekeeper.yml is updated to use standard-evals.sh for all models.

Assisted-by: Claude Opus 4.6
Assisted-by: Gemini 3.1 Pro


eval/gatekeeper/standard-evals.sh
```diff
@@ -0,0 +1,412 @@
...
+
+declare -A spec
+parse_model "$1" spec
+
+candidate_matches_spec() {
+    local -n candidate="$1"
+
+    for k in model provider reasoning quantization; do
+        # If spec has a value, the candidate must match it exactly
+        if [[ -n "${spec[$k]}" && "${spec[$k]}" != "${candidate[$k]}" ]]; then
```
Jazzcort, line 150: I like this mechanism that allows us can provided as less words for the model spec as possible as long as it's unique! 😁

eval/gatekeeper/standard-evals.sh
```diff
@@ -0,0 +1,412 @@
...
+        ;;
+esac
+
+case "$provider" in
+    anthropic)
+        : "${ANTHROPIC_API_KEY:?'api key must be set'}"
+        LINUX_MCP_GATEKEEPER__MODEL="anthropic/$model"
+        ;;
+    llama_cpp)
+        OPENAI_API_KEY=tasty
```
Jazzcort, line 270: Suggesting using "juicy" here. 😆 jk

Jazzcort (approved):
LGTM! Even though reading through this bash script are kinda challenging, I'm very happy with the benefits it brings us. Now, we only need to export those API keys (project id for vertex) as the environment variables without scaffolding the model specs because they get concentrated into just a single string which is amazing!

Note that this script requires Bash version >= 4.0 when associative array got introduced.

owtaylor:
> Note that this script requires Bash version >= 4.0 when associative array got introduced.

I was a bit surprised to realize that MacOS has bash 3, when bash 4 was released 17 years ago (probably a GPL 3 thing) ... if I had known that in advance, maybe I would have just rewritten to Python when the going got rough. But I think it's not a big deal for a devel script when bash 4 is `brew install bash` away.

Jazzcort:
Didn't know the version of bash that macOS is using is from decades ago 😆

Issue #456, opened 2026-05-12
Wrong @field_serializer return type emits non-standard "path" format in JSON Schema


peppekerstens:
**Describe the bug**
The `@field_serializer("unit", "path")` in `linux_mcp_server/models.py` has return type `str | Path | None`, but the serializer always returns `str | None` (never a `Path` object). This causes Pydantic's serialization JSON schema to emit `"format": "path"` (a non-standard JSON Schema format) for **both** fields, including `unit$. Any MCP client that validates tool output schemas sees warnings like:

```
unknown format "path" ignored in schema at path "#/properties/unit/anyOf/1"
unknown format "path" ignored in schema at path "#/properties/path/anyOf/1"
```

**To Reproduce**
1. Connect opencode or another MCP client to linux-mcp-server
2. Run `opencode mcp list`
3. Observe warnings about unknown format "path"

**Expected behavior**
No JSON Schema validation warnings.

**Fix**
```python
@field_serializer("unit", "path")
def serialize_empty_as_null(self, value: str | Path | None) -> str | None:
    return value or None
```

**Environment:**
- linux-mcp-server version: 1.4.1
- Python: 3.14

owtaylor:
Duplicate of #298 , was resolved by #492

PR #492, opened 2026-06-02
Correctly serialize null values

STiFLeR7:
## Summary
Resolves issue #456. Updates serialize_empty_as_null return type annotation to str | None and converts Path to str to avoid generating non-standard format: path in LogEntries schema.

This removes unknown format 'path' warnings on MCP clients like OpenCode.

### DIFF
--- a/tests/test_tool_schemas.py
+++ b/tests/test_tool_schemas.py
@@ -31,3 +31,17 @@ class TestToolSchemaExamples:
         assert param_name in props, f"Parameter '{param_name}' not found in {tool_name}"
         assert "examples" in props[param_name], f"Parameter '{param_name}' in {tool_name} missing examples"
         assert len(props[param_name]["examples"]) > 0, f"Parameter '{param_name}' in {tool_name} has empty examples"
+
+
+def test_log_entries_serialization_schema_no_path_format() -> None:
+    """Verify that LogEntries serialization schema does not contain 'format': 'path' for unit or path."""
+    from linux_mcp_server.models import LogEntries
+
+    schema = LogEntries.model_json_schema(mode="serialization")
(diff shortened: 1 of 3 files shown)


major (dismissed):
Seems reasonable to me. Thanks!

tests/test_tool_schemas.py
```diff
@@ -31,3 +31,17 @@ async def test_parameter_has_examples(self, tool_name: str, param_name: str) ->
         assert param_name in props, f"Parameter '{param_name}' not found in {tool_name}"
         assert "examples" in props[param_name], f"Parameter '{param_name}' in {tool_name} missing examples"
         assert len(props[param_name]["examples"]) > 0, f"Parameter '{param_name}' in {tool_name} has empty examples"
+
+
+def test_log_entries_serialization_schema_no_path_format() -> None:
```
samdoran: This test isn't validating the behavior, just the schema. It's possible to break the behavior and this test will still pass so long as the type hint doesn't change.

Something like this is a better test of the behavior.
```python
def test_log_entries_null_value_serialization():
    """Assert that null values are properly serialized."""

    log_entry = LogEntries(entries=["log"])
    model = log_entry.model_dump()

    assert model["unit"] is None
    assert model["path"] is None
```

samdoran (changes requested):
Tests should go in test files that correspond to their location in the source tree so they are easier to locate. This should go in `test_models.py` which currently doesn't exist.

samdoran:
This is a really good fix and a subtle bug. Thanks for fixing it!

samdoran (approved):

subpop (approved):
LGTM. Thanks @STiFLeR7 for the fix and @samdoran for the review/cleanup!

PR #494, opened 2026-06-02
refactor(gatekeeper): Drop LiteLLM

subpop:
Drop LightLLM and replace with custom, in-tree clients. The combination of custom clients maintain support for OpenAI compatible providers with both Responses and Chat Completion endpoints, Anthropic Messages API, and Gemini’s generateContent. Both direct API and GCP/Vertex AI backends are supported for APIs that are provided by Vertex AI.

There is a slight regression in that we no longer support OpenRouter’s native API and custom settings.


owtaylor:
99% of the difficulty here is figuring out what the config schema should be!

We have prior art with authentication:

```py
class Config(BaseSettings):
    auth: AuthConfig | None = None

class AuthConfig(BaseSettings):
    """Authentication configuration."""

    provider: AuthProvider | None = None
    google: GoogleAuthConfig | None = None
    github: GitHubAuthConfig | None = None
    jwt: JWTAuthConfig | None = None
    introspection: IntrospectionAuthConfig | None = None

class GoogleAuthConfig(BaseSettings):
    """Google OAuth authentication configuration."""

    client_id: str
    client_secret: SecretStr

```

So one writes:

```sh
LINUX_MCP_AUTH__PROVIDER=google
LINUX_MCP_AUTH__GOOGLE__CLIENT_ID="<...>"
LINUX_MCP_AUTH__GOOGLE__SECRET="<...>"
```

Or potentially:

```sh
LINUX_MCP_AUTH='{ "provider": "google", "google": { "client_id": "<...>", "client_secret": "<...>" }}'
```

Which is slightly different than what you did here because the google specific options move under the nested google object, while you instead document what settings are specific to particular providers or backends.

The extra factors for the gatekeeper model are:

  * There are some options that are shared between different models (temperature, reasoning_effort)
  * We have the split between the "provider" and "backend"

I'm also not in love with the name "provider" and "backend" - how is openrouter a "backend" when it routes things to somewhere else? I think "inference_gateway" or "gateway" would be a pretty descriptive name, but when I tried to figure out how that maps into config, I couldn't figure out a good handling for "openai compatible".

The proposal I'm going to make is that we just have a provider - which we autodetect where possible (`gpt-*`, `claude-*`)

 * openai
 * anthropic
 * openai_compatible
 * vertex_ai
 * openrouter

And some provides have provider-specific options. Bunch of worked examples:

```sh
LINUX_MCP_GATEKEEPER__MODEL="gpt-5.4"
LINUX_MCP_GATEKEEPER__PROVIDER="openai"  # would be autodetected

LINUX_MCP_GATEKEEPER__MODEL="openai/gpt-5.4"
LINUX_MCP_GATEKEEPER__PROVIDER="openrouter"

LINUX_MCP_GATEKEEPER__MODEL="gemini-3.5-flash"
LINUX_MCP_GATEKEEPER__REASONING_EFFORT="minimal"
LINUX_MCP_GATEKEEPER__PROVIDER="vertex_ai"
LINUX_MCP_GATEKEEPER__VERTEX_AI__PROJECT="rhel-lightspeed"

LINUX_MCP_GATEKEEPER__MODEL="qwen/qwen3.5-8b"
LINUX_MCP_GATEKEEPER__TEMPERATURE="0"
LINUX_MCP_GATEKEEPER__PROVIDER="openrouter"
LINUX_MCP_GATEKEEPER__OPENROUTER__QUANTIZATION="fp4"

LINUX_MCP_GATEKEEPER__MODEL="qwen/qwen3.5-8b"
LINUX_MCP_GATEKEEPER__PROVIDER="openai_compat"
LINUX_MCP_GATEKEEPER__OPENAI_COMPAT__BASE_URL="http://localhost:8080/v1"
LINUX_MCP_GATEKEEPER__OPENAI_COMPAT__TEMPLATE_KWARGS='{ "enable_thinking": true }'
```

The last is clunky, but clear.  Maybe better as "generic" rather than "openai_compat"?

What about things like needing to use the anthropic API when using claude models through vertex_ai? My feeling is that we just autodetect and implement that internally. Because it's not *exactly* the same - there are often a few quirks. ([Anthropic docs](https://platform.claude.com/docs/en/build-with-claude/claude-on-vertex-ai)]). We *could* have a config parameter `LINUX_MCP_GATEKEEPER__API=anthropic` - but I'm not sure there is any real use case.

What if different endpoints want to share the same config - let's say that some other backend supported template_kwargs. Is that an argument for putting this type of thing on the base class - and documenting restrictions. Perhaps - I'm convincible - but IMO nesting parameters is clearer. We can always duplicate. I definitely want to keep something like the vertex AI location nested.

eval/gatekeeper/standard-evals.sh
```diff
@@ -316,30 +317,37 @@ case "$provider" in
         : "${VERTEXAI_PROJECT:?'project must be set'}"
         uv_args+=("--extra" "gcp")
         max_parallel=50
+        LINUX_MCP_GATEKEEPER__BACKEND=vertex
+        vertex_location="${VERTEXAI_LOCATION:-global}"
+        vertex_openapi_base="https://aiplatform.googleapis.com/v1/projects/${VERTEXAI_PROJECT}/locations/${vertex_location}/endpoints/openapi"
```
owtaylor: I think we can do this generically for vertexai + openai API instead of having it here. Since we do a very similar thing for the google api and the anthropic api. (Do we need to support API base overrides for vertex_ai ... there are likely use cases, but maybe it can wait until someone asks for it?)

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,86 @@
...
+from linux_mcp_server.config import CONFIG
+from linux_mcp_server.config import GatekeeperBackend
+from linux_mcp_server.gatekeeper.gcp_auth import get_gcp_access_token
+from linux_mcp_server.gatekeeper.gcp_auth import get_gcp_location
+from linux_mcp_server.gatekeeper.gcp_auth import get_gcp_project
+from linux_mcp_server.gatekeeper.http_utils import ANTHROPIC_API_URL
+from linux_mcp_server.gatekeeper.http_utils import ANTHROPIC_API_VERSION
+from linux_mcp_server.gatekeeper.http_utils import ANTHROPIC_DEFAULT_MAX_TOKENS
+from linux_mcp_server.gatekeeper.http_utils import anthropic_thinking_block
+from linux_mcp_server.gatekeeper.http_utils import ANTHROPIC_VERTEX_VERSION
```
owtaylor: I wouldn't expect things these things to be in `http_utils.py`

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -121,8 +103,6 @@ def get_model() -> str:
 """
 
 
-# Maximum number of completion tokens (including reasoning)
-GATEKEEPER_MAX_TOKENS = 8000
```
owtaylor, line 125: Where did this go?

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -247,24 +191,13 @@ def __init__(self, message: str, *, stats: GatekeeperStats | None = None):
...
-            input_cost_per_token=CONFIG.gatekeeper.cost[0], output_cost_per_token=CONFIG.gatekeeper.cost[1]
-        )
-    else:
-        custom_cost_per_token = None
-
-    try:
-        return completion_cost(response, custom_cost_per_token=custom_cost_per_token)
-    except Exception:
+def _compute_cost(prompt_tokens: int, completion_tokens: int, *, usage_cost: float | None = None) -> float:
+    if usage_cost is not None:
```
owtaylor: Looks like we lost cost for everything but openrouter and manually configured models?

src/linux_mcp_server/gatekeeper/http_utils.py
```diff
@@ -0,0 +1,160 @@
+"""Shared helpers for gatekeeper HTTP clients."""
+
+import os
+
+from typing import Any
+from urllib.parse import urlparse
+
+import requests
```
owtaylor: This needs to be async using requests - it doesn't matter much for the single user case, but we want to support shared servers as well - we don't want validate_script() calls to block for several seconds.

src/linux_mcp_server/gatekeeper/openai_client.py
```diff
@@ -0,0 +1,113 @@
...
+
+
+def complete_openai(prompt: str, *, timeout: int = DEFAULT_TIMEOUT_SECONDS) -> GatekeeperCompletion:
+    base_url = get_openai_base_url()
+    headers = {
+        **openai_auth_headers(),
+        "Content-Type": "application/json",
+    }
+
+    # Try the Responses API first, falling back to Chat Completions if it's not available.
```
owtaylor: Don't like that this happens on evey validate_script.. Maybe we try *just* supporting responses, and add a config key later if we need it? (Could have `"api": "openai_completions" or something) Most providers seem to have responses at least in beta / recently .. and if chat completions is more stable, we won't find out if we're starting off with responses.

owtaylor (commented):
Some comments. Not a full review to let you figure out what you want for config and implement.

subpop:
I rewrote the config, added support for cost back in, switched to async HTTP client, restored max token capping, and removed chat completion API. They are all additional commits on this PR, since we squash on merge anyway. Hopefully they help with the review.

docs/config-reference.md
```diff
@@ -58,13 +58,21 @@ These are used when `LINUX_MCP_TOOLSET` is set to `run_script` or `both`.
...
+| `--gatekeeper.model`<br>`LINUX_MCP_GATEKEEPER__MODEL` | _(none)_ | Required: provider-native model ID (e.g. `gpt-5.4`, `claude-sonnet-4-6`, `gemini-2.0-flash`, `openai/gpt-oss-120b` for OpenRouter) |
+| `--gatekeeper.reasoning_effort`<br>`LINUX_MCP_GATEKEEPER__REASONING_EFFORT` | _(model specific)_ | Reasoning effort (`none`, `minimal`, `low`, `medium`, `high`, `xhigh`). Not all values are supported for all models. |
+| `--gatekeeper.structured_output`<br>`LINUX_MCP_GATEKEEPER__STRUCTURED_OUTPUT` | `True` | Whether to use structured JSON output from the model |
+| `--gatekeeper.temperature`<br>`LINUX_MCP_GATEKEEPER__TEMPERATURE` | 0.0 | Temperature to use for the model |
+| `--gatekeeper.cost`<br>`LINUX_MCP_GATEKEEPER__COST` | _(none)_ | Optional per-token cost override for gatekeeper accounting (`input$/token:output$/token`). When unset, cost is estimated from [models.dev](https://models.dev) pricing (with a vendored fallback), except OpenRouter which uses API-reported `usage.cost` when available. Local OpenAI-compatible inference (e.g. llama.cpp on localhost) is reported as `$0`. |
```
owtaylor: I don't necessarily see the value of the vendored fallback. We care about cost really only when doing evals (as a way to keep on eye on spend / so we can give user's a sense of the cost) - and we'd expect `models.dev` to be reachable. If it's down, I'm not sure falling back to an old copy that we vendored at some point is the best way to handle that. In what circumstances do you expect the vendored fallback to be used?

> Local OpenAI-compatible inference (e.g. llama.cpp on localhost) is reported as `$0`.

Too complex. I don't like the idea that if we don't have cost information we choose between $0 and default number based on hostname. I think $0 should uniformly mean "no cost available" and we can do things like omit the output from the web viewer if we see a cost of $0.

docs/config-reference.md
```diff
@@ -58,13 +58,21 @@ These are used when `LINUX_MCP_TOOLSET` is set to `run_script` or `both`.
...
+| `--gatekeeper.openai.base_url`<br>`LINUX_MCP_GATEKEEPER__OPENAI__BASE_URL` / `OPENAI_API_BASE` | `https://api.openai.com/v1` | OpenAI provider: API base URL |
+| `--gatekeeper.openai.template_kwargs`<br>`LINUX_MCP_GATEKEEPER__OPENAI__TEMPLATE_KWARGS` | _(none)_ | OpenAI provider: extra chat-template arguments (e.g. llama.cpp `enable_thinking`), sent as `chat_template_kwargs` |
```
owtaylor: `enable_thinking` is not llama.cpp specific ... it's a standard parameter used in various published templates. " Maybe "extra chat-template arguments sent as ` chat_template_kwargs`. Example `{ "enable_thinking": false }` can be used with some models to disable reasoning when using llama.cpp, which doesn't support a reasoning_effort parameter."

docs/guarded-command-execution.md
```diff
@@ -106,17 +106,56 @@ LINUX_MCP_TOOLSET=run_script
...
+`GOOGLE_API_KEY`, or `OPENROUTER_API_KEY`). For the `vertex_ai` provider, install the `gcp` extra and configure
+`GOOGLE_APPLICATION_CREDENTIALS`, `VERTEXAI_PROJECT`, and `VERTEXAI_LOCATION`.
 
-Example:
+Example (OpenAI):
 
 ```sh
-LINUX_MCP_GATEKEEPER__MODEL=openai/chatgpt-5.2
+LINUX_MCP_GATEKEEPER__PROVIDER=openai
+LINUX_MCP_GATEKEEPER__MODEL=gpt-5.4
```
owtaylor, line 118: I don't think it makes sense to implement autodetection of provider but then treat it like "you aren't supposed to use this". Either we should omit it, or this should be more like


```suggestion
# LINUX_MCP_GATEKEEPER__PROVIDER=openai   # autodetected
LINUX_MCP_GATEKEEPER__MODEL=gpt-5.4
```

docs/guarded-command-execution.md
```diff
@@ -106,17 +106,56 @@ LINUX_MCP_TOOLSET=run_script
...
+```sh
+LINUX_MCP_GATEKEEPER__PROVIDER=vertex_ai
+LINUX_MCP_GATEKEEPER__MODEL=gemini-3.1-pro-preview
+LINUX_MCP_GATEKEEPER__VERTEX_AI__PROJECT=my-gcp-project
+GOOGLE_APPLICATION_CREDENTIALS=<path-to-service-account.json>
+```
+
+**Gatekeeper cost estimation**
+
+Eval runs and gatekeeper stats report an estimated dollar cost per check. Resolution order:
```
owtaylor: I don't think this belongs in this file - it's an advanced topic. Maybe move to config-reference.md. (If you are just learning about the gatekeeper, you are going to make zero sense of "Eval runs and gatekeeper stats" - in fact, gatekeeper stats are entirely an internal thing.

docs/guarded-command-execution.md
```diff
@@ -106,17 +106,56 @@ LINUX_MCP_TOOLSET=run_script
...
+
+1. **API-reported cost** — OpenRouter `usage.cost` when present.
```
owtaylor:
```suggestion
1. **API-reported cost** — If the inference provider reports a cost, that will be used. (Currently supported for OpenRouter)
```

docs/guarded-command-execution.md
```diff
@@ -106,17 +106,56 @@ LINUX_MCP_TOOLSET=run_script
...
+2. **Config override** — `LINUX_MCP_GATEKEEPER__COST` as `input$/token:output$/token` (useful for Vertex MaaS eval models).
```
owtaylor: I don't think "Vertex MaaS eval models" will make sense to anybody. I don't think you need to say that. Just

 `LINUX_MCP_GATEKEEPER__COST=<$ / input token>:<$ / output token>` - e.g.`0.1e-6:0.4e-6`

(I'd like to change this to be $ / million tokens, which is industry standard and less confusing, but not in this commit.)

docs/guarded-command-execution.md
```diff
@@ -106,17 +106,56 @@ LINUX_MCP_TOOLSET=run_script
...
+3. **models.dev** — runtime pricing lookup with a vendored snapshot fallback when the network is unavailable.
+4. **Local inference** — OpenAI-compatible providers pointed at `localhost` / `127.0.0.1` (e.g. llama.cpp) report `$0`.
+5. **Fallback** — hardcoded rates for known eval models, then a conservative cloud default.
```
owtaylor: See comments above.

eval/gatekeeper/README.md
```diff
@@ -64,11 +64,21 @@ Runs test cases through the gatekeeper and reports results.
 
 ```bash
 # Set the gatekeeper model
-export LINUX_MCP_GATEKEEPER__MODEL="openrouter/anthropic/claude-3.5-sonnet"
+export LINUX_MCP_GATEKEEPER__PROVIDER="anthropic"
+export LINUX_MCP_GATEKEEPER__MODEL="claude-sonnet-4-6"
+export ANTHROPIC_API_KEY="..."
```
owtaylor, line 69: Again, we should either consider provider inference as supported, or don't implement it.

eval/gatekeeper/standard-evals.sh
```diff
@@ -237,7 +236,6 @@ LINUX_MCP_GATEKEEPER__COST=""
 case "$model" in
     claude-*)
         if [[ $reasoning == "none" ]] ; then
-            # LiteLLM doesn't support reasoning_effort=none for Anthropic models
             LINUX_MCP_GATEKEEPER__REASONING_EFFORT=
```
owtaylor: The old comment may no longer be right in detail, but a comment is still needed. I think we want "low" here rather than no reasoning effort parameter. (https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking#adaptive-thinking-with-the-effort-parameter)

eval/gatekeeper/standard-evals.sh
```diff
@@ -320,30 +317,25 @@ case "$provider" in
         : "${VERTEXAI_PROJECT:?'project must be set'}"
         uv_args+=("--extra" "gcp")
         max_parallel=50
+        LINUX_MCP_GATEKEEPER__PROVIDER=vertex_ai
+        LINUX_MCP_GATEKEEPER__VERTEX_AI__PROJECT="${VERTEXAI_PROJECT}"
```
owtaylor: Doesn't make sense that we support `VERTEXAI_PROJECT` in the code but copy it over here.

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,89 @@
...
+        ReasoningEffort.MINIMAL: 1024,
+        ReasoningEffort.LOW: 4096,
+        ReasoningEffort.MEDIUM: 8192,
+        ReasoningEffort.HIGH: 16384,
+        ReasoningEffort.XHIGH: 32768,
+    }
+    budget = budget_by_effort.get(reasoning_effort)
+    if budget is None:
+        return None
+    return {"type": "enabled", "budget_tokens": budget}
```
owtaylor: This is not supported for the most recent models (except for Claude Haiku, where it's the only thing supported.)
https://platform.claude.com/docs/en/build-with-claude/adaptive-thinking#supported-models
https://platform.claude.com/docs/en/build-with-claude/extended-thinking#supported-models

My suggested is:

 * ReasoningEffort.DEFAULT - don't provide a thinking block.
 * ReasoningEffort.NONE - use `thinking={"type": "disabled"}` (Will error out for some models.)
 * ReasoningEffort.<Other> - use `thinking={"type": "adaptive"} output_config={"effort": "<effort>"}` (Will error out for Haiku and old models.)

This provides no way to enable thinking for Haiku and no way to control the effort for Opus <= 4.5, but I think that's fine for our needs. I'd rather not go to having to parse the model string and do different behaviors for different versions.

src/linux_mcp_server/gatekeeper/check_run_script.py
```diff
@@ -290,44 +219,33 @@ async def check_run_script_with_stats(
...
-        raise GatekeeperException("Timeout calling gatekeeper model")
-    assert isinstance(response, ModelResponse)
+    except asyncio.TimeoutError:
+        raise GatekeeperException("Timeout calling gatekeeper model") from None
+
+    cost, cost_source = compute_cost(
+        completion.prompt_tokens,
+        completion.completion_tokens,
+        usage_cost=completion.usage_cost,
+    )
```
owtaylor: I think it's worth the complexity to make sure that we don't call this unless we're actually interested in the stats - as a server that will be hosted somewhere, we shouldn't externally download a file from `models.dev` unless we actually are using the result.

Various ways to set it up would be fine - one way would be to switch to `check_run_script(include_stats: bool = False)` and use a overload to make the return type vary based on that parameter.

src/linux_mcp_server/gatekeeper/gemini_client.py
```diff
@@ -0,0 +1,84 @@
...
+    api_key = os.environ.get("GOOGLE_API_KEY") or os.environ.get("GEMINI_API_KEY")
+    if not api_key:
+        raise ValueError("GOOGLE_API_KEY or GEMINI_API_KEY is required for Gemini direct backend.")
+    return api_key
+
+
+def _gemini_thinking_level(reasoning_effort: ReasoningEffort | None) -> str | None:
+    if reasoning_effort is None or reasoning_effort in {ReasoningEffort.NONE, ReasoningEffort.DEFAULT}:
+        return None
+    mapping = {
```
owtaylor: Mapping `ReasoningEffort.NONE` to "no thinking config" seems wrong to me - no thinking config will give HIGH thinking for many models.

Looking at:
https://ai.google.dev/gemini-api/docs/generate-content/thinking#generating-content-with-thinking
the only models where you can disable thinking are the 2.5 flash series where you use `thinkingBudget = 0` - otherwise, thinkingLevel="minimal" is the closest. 

Two reasonable options are:
 * ReasoningEffort.NONE - an error
 * ReasoningEffort.NONE => "MINIMAL" (my recommendation)

src/linux_mcp_server/gatekeeper/http_utils.py
```diff
@@ -0,0 +1,42 @@
...
+    response = await HTTP_CLIENT.post(url, headers=headers, json=body, timeout=timeout)
+    if not response.is_success:
+        raise GatekeeperHTTPError(provider, response.status_code, response.text)
+    return response.json()
+
+
+def normalize_model_id(model: str) -> str:
+    for prefix in ("openai/", "anthropic/", "vertex_ai/", "gemini/"):
+        if model.startswith(prefix):
+            return model[len(prefix) :]
```
owtaylor: I'm confused what the models names are now! Is the model name for `vertex_ai` `claude-sonnet-4-6` or `anthropic/claude-sonnet-4-6` (or `gemini/claude-sonnet-4-6` :smile:) ... suggestion:

 * call this something specific not just "normalize"
 * have a rule that what the user passes in gets passed literally to the platform. So, strip prefixes if needed to match on models.dev - but don't normalize when passing to the platform.

src/linux_mcp_server/gatekeeper/llm.py
```diff
@@ -0,0 +1,53 @@
...
+logger = logging.getLogger("linux-mcp-server")
+
+
+def _infer_provider_from_model(model: str) -> GatekeeperProvider:
+    if model.startswith("openrouter/"):
+        return GatekeeperProvider.OPENROUTER
+    if model.startswith("anthropic/") or model.startswith("claude"):
+        return GatekeeperProvider.ANTHROPIC
+    if model.startswith("gemini"):
+        return GatekeeperProvider.GEMINI
```
owtaylor: The `openrouter/` and `anthropic/`pieces don't match the intent of my suggestion for provider inference. I wasn't suggesting that we support litellm-style strings, I was suggesting that *if* the model string is of certain well known forms (like claude-*) we infer the provider, without adjusting the model string.

src/linux_mcp_server/gatekeeper/openrouter_client.py
```diff
@@ -0,0 +1,111 @@
...
+        "quantization": CONFIG.gatekeeper.openrouter.quantization,
+        "template_kwargs": CONFIG.gatekeeper.openrouter.template_kwargs,
+    }
+
+
+def _build_chat_completions_body(prompt: str, *, max_tokens: int) -> dict[str, Any]:
+    openrouter = _openrouter_config()
+    provider: dict[str, Any] = {"require_parameters": True}
+    if openrouter["quantization"]:
+        provider["quantizations"] = [openrouter["quantization"]]
```
owtaylor: It feels here that `_openrouter_config()` turns a struct into a dict then we say "oh, that wasn't the dict we need, let's turn it into another dict" - I think just get rid of `_openrouter_config()` and put the logic inline in this function.

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,186 @@
...
+
+
+MODELS_DEV_API_URL = "https://models.dev/api.json"
+FALLBACK_PRICING_PATH = Path(__file__).resolve().parent / "data" / "models_dev_fallback.json"
+
+FALLBACK_COST_PER_MTOK: dict[str, tuple[float, float]] = {
+    "gemma-4-26b-a4b-it-maas": (0.15, 0.60),
+    "gpt-oss-20b-maas": (0.07, 0.25),
+    "gpt-oss-120b-maas": (0.09, 0.36),
+}
```
owtaylor: Cost is pretty much irrelevant unless we're going through standard-evals.sh. Let's just leave this kind of hardcoding there ... that keeps the selection of models and providers we care about in one place.

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,186 @@
...
+
+DEFAULT_CLOUD_COST_PER_MTOK = (3.0, 15.0)
```
owtaylor: See my comments elsewhere on this ... if we don't know the cost, we shouldn't just make stuff up.

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,186 @@
...
+    match provider:
+        case GatekeeperProvider.GEMINI:
+            return "google"
+        case GatekeeperProvider.VERTEX_AI:
+            normalized = normalize_model_id(model)
+            if normalized.startswith("claude"):
+                return "anthropic"
+            if normalized.startswith("gemini"):
+                return "google"
+            return "openai"
```
owtaylor: That doesn't seem like a reasonable default ... the only openai models on VERTEX_AI are the gpt-oss-* models, which presumably priced without any coordination with the pricing via openai.

In fact, there's a "google-vertex" provider on models.dev with at least many of the things we want `curl https://models.dev/api.json | jq  '.["google-vertex"].models | keys'`

src/linux_mcp_server/gatekeeper/schema.py
```diff
@@ -0,0 +1,107 @@
+"""JSON schema helpers for gatekeeper structured output across LLM providers."""
+
+from copy import deepcopy
+from typing import Any
+
+
+_GATEKEEPER_STATUS_VALUES = [
```
owtaylor: I don't see that this needs to avoid importing the real types, can we just do

```
_GATEKEEPER_STATUS_VALUES = [m.value for m in GatekeeperStatus]
```

and avoid duplication?

src/linux_mcp_server/gatekeeper/schema.py
```diff
@@ -0,0 +1,107 @@
...
+        "type": "object",
+        "properties": {
+            "status": {
+                "type": "string",
+                "enum": _GATEKEEPER_STATUS_VALUES,
+                "description": "Gatekeeper verdict for the script.",
+            },
+            "detail": {
+                "type": "string",
+                "description": "Short explanation when status is not OK.",
```
owtaylor: Add descriptions to these fields in the schema will change the gatekeeper performance and evals. I'd rather it wasn't mixed into a giant unrelated patch.

src/linux_mcp_server/models.py
```diff
@@ -198,3 +200,28 @@ class LogEntries(BaseModel):
     @field_serializer("unit", "path")
     def serialize_empty_as_null(self, value: str | Path | None) -> str | None:
         return str(value) if value else None
+
+
+class GatekeeperCompletion(BaseModel):
```
owtaylor: Let's keep `models.py` for the types we use in our tool calls - I don't think it makes sense to have a file that is "every type that derives from BaseModel". I find it quite confusing to have DiskUsage which is about our tool calls and disk usage on the target system then Usage - which is a type related to calling an LLM. I'd keep these added types within the gatekeeper directory. If they don't seem clearly related to a single file, then we can add `gatekeeper/types.py`.

src/linux_mcp_server/gatekeeper/openai_client.py
```diff
@@ -0,0 +1,98 @@
...
+    reasoning = _openai_reasoning_block(CONFIG.gatekeeper.reasoning_effort)
+    if reasoning is not None:
+        body["reasoning"] = reasoning
+    return body
+
+
+def _extract_responses_text(response: dict[str, Any]) -> str:
+    output_text = response.get("output_text")
+    if isinstance(output_text, str) and output_text.strip():
+        return output_text.strip()
```
owtaylor: [Docs](https://developers.openai.com/api/reference/resources/responses/methods/retrieve#(resource)%20responses%20%3E%20(model)%20response%20%3E%20(schema)%20%3E%20(property)%20output_text) say "SDK-only convenience property that contains the aggregated text output from all output_text items in the output array, if any are present. Supported in the Python and JavaScript SDKs." - so I think it can never be in the HTTP response.

AGENTS.md
```diff
@@ -29,6 +29,7 @@ make verify                  # All checks (required before commit)
 - Use pytest-mock (`mocker` fixture) for mocking instead of `unittest.mock` imports
 - Use `autospec=True` when patching; `spec=<object>` with Mock
 - 100% patch coverage for new code
+- Check the project config and test runners and only decorate async functions if the ascyio_mode is not set to auto.
```
owtaylor:
```suggestion
- Do not use `@pytest.mark.asyncio` for tests. It's not necessary because the project uses `asyncio_mode = "auto"`
```

owtaylor (changes requested):
OK, this is really comprehensive now! Mostly looks great. Summary of some of the highlights of the review below:

 * I'd like to simplify the cost calculations - from api, from models.dev, otherwise $0 == don't know. Vendored fallback is going to be a pain to maintain and doesn't seem that commonly useful.
 * I'd like to be stricter about model names and say - "model names are passed to the provider, unmodified". There might be cases where the provider supports aliases, but let's not create our own aliases.

Most other things here are details. I haven't reviewed the tests in detail yet. I do want to stick to "strict" coverage - newly added code should be 100% including error return paths. Otherwise, it's hard to tell if we accidentally miss something that is important to be covered.

docs/config-reference.md
subpop: Yea, it's demonstrating an example, but I can see the thought process behind removing "llama.cpp" specifically to avoid wrongly insinuating that it's llama.cpp specific.

subpop:
> OK, this is really comprehensive now! Mostly looks great. Summary of some of the highlights of the review below:
> 
> * I'd like to simplify the cost calculations - from api, from models.dev, otherwise $0 == don't know. Vendored fallback is going to be a pain to maintain and doesn't seem that commonly useful.
> * I'd like to be stricter about model names and say - "model names are passed to the provider, unmodified". There might be cases where the provider supports aliases, but let's not create our own aliases.
> 
> Most other things here are details. I haven't reviewed the tests in detail yet. I do want to stick to "strict" coverage - newly added code should be 100% including error return paths. Otherwise, it's hard to tell if we accidentally miss something that is important to be covered.

I'm not sure I'm convinced of the utility of autodetecting the provider. If we use "bare" model names (for example, "claude-4-6-haiku"), would that autodetect to using the direct Anthropic API provider or the model hosted on Vertex AI? I feel like I would rather always require the "provider" option. If we wanted to follow the pattern of LiteLLM and use "<provider>/<model>" formatting, we also could avoid heuristically detecting the provider. I don't like that format either, personally. So I'd prefer to fall back to just explicitly requiring "provider" and "model".

Good point about the pricing fallback. I just kinda assumed a graceful fallback approach, but since it's for internal evaluations, we don't really need to carry that complexity.

docs/config-reference.md
subpop: Does it ever make sense to override cost since it's meant for evaluation purposes only? Why don't we just remove this from the public CLI entirely?

AGENTS.md
subpop: Have you found that this instruction, as written, isn't effective?

owtaylor:
> I'm not sure I'm convinced of the utility of autodetecting the provider. If we use "bare" model names (for example, "claude-4-6-haiku"), would that autodetect to using the direct Anthropic API provider or the model hosted on Vertex AI? I feel like I would rather always require the "provider" option. If we wanted to follow the pattern of LiteLLM and use "/" formatting, we also could avoid heuristically detecting the provider. I don't like that format either, personally. So I'd prefer to fall back to just explicitly requiring "provider" and "model".

I'm 100% fine with requiring the provider always. autodetecting the provider seemed like a convenience but does make things more squishy and weird ... here's another argument for not doing it: if we autodetect the provider then the validity of the configuration depends on the autodetection. `model: gpt-5.4; openai: { template_kwargs: {...} }` is valid `model: xpt-5.4; openai: { template_kwargs: {...} }` is not.

AGENTS.md
owtaylor: I haven't had any issues with Claude adding `@pytest.mark.asyncio` so I'm not sure how I'd even test it. But when I read the original form:

 * I didn't understand it until I did some research to figure out what it was about - not sure if it would be more clear to models.
 * It seemed weird to tell the model to look up things in the code base, when the answer is static and known.

So I tried to find a phrasing that was more specific and conservative of tokens.

docs/guarded-command-execution.md
subpop: I just removed the mention of cost entirely from the user-facing docs. It's an edge case and one that doesn't need to be documented up front.

eval/gatekeeper/standard-evals.sh
subpop: Are you referring to the `os.environ` lookup in `gcp_auth.py`?

subpop: The `get_gcp_project` logic has a fallthrough; it uses the `LINUX_MCP_GATEKEEPER__VERTEX_AI__PROJECT` value first, then falls back to `VERTEXAI_PROJECT` and `GOOGLE_CLOUD_PROJECT`. This line here is ensuring that the project gets set through the first path of `get_gcp_project`.

```python
    project = (
        (vertex_ai.project if vertex_ai else None)
        or os.environ.get("VERTEXAI_PROJECT")
        or os.environ.get("GOOGLE_CLOUD_PROJECT")
    )
```

AGENTS.md
subpop: OK. I like the more explicit language in your edit. I was just curious if you had a reason for the suggested change.

src/linux_mcp_server/gatekeeper/http_utils.py
subpop: Removed as part of the removal of provider detection.

eval/gatekeeper/standard-evals.sh
owtaylor: What i'm saying is that VERTEXI_PROJECT is already in the environment, and the code will pick it up, so why are we bothering to set    LINUX_MCP_GATEKEEPER__VERTEX_AI__PROJECT ?

src/linux_mcp_server/models.py
subpop: I agree. Putting models into `models.py` was a convention of the project, so I've just been following that pattern. I personally like types to live as close to the functions that use them. I'll move this type definition into 'gatekeeper' somewhere.

src/linux_mcp_server/gatekeeper/schema.py
subpop: I see. The LiteLLM-generated schema didn't include descriptions.

src/linux_mcp_server/gatekeeper/openai_client.py
subpop: Good catch! Should we mirror support for building 'output_text' into our SDK then too?

subpop: > Good catch! Should we mirror support for building 'output_text' into our SDK then too?

No we should not.

eval/gatekeeper/standard-evals.sh
subpop: Yea, I understand your reasoning. I'll drop it; the eval script is all internal stuff and doesn't need to follow conventions for how to use/how to configure.

src/linux_mcp_server/gatekeeper/llm.py
```diff
@@ -0,0 +1,46 @@
...
+class GatekeeperCompletion(BaseModel):
+    text: str
+    prompt_tokens: int = 0
+    completion_tokens: int = 0
+    usage_cost: float | None = None
+
+
+async def complete_gatekeeper(prompt: str, *, max_tokens: int) -> GatekeeperCompletion:
+    # Deferred imports avoid circular dependency: clients import GatekeeperCompletion from this module.
+    from linux_mcp_server.gatekeeper.anthropic_client import complete_anthropic
```
owtaylor: It would read better to me if these were embedded within the specific branches of the case statement.

src/linux_mcp_server/gatekeeper/openrouter_client.py
```diff
@@ -0,0 +1,101 @@
...
+
+def _build_chat_completions_body(prompt: str, *, max_tokens: int) -> dict[str, Any]:
+    assert CONFIG.gatekeeper is not None
+    openrouter = {
+        "quantization": CONFIG.gatekeeper.openrouter.quantization if CONFIG.gatekeeper.openrouter else None,
+        "template_kwargs": CONFIG.gatekeeper.openrouter.template_kwargs if CONFIG.gatekeeper.openrouter else {},
+    }
+    provider: dict[str, Any] = {"require_parameters": True}
+    if openrouter["quantization"]:
+        provider["quantizations"] = [openrouter["quantization"]]
```
owtaylor: Avoid the dict entirely!

```suggestion
    provider: dict[str, Any] = {"require_parameters": True}
    quantization = CONFIG.gatekeeper.openrouter.quantization if CONFIG.gatekeeper.openrouter else None
    quantization:
        provider["quantizations"] = [quantization]
```

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,172 @@
...
+        if not isinstance(cost, dict):
+            continue
+        input_mtok = cost.get("input")
+        output_mtok = cost.get("output")
+        if isinstance(input_mtok, (int, float)) and isinstance(output_mtok, (int, float)):
+            return float(input_mtok), float(output_mtok)
+    return None
+
+
+def is_local_inference() -> bool:
```
owtaylor: Can we get rid of this (and CostSource overall?) - tracking a distinction between a rate of 0.0 coming from "local" and "fallback" seems not worth the complexity.

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,172 @@
...
+        case GatekeeperProvider.GEMINI:
+            return "google"
+        case GatekeeperProvider.VERTEX_AI:
+            return "google-vertex"
+        case _:
+            return provider.value
+
+
+@cache
+def _load_models_dev_pricing() -> dict[str, Any]:
```
owtaylor: Maybe _load_models_dev_payload() or something?

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,172 @@
...
+    """Loads pricing from the models.dev API. Returns an empty dict if unavailable."""
+    try:
+        with httpx.Client(timeout=10) as client:
+            response = client.get(MODELS_DEV_API_URL)
+            response.raise_for_status()
+            pricing = response.json()
+            logger.debug("Loaded gatekeeper pricing from models.dev API")
+            return pricing
+    except Exception as exc:
+        logger.debug("Failed to fetch models.dev pricing (%s); defaulting to $0", exc)
```
owtaylor: I think .warning() not .debug() - something went wrong

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,172 @@
...
+        if not isinstance(entry, dict):
+            continue
+        cost = entry.get("cost")
+        if not isinstance(cost, dict):
+            continue
+        input_mtok = cost.get("input")
+        output_mtok = cost.get("output")
+        if isinstance(input_mtok, (int, float)) and isinstance(output_mtok, (int, float)):
+            return float(input_mtok), float(output_mtok)
+    return None
```
owtaylor, line 120: This defensive style means that if things go wrong, it's utterly mysterious - you could easily just assume that the pricing data doesn't exist.

Something like:

```suggestion
    for candidate in _model_lookup_candidates(model):
        entry = models.get(candidate)
        if not entry:
            continue
        try:
            cost = entry["cost"]
            return float(cost["input"]), float(cost["output"])
        except KeyError, TypeError:
            logger.exception("Failed to parse models.dev entry for %s/%s", provider_key, model)
```

Is equally robust, though a lot noisier in the logs. But generally, if models.dev somehow changed their data structures, we'd want to know and fix it. It also a lot easier to get to 100% code coverage on.

src/linux_mcp_server/gatekeeper/openai_client.py
```diff
@@ -0,0 +1,95 @@
...
+    if reasoning is not None:
+        body["reasoning"] = reasoning
+    return body
+
+
+def _extract_responses_text(response: dict[str, Any]) -> str:
+    chunks: list[str] = []
+    for item in response.get("output", []):
+        if not isinstance(item, dict) or item.get("type") != "message":
+            continue
```
owtaylor: This is another case where writing in a super-defensive style is producing a lot of new code paths to cover. If we assume that the response is actually conforming to the documented API, then:
```python
for item in response["output"]:
    if item["type"] != "message":
        continue
    for part in item["content"]:
         if part["type"] == "output_text":
              chunks.append(part["text"])
```
is all you need. And if it's *not* conforming to the documented API, then just silently getting an empty response and throwing an exception later doesn't make sense. If you want to make things a little nicer, you can wrap that in something like:
```python
except KeyError, TypeError as e:
   raise ValueError("Failed to parse response from Responses API: %s", json.dumps(response)) from e
```

owtaylor (changes requested):
OK, down to style here. And I'd also like to see 100% coverage on the new files - this patch does something like take our code coverage from: 98% => 97% - roughly a 50% increase in uncovered lines. It is definitely annoying to add tests for obviously correct error cases, but then we can actually notice lines in the code that should be covered but aren't.

src/linux_mcp_server/gatekeeper/openai_client.py
subpop: Yea, these unpacking routines have been bothering me a lot. I've been tempted to explore the idea of using data structures and deserializing the JSON into data structures instead of working with simple "dict[str, any]" types.

owtaylor: Maybe try rewriting this using local-to-the-file "just enough" Pydantic data structures - I'm not sure if it's going to be a win, but in the days of LLM's, it's fast enough to try!

subpop: Ok. I implemented this idea as a fairly significant refactor, but the general idea is that requests and responses are modeled as Pydantic types and are (de)serialized around the HTTP requests. I also consolidated a lot of the functionality to make the Vertex AI client as much of an "API style router" as possible. This way it reuses the complete functions from the API surfaces that it implements instead of carrying its own duplicate implementations. This is done in commit 5612bb9a90fcca4a732d50f3dffd201d9d08c96c.

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,149 @@
...
+    max_tokens: int
+    messages: list[AnthropicMessage]
+    temperature: float
+    model: str | None = None
+    thinking: AnthropicThinking | None = None
+    output_config: AnthropicOutputConfig | None = None
+
+
+class AnthropicContentText(BaseModel):
+    model_config = ConfigDict(extra="ignore")
```
owtaylor, line 60: This is default ... I think it's a bit confusing to have it explicitly just on the models where we care. If you want to document that we're not being comprehensive, I'd use a comment.

```py
class AnthropicContentText(BaseModel):
    type: Literal["text"]
    text: str
    # ... other fields ...
```

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,149 @@
...
+    content: list[AnthropicContentText] = []
+    usage: AnthropicUsage = Field(default_factory=AnthropicUsage)
+
+    @field_validator("content", mode="before")
+    @classmethod
+    def _text_items_only(cls, value: Any) -> list[Any]:
+        if not isinstance(value, list):
+            return []
+        return [item for item in value if isinstance(item, dict) and item.get("type") == "text"]
+
```
owtaylor, line 85: I don't love this:
 a) It doesn't produce a validation error if input isn't what we expect
 b) it introduces extra code paths that are hard to cover in testing
 c) It's "spooky action at a distance" - `response.content` magically misses things that you'd expect to be there

If we're willing to up the verbosity/complexity:

```py
class AnthropicContentText(BaseModel):
    type: Literal["text"]
    text: str

class AnthropicContentOther(BaseModel):
    type: str

def _tag(v: Any) -> str:
    t = v.get("type") if isinstance(v, dict) else getattr(v, "type", None)
    return t if t in {"text"} else "other"

AnthropicContent = Annotated[
    Union[
        Annotated[AnthropicContentText, Tag("text")],
        Annotated[AnthropicContentOther, Tag("other")],
    ],
    Discriminator(_tag),
]

class AnthropicResponse(BaseModel):
    content: list[AnthropicContent] = Field(default_factory=list)
    usage: AnthropicUsage = Field(default_factory=AnthropicUsage)

    @property
    def text(self) -> str:
        return "".join(c.text for c in self.content if isinstance(c, AnthropicContentText))
```
is considerably cleaner.

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,149 @@
...
+class AnthropicOutputFormat(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
+
+    type: Literal["json_schema"]
+    schema_: dict[str, Any] = Field(alias="schema")
+
+
+class AnthropicOutputConfig(BaseModel):
+    model_config = ConfigDict(populate_by_name=True)
```
owtaylor, line 42: Do you really want `populate_by_name` here?

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,149 @@
...
+        temperature=CONFIG.gatekeeper.temperature,
+        model=CONFIG.gatekeeper.model if include_model else None,
+        thinking=(
+            AnthropicThinking(type="disabled")
+            if reasoning_effort == ReasoningEffort.NONE
+            else AnthropicThinking(type="adaptive")
+            if reasoning_effort is not None
+            else None
+        ),
+        output_config=output_config if output_config.format is not None or output_config.effort is not None else None,
```
owtaylor, line 126: presumably `"output_config": {}` is valid - I'm not sure it's worth effort to avoid that

src/linux_mcp_server/gatekeeper/anthropic_client.py
```diff
@@ -0,0 +1,149 @@
...
+        max_tokens=max_tokens,
+        messages=[AnthropicMessage(role="user", content=prompt)],
+        temperature=CONFIG.gatekeeper.temperature,
+        model=CONFIG.gatekeeper.model if include_model else None,
+        thinking=(
+            AnthropicThinking(type="disabled")
+            if reasoning_effort == ReasoningEffort.NONE
+            else AnthropicThinking(type="adaptive")
+            if reasoning_effort is not None
+            else None
```
owtaylor, line 124:
```suggestion
            None if reasoning_effort is None
            else AnthropicThinking(type="disabled") if reasoning_effort == ReasoningEffort.NONE
            else AnthropicThinking(type="adaptive")
```

src/linux_mcp_server/gatekeeper/gemini_client.py
```diff
@@ -0,0 +1,140 @@
...
+
+class GeminiRequest(BaseModel):
+    contents: list[GeminiContent]
+    generationConfig: GeminiGenerationConfig
+
+
+class GeminiResponsePart(BaseModel):
+    model_config = ConfigDict(extra="ignore")
+
+    text: str = ""
```
owtaylor, line 61: I'd do `str | None = None` and handle `None` when building the total text - small difference but I got confused when I read this "aren't there non-text parts that don't even have a text field?

src/linux_mcp_server/gatekeeper/openai_client.py
```diff
@@ -0,0 +1,142 @@
...
+
+    type: Literal["output_text"]
+    text: str
+
+
+class OpenAIResponseOutputMessage(BaseModel):
+    model_config = ConfigDict(extra="ignore")
+
+    type: Literal["message"]
+    content: list[OpenAIResponseOutputText]
```
owtaylor, line 34: It's a bit confusing reading this why you are aren't filtering other types here - suggest commenting explicitly.
```
    # The only child type is `"type": "output_text"` according to the responses API definition
```

src/linux_mcp_server/gatekeeper/openrouter_client.py
```diff
@@ -0,0 +1,149 @@
...
+
+
+async def complete_openrouter(
+    prompt: str, *, max_tokens: int, timeout: int = DEFAULT_TIMEOUT_SECONDS
+) -> GatekeeperCompletion:
+    assert CONFIG.gatekeeper is not None
+    configured = CONFIG.gatekeeper.openrouter.base_url if CONFIG.gatekeeper.openrouter else None
+    base_url = (configured or OPENROUTER_DEFAULT_BASE_URL).rstrip("/")
+    quantization = CONFIG.gatekeeper.openrouter.quantization if CONFIG.gatekeeper.openrouter else None
+    template_kwargs = CONFIG.gatekeeper.openrouter.template_kwargs if CONFIG.gatekeeper.openrouter else {}
```
owtaylor, line 105:
```
openrouter_config = CONFIG.gatekeeper.openrouter or OpenRouterGatekeeperConfig()
```
Then the intermediate variables probably aren't even necessary.

src/linux_mcp_server/gatekeeper/openrouter_client.py
```diff
@@ -0,0 +1,149 @@
...
+            ),
+        )
+        if CONFIG.gatekeeper.structured_output
+        else None,
+        reasoning=(
+            OpenRouterReasoning(enabled=False)
+            if reasoning_effort == ReasoningEffort.NONE
+            else OpenRouterReasoning(enabled=True, effort=reasoning_effort.value)
+            if reasoning_effort is not None
+            else None
```
owtaylor, line 128: Again, this gets cleaner if you reorder
```suggestion
            None if reasoning_effort is None
            else OpenRouterReasoning(enabled=False) if reasoning_effort == ReasoningEffort.NONE
            else OpenRouterReasoning(enabled=True, effort=reasoning_effort.value)
```

src/linux_mcp_server/gatekeeper/pricing.py
```diff
@@ -0,0 +1,162 @@
...
+            response.raise_for_status()
+            raw = response.json()
+            if not isinstance(raw, dict):
+                return {}
+            pricing: dict[str, ModelsDevProvider] = {}
+            for key, value in raw.items():
+                try:
+                    pricing[key] = ModelsDevProvider.model_validate(value)
+                except ValidationError:
+                    continue
```
owtaylor: Better to never silently suppress errors.

subpop:
@owtaylor I added tests or test cases to cover the remaining edge cases and uncovered lines. This should be 100% covered now.

owtaylor (commented):
Hmm, not entirely sure how I feel about the update. After the change, things are:

* more declarative
* better type checked by pyright
* more verbose
* a bit harder to read because we separate out the API definition from the API usage

But I think it's definitely a legitimate way to do things, especially considering that we heavily use Pydantic already. If you are happy with it, let's go with this. Various niggling around on the details in the comments - one thing that we might want to consider doing differently is handling unknown children - there's no really *nice* way to do it, but I comment an alternate approach below.

owtaylor (approved):
This is getting massive as a review. I think the testing changes are fine. You can go ahead and land it, and one of us can implement the rest of my review comments (or the ones we care about) as a cleanup on top.

PR #501, opened 2026-06-11
Improve documentation for running in a container

owtaylor:
The documentation mixed together Podman and Docker specifics in a quite confusing way. Split things apart into instructions for Podman (using --userns=keep-id) and Docker (changing ownership of files to 1001).

Remove the example of running the server locally with a HTTP transport - we now require an auth_policy.json for the HTTP transport, so it no longer works at all, and it's also not something that we want to support.

### DIFF
--- a/docs/clients.md
+++ b/docs/clients.md
@@ -43,7 +43,7 @@ Edit `~/.claude.json`:
     }
     ```
 
-=== "Container (stdio transport)"
+=== "Container (Podman)"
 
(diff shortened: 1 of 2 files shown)


subpop (approved):

panyamkeerthana (approved):

owtaylor:
Rebased to fix broken CI tests that were blocking merging.

PR #511, opened 2026-06-18
Containerfile: Add a release_type label to control floating tag

owtaylor:
We currently tag every single build - whether from a tagged release or not as `:latest` in our Konflux-backed quay.io builds. To allow us to only label tagged releases with the `:latest` tag add a label to the Containerfile that is `latest` (for a tagged release) and `dev` otherwise.

### DIFF
--- a/.tekton/pipeline-build-multiarch.yaml
+++ b/.tekton/pipeline-build-multiarch.yaml
@@ -244,6 +244,7 @@ spec:
             - $(params.build-args[*])
             - PSEUDO_VERSION=$(tasks.get-version.results.pseudo_version)
             - VERSION=$(tasks.get-version.results.version)
+            - RELEASE_TYPE=$(tasks.get-version.results.release_type)
 
         - name: BUILD_ARGS_FILE
           value: $(params.build-args-file)
--- a/.tekton/task-get-version.yaml
+++ b/.tekton/task-get-version.yaml
@@ -18,6 +18,9 @@ spec:
     - name: version
       description: Version that is also a valid container tag
 
+    - name: release_type
+      description: Floating tag we should use for this release (latest/dev)
+
   volumes:
     - name: workdir
       emptyDir: {}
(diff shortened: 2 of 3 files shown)


.tekton/pipeline-build-multiarch.yaml
Jazzcort, line 643: Shouldn't we just add an entry here to grab the release_type from get version?

```yaml
- name: ADDITIONAL_TAGS
   value:
     - $(params.additional-tags[*])
     - $(params.tag-prefix)$(tasks.clone-repository.results.short-commit)
     - $(tasks.get-version.results.release_type)
```

And we delete the entire `additional-tags` entry from spec.params in `push.yaml`.

Or I might misunderstand something. 😆

Jazzcort (commented):
My thought could be wrong. 😆

samdoran:
This is a lot trickier than it should be. There are two places where the `latest` tag is applied: 1) in the [push pipeline](https://github.com/rhel-lightspeed/linux-mcp-server/blob/d0c2bf320ccbd5af3306e8b7f3b436bee52262d2/.tekton/push.yaml#L41-L43) which we directly control and 2) in the release plan admission which we do not **directly** control (we can pass parameters but cannot define custom tasks):

Excerpt from the release plan admission parameters:
```yaml
...
  data:
    mapping:
      registrySecret: konflux-release-service-access-management-token
      components:
        - name: linux-mcp-server
          public: true
          pushSourceContainer: false
          repositories:
            - url: quay.io/redhat-services-prod/rhel-lightspeed-tenant/linux-mcp-server
              tags:
                - latest
                - "{{ git_short_sha }}"
                - "{{ labels.version }}"
...
```

For the push task, we can directly control the labels that get applied by using outputs from custom tasks. But the only way to send data to the release pipeline is through container image labels. We need to remove `additional-tags` from the push pipeline as @Jazzcort suggested and remove the hard coded value in the release plan admission to use `{{ labels.release_type }}` (which is in another repo).

That will solve one problem, but what we need is actually a bit more complicated. The `latest` tag is expected to mean "latest stable release". We should only apply the `latest` label if the version matching the ref is the latest stable release.

Say we have version 1.y and 2.y released. If we push a new 1.y.z version, it will get the `latest` label, which is incorrect. That's a future edge case but worth thinking about. I [fixed this problem in lightspeed-core](https://github.com/lightspeed-core/lightspeed-stack/pull/1921). Doing the equivalent in Konflux is possible but would be slightly more complicated.

.tekton/pipeline-build-multiarch.yaml
samdoran: You're absolutely correct. Good catch!

Jazzcort: Now I can say I know a little bit about Konflux 😆

PR #517, opened 2026-06-23
[RSPEED-2265] Convert network tools to structured output

subpop:
Straightforward port of the network.py tools to return structured output instead of string-formatted text.

### DIFF
--- a/src/linux_mcp_server/tools/network.py
+++ b/src/linux_mcp_server/tools/network.py
@@ -1,11 +1,12 @@
 """Network diagnostic tools."""
 
+from fastmcp.exceptions import ToolError
 from mcp.types import ToolAnnotations
 
 from linux_mcp_server.audit import log_tool_call
 from linux_mcp_server.commands import get_command
 from linux_mcp_server.formatters import format_listening_ports
-from linux_mcp_server.formatters import format_network_connections
+from linux_mcp_server.models import NetworkConnection
 from linux_mcp_server.models import NetworkInterface
 from linux_mcp_server.parsers import merge_network_interfaces
 from linux_mcp_server.parsers import parse_ip_brief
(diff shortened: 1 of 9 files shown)


tests/functional/network_diagnostics/test_get_listening_ports.py
```diff
@@ -7,22 +7,18 @@ async def test_listening_ports(mcp_session):
...
     Verify that the server lists correctly all the available listening ports.
     """
-
     response = await mcp_session.call_tool("get_listening_ports")
     assert response is not None
 
-    mcp_output = response.content[0].text
-    shell_output = shell("ss -tulnp", silent=True).stdout.strip()
+    content = response.structured_content["result"]
+    assert content is not None
```
owtaylor, line 14: The functional tests are using `mcp.client.session.ClientSession` *not* fastmcp's client. A significant difference between the two is that for fastmcp, CallToolResult has `.structured_content`, but the MCP equivalent has `.structuredContent` - this is causing the functional test CI failures. (The rest of the adapted functional tests just parse the json from requests.content - we probably be consistent across tests and have some infrastructure that we use across tests that asserts that the two are identical - but I'm fine here with just the s/structured_content/structuredContent/ change) 

Note: On a Linux system, you can try running the functional tests directly with: `uv run pytest tests/functional/network_diagnostics/test_get_listening_ports.py` - you'd probably need to do that in a VM on a mac.

src/linux_mcp_server/tools/network.py
```diff
@@ -50,7 +52,7 @@ async def get_network_interfaces(
     if is_successful_output(returncode, stdout):
         stats = parse_proc_net_dev(stdout)
 
-    return format_network_interfaces(interfaces, stats)
+    return merge_network_interfaces(interfaces, stats)
```
owtaylor, line 55: This is inconsistent with the other two tools which throw ToolError if the underlying command fails. I think this should throw ToolError if *either* of the two commands fail. This is particularly true because we don't have any way of distinguishing "could not retrieve interface stats" from "interface has no traffic".

owtaylor (changes requested):

PR #521, opened 2026-07-01
Sync exec timeout with mcp app

Jazzcort:
Add a get_execution_timeout server tool so the mcp-app can fetch the configured command timeout and apply it (plus a 5s grace period) to the execute_script call. This ensures long-running scripts aren't cut short by the SDK's default request timeout when the server allows more time.

### DIFF
--- a/mcp-app/src/run-script-app.tsx
+++ b/mcp-app/src/run-script-app.tsx
@@ -6,11 +6,12 @@ import {
   useApp,
 } from "@modelcontextprotocol/ext-apps/react";
 import type { CallToolResult } from "@modelcontextprotocol/sdk/types.js";
+import { DEFAULT_REQUEST_TIMEOUT_MSEC } from "@modelcontextprotocol/sdk/shared/protocol";
 import { StrictMode, useEffect, useMemo, useState } from "react";
 import { createRoot } from "react-dom/client";
 import {
   ExecuteScriptResultSchema,
-  GetExecutionStateResultSchema,
+  GetExecutionDetailsResultSchema,
   McpAppToolParamsSchema,
   McpAppToolResultSchema,
   type ExecutionState,
@@ -119,6 +120,9 @@ function RunScriptAppInner({
   const [executionState, setExecutionState] =
     useState<ExecutionState>("initial");
   const [executionResult, setExecutionResult] = useState<string>("");
+  const [executionTimeout, setExecutionTimeout] = useState<number>(
+    DEFAULT_REQUEST_TIMEOUT_MSEC,
+  );
 
   const validatedToolRequestParams = useMemo(() => {
     if (!toolRequestParams) {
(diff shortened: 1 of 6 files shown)


Jazzcort:
Not sure if we want to surface the entire error or just the `error.message` here. The previous `JSON.stringfy(e)` will result in an empty object `{}`. 

Screenshot 2026-07-01 at 4 26 00 PM
The entire error is kind of too noisy.

src/linux_mcp_server/tools/run_script.py
```diff
@@ -283,6 +283,17 @@ async def reject_script(
...
 
+@mcp.tool(
+    tags={"run_script", "mcp_apps_only", "hidden_from_model"},
+    description="Get the execution timeout for mcp-app",
+    app=AppConfig(visibility=["app"]),
+)
+@log_tool_call
+@disallow_local_execution_in_containers
+async def get_execution_timeout():
+    return {"timeout": CONFIG.command_timeout * 1000}
```
owtaylor: Making this part of `get_execution_state()` result would keep things simpler and avoid the racy state where we've gotten the execution state back but the execution timeout fetch failed or timed out.

mcp-app/src/run-script-app.tsx
```diff
@@ -212,11 +243,11 @@ function RunScriptAppInner({
       outputToModel = formatOutputForToolError(
         validatedToolRequestParams,
         validatedToolResult,
-        JSON.stringify(e),
+        errorToJSON(e),
```
owtaylor: Maybe this should just be e.message (the parameter is called message...). I don't think providing the stack trace to the model is useful. If we want that to be accessible to developers, maybe better to log the exception?

mcp-app/src/utils.ts
```diff
@@ -69,3 +69,15 @@ export function formatExecutionState(executionState: ExecutionState): string {
       return "Initial";
   }
 }
+
+export function errorToJSON(err: any) {
+  // Get all property names, including the hidden (non-enumerable) ones
```
owtaylor: If we keep this, I would explain here the *why* as well as the what ... that message and stack are non-enumerable.

owtaylor (changes requested):
Generally looks functional. Wondering if we can things more in sync and compact if we make this an additional return from get_execution_state().

src/linux_mcp_server/tools/run_script.py
Jazzcort: I also thought about this! Maybe change the name to get_execution_details? 🤔

mcp-app/src/run-script-app.tsx
Jazzcort: Yes! totally agree! I will change it to just e.message. I think log the exception here is helpful, I'll also add it here.

mcp-app/src/utils.ts
Jazzcort: Yes! But, since we just want the e.message, I will just remove this function. 😁

owtaylor (approved):
I like this - it's a lot simpler.

Jazzcort:
Thanks for the review! 😁

PR #522, opened 2026-07-06
functional-tests: replace /retest comment with ok_to_test label

owtaylor:
The issue_comment trigger produces check runs associated with main's HEAD, not the PR's head commit, so a successful retest never replaced the original failed check in the PR status summary. Switching to a pull_request_target labeled trigger keeps the same (workflow, event, SHA) identity, so the new result properly supersedes the old one.

The ok_to_test label is removed immediately — before the authorization check — so it can always be re-added regardless of whether the actor was authorized.

This is an implementation of the idea from:  https://github.com/rhel-lightspeed/linux-mcp-server/pull/389#discussion_r3070178964

I used a generic ok_to_test label here, because if we had multiple tests that we wanted to gate this way (needs security look-over from a maintainer) we'd only want to do that *once* per commit.

Assisted-by: Claude Opus 4.6 <noreply@anthropic.com>
Closes: #389

### DIFF
--- a/.github/workflows/functional-tests.yml
+++ b/.github/workflows/functional-tests.yml
@@ -5,28 +5,37 @@ on:
     branches:
       - main
   pull_request_target:
+    types: [opened, synchronize, labeled]
     branches:
       - main
-  issue_comment:
-    types: [created]
   workflow_dispatch:
 
 permissions:
   contents: read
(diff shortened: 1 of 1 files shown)


subpop (approved):
Seems like a reasonable approach to me.

PR #523, opened 2026-07-06
auth_policy: add tool exclusion support to policy rules

owtaylor:
Tools prefixed with `-` in a rule's tools list now exclude that tool from matching. Exclusions take precedence over inclusions regardless of order. This makes the code match what has already been documented.

Assisted-By: Claude Opus 4.6 <noreply@anthropic.com>

### DIFF
--- a/src/linux_mcp_server/auth_policy.py
+++ b/src/linux_mcp_server/auth_policy.py
@@ -1,5 +1,6 @@
 import fnmatch
 import logging
+import re
 
 from enum import Enum
 from functools import cache
@@ -63,6 +64,14 @@ class PolicyRule(BaseModel):
                 raise ValueError(f"Rule with host: '{self.host}' cannot be use action 'local'")
         return self
 
+    # Check that tool entries are syntactically valid
+    @model_validator(mode="after")
+    def validate_tools(self):
+        for pattern in self.tools:
+            if not re.match(r"^(?:-?@?[a-z][a-z0-9_]*|\*)$", pattern):
+                raise ValueError("Entry in tools list must be '*', '[-]<tool_name>', or '[-]@<group_name>'")
+        return self
+
     # Validate that SSH_KEY action has ssh_key configuration
     @model_validator(mode="after")
     def validate_ssh_key_config(self):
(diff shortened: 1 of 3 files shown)


src/linux_mcp_server/auth_policy.py
```diff
@@ -77,9 +77,14 @@ def matches_host(self, target_host: str | None) -> bool:
...
-    # Supports exact tool name: "run_script_readonly", prefixes: "@fixed" and wildcard: "*"
+    # Supports exact tool name: "run_script_readonly", prefixes: "@fixed", wildcard: "*",
+    # and exclusions: "-tool_name" (exclusions take precedence over inclusions)
     def matches_tool(self, tool_name: str, tool_tags: set[str]) -> bool:
-        for allowed_tool in self.tools:
+        for denied_tool in (t[1:] for t in self.tools if t.startswith("-")):
+            if denied_tool == tool_name:
+                return False
+
+        for allowed_tool in (t for t in self.tools if not t.startswith("-")):
```
panyamkeerthana: The the tool exclusion logic looks solid  but I want to understand if this matches a specific requirement or if its general infrastructure

panyamkeerthana (dismissed):
Nice implementation! And the test coverage looks good too! LGTM ✅

src/linux_mcp_server/auth_policy.py
owtaylor: It's something I had in my early notes, and thought it had been implemented, so documented it. And decided to leave it in the documentation when I realized that it actually hadn't been implemented, and do it later.

The reason I decided to make excluding groups disallowed is it seemed to me that `['*', "-@run_script"]` was just a confusing way to write `["@fixed"]`

But actually, there's a worse problem now `['*', "-@run_script"]` is just a confusing way to write `['*']`. Which is a trap. We should either make it work or add validation to catch it. Probably just making it work is easiest.

panyamkeerthana: Should we add validation to reject  -@ patterns with an error or actually handle excluding tool groups?

owtaylor: I went ahead and implemented excluding tool groups to remove a corner case. Then I went ahead and implemented basic validation to avoid people using `-get_*` or something like that...

[I *didn't* validate that tool names and group names actually existed - to keep things simple and because that makes it more robust if your config file slightly mismatches your MCP server version ... but perhaps that would be a good idea as well. Not sure.]

panyamkeerthana (approved):
Perfect! The toolset exclusion support fixes the security trap I was worried about. Love the validation that rejects patterns like `get_*` - makes it fail-fast instead of silently doing nothing.

The refactored `matches()` helper is clean and the test coverage is comprehensive. Ready to merge. 🚀

PR #524, opened 2026-07-08
ci: Upgrade to the latest mcp-publisher version

owtaylor:
The 1.4.0 version we were using isn't compatible with the current registry, because of a change in OIDC usage.

See https://github.com/modelcontextprotocol/registry/pull/1229


owtaylor:
We might want to consider redoing things to use the official github action for this ... that would bring us into the world of renovate (haven't looked at what would be involved)

panyamkeerthana (approved):

PR #532, opened 2026-07-14
docs: add Keycloak configuration guide

panyamkeerthana:
- Adds `docs/keycloak.md` with instructions for configuring Keycloak for JWT and Introspection-based authentication with linux-mcp-server.

### DIFF
--- a/docs/keycloak.md
+++ b/docs/keycloak.md
@@ -1,0 +1,123 @@
+# Keycloak Configuration
+
+[Keycloak](https://www.keycloak.org) is a open-source identity and access management solution that can be used as the OAuth2 authorization server for linux-mcp-server. This section describes how to configure a standalone instance of KeyCloak for testing purposes. It will also be useful if your organization is already using KeyCloak and you want to add linux-mcp-server to it.
+
+
+## Prerequisites
+
+- Keycloak server running
+- Admin access to Keycloak Admin Console
+
+### Running Keycloak Locally with Podman
+
+```bash
+# Pull the latest Keycloak image
+podman pull quay.io/keycloak/keycloak:latest
+
+# Run Keycloak in development mode
+podman run -d \
+  --name keycloak \
+  -p 8080:8080 \
+  -e KEYCLOAK_ADMIN=admin \
+  -e KEYCLOAK_ADMIN_PASSWORD=admin \
+  quay.io/keycloak/keycloak:latest start-dev
+
+# Check logs
+podman logs -f keycloak
+
+# Access admin console at: http://localhost:8080/admin
+# Login with: admin / admin
+```
+
+---
+
+The approach we'll take here is using **Dynamic Client Registration (DCR)**. With DCR, MCP clients automatically register themselves with the authorization. Otherwise, clients would have to be added through the admin console, which is not typically supported for chat clients like Goose or Claude Desktop.
+
+Either the jwt or introspection authorization backend for linux-mcp-server can be used with KeyCloak. The difference is slight - the jwt backend is easier to set up and might offer a bit better performance, while the introspection backend offers immediate token revocation without waiting for token expiration if, for example, a user account is removed.
+
+### Step 1: Create Realm
+
+A realm is an isolated workspace inside Keycloak. It acts as a standalone container with its own users, apps, and login rules that don't mix with anything else on the server.
+
+1. Navigate to **Keycloak Admin Console**: `http://localhost:8080/admin/master/console/`
+2. Log in with admin credentials
+3. Click the realm dropdown (top left, shows "master") → **Create Realm**
+4. Configure:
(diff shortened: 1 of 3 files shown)


owtaylor:
This needs to be added to mkdocs.yaml so that it gets built as part of the docs. There's not a great place for it in the docs tree, I think I'd do it as the last item under "Usage" "Configuring Keycloak: keycloak.md"

You can run `make docs-serve` to test the build locally.

docs/keycloak.md
```diff
@@ -0,0 +1,119 @@
+# Keycloak Configuration
+
+This guide covers how to configure Keycloak for JWT  and Introspection based authentication.
```
owtaylor: Needs expansion and context: 
```suggestion
[Keycloak](https://www.keycloak.org) is a open-source identity and access management solution that can be used as the OAuth2 authorization server for linux-mcp-server. This section describes how to configure a standalone instance of KeyCloak for testing purposes. It will also be useful if your organization is already using KeyCloak and you want to add linux-mcp-server to it.
```

docs/keycloak.md
```diff
@@ -0,0 +1,119 @@
...
+
+
+## Prerequisites
+
+- Keycloak server runnings
```
owtaylor:
```suggestion
- Keycloak server running
```

docs/keycloak.md
```diff
@@ -0,0 +1,119 @@
...
+# Login with: admin / admin
+```
+
+---
+
+This approach uses **Dynamic Client Registration (DCR)** where MCP clients automatically register themselves with Keycloak. The MCP server validates JWT tokens or uses introspection.
+
+### Step 1: Create Realm
+
+1. Navigate to **Keycloak Admin Console**: `http://localhost:8080/admin/master/console/`
```
owtaylor, line 42: I think you should have a sentence here to say what a realm is.

docs/keycloak.md
```diff
@@ -0,0 +1,119 @@
...
+2. Log in with admin credentials
+3. Click the realm dropdown (top left, shows "master") → **Create Realm**
+4. Configure:
+   - **Realm name**: Choose your realm name
```
owtaylor: Provide an example

docs/keycloak.md
```diff
@@ -0,0 +1,119 @@
...
+export LINUX_MCP_AUTH__JWT__AUDIENCE="account"  # Must match the audience mapper value
+
+```
+
+**Authentication Flow**: Clients discover Keycloak via `/.well-known/oauth-protected-resource/mcp`, auto-register using DCR, obtain JWT tokens through Keycloak's OAuth flow, and send tokens to the MCP server which validates them locally using JWKS.
+
+
+### Token Introspection
+
+The MCP server validates tokens by calling Keycloak's introspection endpoint. Requires a confidential client.
```
owtaylor: Should briefly say what a confidential client is

owtaylor (changes requested):
Looks great! - beyond adding to mkdocs.yaml, I have a few tweaks and suggested expansions.

docs/keycloak.md
```diff
@@ -0,0 +1,121 @@
...
+# Check logs
+podman logs -f keycloak
+
+# Access admin console at: http://localhost:8080/admin
+# Login with: admin / admin
+```
+
+---
+
+This approach uses **Dynamic Client Registration (DCR)** where MCP clients automatically register themselves with Keycloak. The MCP server validates JWT tokens or uses introspection.
```
owtaylor: This paragraph mixes two things, both of which could do with some expansion.

```suggestion
The approach we'll take here is using **Dynamic Client Registration (DCR)**. With DCR, MCP clients automatically register themselves with the authorization. Otherwise, clients would have to be added through the admin console, which is not typically supported for chat clients like Goose or Claude Desktop.

Either the jwt or introspection authorization backend for linux-mcp-server can be used with KeyCloak. The difference is slight - the jwt backend is easier to set up and might offer a bit better performance, while the introspection backend offers immediate token revocation without waiting for token expiration if, for example, a user account is removed.
```

owtaylor: Meant to add this comment on the last review, but it was still sitting pending.

docs/keycloak.md
```diff
@@ -0,0 +1,121 @@
...
+
+### Step 1: Create Realm
+
+A realm is Keycloak's top-level unit of tenancy, it manages its own set of users, clients, roles, and authentication settings, isolated from other realms on the same server.
```
owtaylor: "unit of tenancy" is obscure - here's a suggestion from Gemini - "A realm is an isolated workspace inside Keycloak. It acts as a standalone container with its own users, apps, and login rules that don't mix with anything else on the server."

owtaylor (changes requested):
Looking good, couple of more suggestions

owtaylor (approved):
Looks great!

PR #534, opened 2026-07-20
fix: split lines parameter into first_lines and last_lines

### LINKED ISSUE #472: get_journal_logs: split lines into first_lines / last_lines
Based on discussion in #313 (cc: @umglurf)

If the model calls get_journal_logs() with `{ "since: "2026-01-01", "lines": 10 }` they will get the last 10 lines, not the first 10 lines after the time.  This is consistent with `journalctl --since <TIME> -n 10` but probably not what was intended. For systemd, the more useful behavior is expressible with `journalctl --since <TIME> -n +10`

Rather than introducing funny string syntax, I suggest splitting `lines` into `first_lines` and `last_lines`, so that the parameters are self-explanatory.

Note that before https://github.com/systemd/systemd/pull/26669 (system 254 I think, RHEL 8 is 252, RHEL 9 257) behaved pretty badly with `-n` and `--since/--until/--unit`.

```
    journalctl -n 10 - last 10 lines
    journalctl -u <UNIT> -n 10 - last 10 lines for unit
    journalctl --since <TIME> -n 10 - first 10 lines after time
    journalctl --until <TIME> -n 10 - nothing
```

But I think we can let `{ "since: "2026-01-01", "last_lines": 10 }` be broken for RHEL 8 rather than adding some complicated fallback logic, since that combination is rarely useful.

panyamkeerthana:
Fixes #472

Split the 'lines' parameter in get_journal_logs into 'first_lines' and 'last_lines' to make journalctl -n vs -n +N behavior explicit.

Changes:
- Updated get_journal_logs tool to accept first_lines and last_lines instead of lines parameter
- first_lines uses 'journalctl -n +N' to get first N lines after filters
- last_lines uses 'journalctl -n N' to get last N lines (default: 100)
- Parameters are mutually exclusive with validation
- Updated command registry to support both optional_flags
- Updated all unit and functional tests

### DIFF
--- a/src/linux_mcp_server/tools/logs.py
+++ b/src/linux_mcp_server/tools/logs.py
@@ -43,7 +43,8 @@ class Transport(StrEnum):
 
 
 async def _get_journal_logs(
-    lines: int,
+    first_lines: int | None = None,
+    last_lines: int | None = None,
     host: Host | None = None,
     unit: str | None = None,
     priority: str | None = None,
@@ -53,7 +54,8 @@ async def _get_journal_logs(
     """Execute journalctl command with optional filters.
 
     Args:
-        lines: Number of log lines to retrieve.
+        first_lines: Number of first log lines to retrieve (using -n +N).
+        last_lines: Number of last log lines to retrieve (using -n N).
         host: Optional remote host address.
         unit: Filter by systemd unit name.
         priority: Filter by priority level.
@@ -70,7 +72,8 @@ async def _get_journal_logs(
     cmd = get_command("journal_logs")
     return await cmd.run(
         host=host,
-        lines=lines,
+        first_lines=first_lines,
+        last_lines=last_lines,
         unit=unit,
         priority=priority,
         since=since,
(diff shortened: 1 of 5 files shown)


panyamkeerthana (commented):
LGTM! The only thing I can think of is adding functional test cases for first_lines to cover the -n +N path and fixing the failed CI formatting

owtaylor (changes requested):
Generally looks great. I'd like to see read_log_file changed to match - keeping things consistent might *slightly* reduce the chance of model confusion, and it just is tidier to have things the same.

Claude Code /code-review points out:

> Test helper string replacement mismatch (tests/functional/logs_audit/test_get_journal_logs.py:22) — The _assert_journal_logs_match helper does .replace("-n 5", "-n 100") but the first_lines test uses "-n +5" (with a +), so the widening replacement is a no-op. Low practical risk since oldest log entries are stable, but inconsistent with the other tests.

src/linux_mcp_server/tools/logs.py
```diff
@@ -157,14 +192,53 @@ async def read_log_file(
...
+
+    Examples:
+        Get last 50 lines from messages log:
+            read_log_file(log_path="/var/log/messages", last_lines=50)
+
+        Get first 20 lines from a log file:
+            read_log_file(log_path="/var/log/audit/audit.log", first_lines=20)
+
+    Use first_lines to get the first N lines from the beginning of the file.
+    Use last_lines to get the last N lines from the end (default behavior).
```
owtaylor, line 219: If this was the docs that were given to the LLM, I would think it was oververbose, because we're cluttering the context with three separate copies of the first_line/last_lines parameter docs. But it's not - it's simply internal documentation. I think Claude Code missed that. The description is simply "Read a specific log file."

As internal documentation, I think the examples are also unnecessary and can be omitted.

tests/functional/logs_audit/test_read_log_file.py
```diff
@@ -21,7 +21,7 @@ async def test_read_existing_log_file(mcp_session):
         f.writelines(file_content)
     try:
         response = await mcp_session.call_tool(
-            "read_log_file", arguments={"log_path": "/tmp/existing_file.txt", "lines": 5}
+            "read_log_file", arguments={"log_path": "/tmp/existing_file.txt", "last_lines": 5}
         )
         assert response is not None
```
owtaylor, line 26: I think it would make sense to add a functional test for first_lines ... make sure that 'head' behaves as we expect.

owtaylor (changes requested):
Just a couple of more things.

owtaylor (approved):
Looks great!

PR #555, opened 2026-08-18
Repoint Konflux configuration and update Konflux pipeline

owtaylor:
This is equivalent to #545, which was automatically filed when onboarding to the public cluster, but compared to that:

 * It retains the better structure both pull-request and push reference a common pipeline
 * It retains our local customizations

Compared to the current state:

 * The source repository is changed to our current github repository
 * The pipeline definition is updated. Most of the changes are either task reference updates or minor changes that don't matter for us. The most significant change is probably that some checks like clair-scan are now run per architecture.

This PR was created by running the pipeline definition from #545 and our build definitions through `yq` to eliminate formatting differences, and then working until the diff contained only our local changes that we need to retain. Might be better to maintain our changes from the common pipeline as a kustomize patch.

### DIFF
--- a/.tekton/pipeline-build-multiarch.yaml
+++ b/.tekton/pipeline-build-multiarch.yaml
@@ -4,6 +4,19 @@ metadata:
   name: pipeline-build-multiarch
 
 spec:
+  description: |
+    This pipeline is ideal for building multi-arch container images from a Containerfile while maintaining trust after pipeline customization.
+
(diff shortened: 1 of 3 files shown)


samdoran:
I'm curious why the change to run certain checks per architecture?

samdoran (approved):

owtaylor:
Thanks for the review!

> I'm curious why the change to run certain checks per architecture?

I was just mirroring what the default pipeline does to reduce the delta. Digging into it, it seems to be a performance thing, see, e.g.:

https://github.com/konflux-ci/build-definitions/commit/b7d9032914a74621ca48b8b24136d35b3ba159da

PR #557, opened 2026-08-19
renovate.json: drop grouping

owtaylor:
From a couple of weeks experience, grouping Renovate changes is a complete pain - we get updates in every group every week, meaning that instead of *1* weekly renovate commit to handle we have 4 or so, plus lockfile updates. More CI tests to trigger, more CI flakes to deal with, more opportunity for error. Better to do everything at once, even if that means dealing with a commit that mixes together different stuff.

### DIFF
--- a/renovate.json
+++ b/renovate.json
@@ -3,12 +3,6 @@
   "extends": ["github>rhel-lightspeed/renovate-config"],
   "packageRules": [
     {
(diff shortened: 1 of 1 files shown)


panyamkeerthana (approved):

panyamkeerthana (approved):
Makes sense! instead of juggling multiple renovate PRs each week we'll just have one to review and merge😊

PR #558, opened 2026-08-19
CI: remove gitlab PR mirroring and CI

owtaylor:
Now that we have Konflux running directly on our github repository, that removes the main reason we were mirroring PR's to gitlab.cee.redhat.com MR's. Drop all the infrastructure for mirroring.

This leaves a hole - we are no longer running gatekeeper evals on PR's, but that wasn't working properly for various reasons anyways. It can be added back directly as a GitHub action - though we'll need to make some changes to which inference providers and models we test on.


Jazzcort (approved):
LGTM! I can also follow up on updating the Github action! 😁

PR #564, opened 2026-08-24
Build our container hermetically from source

owtaylor:
Instead of pulling binary packages from PyPI, declare all our dependencies in lock files, and build everything from source.

Build requirements are declared in files:

  `requirements-build[-N].in` => `requirements-build[-N].txt`

which allows renovate's lock-file-maintenance to keep them updated using the pip-compile backend.

Maintaining the requirements files is done by [slushy-build](https://github.com/owtaylor/slushy-build) - a custom solution that does exactly what we need. If this works out, similar functionality could be added to the upstream pybuild-deps.

When a PR touches uv.lock, a workflow verifies that the requirements files are updated, and errors out, or if possible, pushes a fix-up commit.

New Makefile targets:

  * `requirements-lock`: update Python requirements files
  * `rpm-lock`: update RPM lockfiles
  * `hermeto-prefetch` `hermeto-clean`: prefetch dependencies
  * `container-build-fast`: builds the container from binary deps
  * `container-build-production`: builds the container from source

Note:
 * On my system, container-build-fast takes about 2 minutes, container-build-production
  about 15 minutes.
 * Based partially on the setup from rhel-lightspeed/okp-mcp
 * Assisted by Claude Code


Containerfile
```diff
@@ -1,41 +1,64 @@
...
+         ${VENV}/bin/pip install --no-binary :all: -r $SOURCE/.konflux/requirements.txt ; \
+       else \
+         ${VENV}/bin/pip install -r $SOURCE/.konflux/requirements.txt ; \
+       fi \
+    && ${VENV}/bin/pip uninstall -y diskcache
 
+# Provide the version to avoid the need to pass in the .git directory.
+# https://setuptools-scm.readthedocs.io/en/latest/usage/#with-dockerpodman
+# FIXME: This should be SETUPTOOLS_SCM_PRETEND_VERSION_FOR_${DIST_NAME} but I
+#        can't figure out what exactly the value for DIST_NAME should be.
```
r0x0d: should be `LINUX_MCP_SERVER`, no?

r0x0d: For instance: https://src.fedoraproject.org/rpms/python-distributed/blob/rawhide/f/python-distributed.spec#_118

.tekton/pull_request.yaml
```diff
@@ -34,6 +34,15 @@ spec:
...
       value: /Containerfile
 
+    - name: hermetic
+      value: "true"
+
+    # Hermeto prefetch: pip sdists (no binary field = no binary SBOM annotations)
+    # plus the build-toolchain RPMs (rpms.lock.yaml). hermetic builds compile
+    # every wheel from source, so the network is fully isolated.
+    - name: prefetch-input
+      value: '[{"type": "pip", "path": ".", "requirements_files": [".konflux/requirements.txt"], "requirements_build_files": [".konflux/requirements-build.txt", ".konflux/requirements-build-1.txt"]}, {"type": "rpm", "path": "."}]'
```
r0x0d, line 44: I think you could move this to https://github.com/rhel-lightspeed/linux-mcp-server/blob/main/.tekton/pipeline-build-multiarch.yaml to avoid drifting those settings in the future (for pr and push builds)

Makefile
```diff
@@ -63,3 +67,68 @@ docs:
...
 	uv run --locked --group docs mkdocs serve --dev-addr localhost:8010 --livereload
+
+# Run Hermeto locally to validate sdist-only prefetch (no binary annotations).
+# Requires podman. Output lands in .hermeto-out/ (gitignored).
+# The extra GIT_COMMON_DIR mount handles git worktrees: .git is a file pointing
+# to the main repo, so Hermeto needs both paths visible inside the container.
+HERMETO_IMAGE ?= ghcr.io/hermetoproject/hermeto:0.56.0
+hermeto-prefetch:
+	GIT_COMMON=$$(cd "$$(git rev-parse --git-common-dir)" && pwd -P) && \
+	podman run --rm \
```
r0x0d: If I may suggest something that we have in our okp-mcp project: https://github.com/rhel-lightspeed/okp-mcp/blob/main/Makefile#L3-L8

That feels much nicer for folks that have docker but not podman

Containerfile
owtaylor: You'd think so - but @samdoran  and I have tried previously to no luck. After trying again, I threw the issue at Gemini with some code snippets, and it was able to to figure out the problem is that dist_name doesn't get from hatch_vcs =-> setuptools_scm, and you can work around it in pyproject.toml with:

```
 [tool.hatch.version]
 source = "vcs"

+[tool.hatch.version.raw-options] 
+dist_name = "linux-mcp-server"
```

(raw-options are options to pass to the versioning backend.)

samdoran: I just wasn't sure what the string was supposed to be and got tired of fighting with it at the time. The docs don't make it very clear. Is `DISTRIBUTED` the package name in that example @r0x0d? And in the case of a package with dashes, is that supposed to normalized to underscores? Maybe it's supposed to be `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LINUX_MCP_SERVER`?

owtaylor: Once you apply the workaround above, yes, it's `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LINUX_MCP_SERVER`

samdoran: Maybe that's why I could never get it working. It wasn't obvious that extra setting was required.

owtaylor: Newest version has the workaround and `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LINUX_MCP_SERVER`

Makefile
owtaylor: Thanks - added that!

.tekton/pull_request.yaml
owtaylor: Hmm, probably! It would require some small changes to slushy-build which currently updates PipelineRun objects but not Pipeline objects. Since the updating is done by slushy-build in any case, I'll leave it as is for now.

owtaylor:
Changes since first version:

 * Moved tricky bits for rpm/pip installation out of Containerfile into scripts - thinking forward to a shared `build-common/` submodule. 
 * Made `rpms.in.yaml` the canonical place to declare what is installed - the Containerfile uses this rather than having a duplicated list. (Magic comments and using awk to extract package names - ugh - but don't have a good alternative.)
 * Added workaround to make `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LINUX_MCP_SERVER` work.
 * Added logic from okp-mcp for autodetecting the container-runtime

Containerfile
r0x0d: Yeah, the package name is `distributed`, and it would be like this `SETUPTOOLS_SCM_PRETEND_VERSION_FOR_LINUX_MCP_SERVER` for us. But glad you folks figured it out!

r0x0d (approved):

PR #566, opened 2026-09-02
Pin FastMCP to version 3

owtaylor:
Our server middleware will require changes to handle FastMCP 4 and Version 2 of the Python SDK. Until we do that, pin FastMCP to version 3.


Jazzcort (approved):
LGTM! 😁

PR #568, opened 2026-09-03
feat(tools): add PCP historical performance data tools

panyamkeerthana:
Add four new mcp tools for querying historical performance data from Performance CoPilot (PCP) archives:

- list_pcp_metrics: list available metrics with descriptions
- list_pcp_archives: show archive time periods
- query_pcp_metrics: query historical metric data
- get_performance_summary: system performance snapshot


src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+from linux_mcp_server.commands import CommandSpec
+from linux_mcp_server.commands import get_command
+from linux_mcp_server.server import mcp
+from linux_mcp_server.utils.decorators import disallow_local_execution_in_containers
+from linux_mcp_server.utils.types import Host
+
+
+@mcp.tool(
+    title="List available PCP metrics",
+    description="Returns PCP metrics available on the system with descriptions. Accepts an optional keyword to filter results.",
```
owtaylor, line 18: The size of the metric set is just really challenging here:

``` pminfo -t  | wc -c
187794
```

So that's like 45k tokens. Maybe say "Results will be large; if possible, use a subagent to identify appropriate metrics". I'm really not sure that substring searching is an effective strategy for finding the best metrics -

```
$ pminfo -t  | grep -i mem | wc -c
47074
```

Still large, and includes largely irrelevant stuff. (`xfs.inode_ops.ig_attempts [number of in memory inode lookup operations]`) And sometimes there are false negatives too - say you wanted to find available disk space - And what would be a good keyword to find available free disk space? I'm not sure `filesys.free` would be found.

Maybe we should just remove the keyword search and count on either the model *knowing* the right metric or using a subagent?

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+    tags={"fixed", "performance", "pcp"},
+    annotations=ToolAnnotations(readOnlyHint=True),
+)
+@log_tool_call
+@disallow_local_execution_in_containers
+async def list_pcp_metrics(
+    search_keyword: t.Annotated[
+        str | None,
+        Field(
+            description="Optional keyword to filter metrics",
```
owtaylor, line 28: If we keep this, I'd be explicit - `description="Case insensitive string to search for in metric names and descriptions"`

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+    if not stdout.strip():
+        return "No PCP archives found. Ensure pmlogger is running"
+
+    index_files = [line.strip() for line in stdout.strip().split("\n") if line.strip()]
+
+    results = []
+    for index_file in index_files:
+        base = index_file.replace(".index", "")
+        dump_cmd = get_command("pcp_archive_info")
+        returncode, dump_stdout, dump_stderr = await dump_cmd.run(host=host, archive=base)
```
owtaylor: I think running commands over ssh in a loop is generally something we want to avoid - we could use a command like:

```
  find /var/log/pcp/pmlogger -name *.index -type f -print -exec pmdumplog -L \{\} \;
```

Though even there, I wonder if we need a limit? Could a system have day-by-day archives for the last 5 years on it or something? If so, then we'd need to go with small shell script to limit to the most recent 50 log files or something like that.

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+        if returncode == 0:
+            results.append(dump_stdout.strip())
+        else:
+            results.append(f"Archive: {base}\n  Error reading archive: {dump_stderr.strip()}")
+
+    return "\n\n".join(results)
```
owtaylor, line 95: Whether this stays here or moves to get_system_information(), I'd rather parse the result and return structured information than just dumping the results to the caller.

A question we need to figure out for linux-mcp-server in general is timezone for times. (Related #576) When a timestamp is *provided* or *returned* is it a UTC time? A timezone in a specific time? My proposal would be:
 
 * Returned times are always in the timezone of the target system, formated as rfc-3339 times with an explicit offset 
 * Passed in times without an offset are treated in the timezone of the target systems, passed in times with an offset the offset is honored.

Need to verify what pcp uses - it looks like local timestamps without an explicit offset, unless `-z` is supplied, then it's UTC times without an explicit offset. (Note that we do *not* want to return times relative to the local timezone where the MCP server is running - this is irrelevant. So that makes things more complex. But the basic thing we're trying to do is make sure that say correlating log timestamps with PCP timestamps works correctly.)

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+
+    archive = ""
+    for line in stdout.split("\n"):
+        if "pmlogger" in line:
+            full_path = line.strip().split()[-1]
+            archive = full_path.rsplit("/", 1)[0]
+            break
+
+    if not archive:
+        return "No PCP archives found. Ensure pmlogger is running."
```
owtaylor, line 150: I don't particularly like parsing the free form text output of `pcp` like this. Can we find an alternative? If not, I'd name the command something that better represents "run pcp, get whatever it dumps on stdout" than "pcp_archive_dir"

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+    title="List available PCP archive time periods",
+    description="Returns available PCP archive time periods with start and end times for each archive.",
+    tags={"fixed", "performance", "pcp"},
+    annotations=ToolAnnotations(readOnlyHint=True),
+)
+@log_tool_call
+@disallow_local_execution_in_containers
+async def list_pcp_archives(
+    host: Host = None,
+) -> str:
```
owtaylor, line 66: I don't think listing the archives is that useful, especially without the ability to pass them to query_pcp_metrics(). You are already passing the whole directory to `pmrep` - and `pcp-xsos` will accept a directory as well along with a time (Needs some checking to make sure that `pcp xsos` is doing the right thing .)

Reporting timeranges, but I think the way I'd work it is report a unified PCP status for the system, *including* the time ranges available for $(hostname). My idea was to have that in get_system_info()

```json
"pcp": {
   "installed": true,
   "pmcd_running": true,
   "pmlogger_running": true,
   "available_time_ranges": [{
       "start": "2026-08-01T11:01:00-04:00",
       "end": "2026-09-10T11:08:00-04:00",
   }]
}
```

The backend here would be something like:

```
sh -c "sh -c 'which pmlogger ; systemctl is-active pmlogger pmcd ; pmlogdump -l /var/log/pcp/pmlogger/$(hostname)'"
```

The processing would *collapse* ranges that are adjacent with some grace period (1 hour?). I'm a bit worried that computing the ranges could be quite slow if there are a couple of years of data on the system, but it's probably still pretty fast on a LLM scale.

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+    ] = "1minute",
+    host: Host = None,
+) -> str:
+    """Query historical PCP metric data from archives."""
+    # Find the latest archive from pcp output
+    cmd = get_command("pcp_archive_dir")
+    returncode, stdout, stderr = await cmd.run(host=host)
+
+    if returncode != 0:
+        return "PCP is not installed on this system."
```
owtaylor, line 140: Errors should be raised as ToolError(), not just be freeform text.

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+        if "pmlogger" in line:
+            full_path = line.strip().split()[-1]
+            archive = full_path.rsplit("/", 1)[0]
+            break
+
+    if not archive:
+        return "No PCP archives found. Ensure pmlogger is running."
+
+    base_cmd = get_command("pcp_query_metrics")
+    full_cmd = CommandSpec(args=base_cmd.args + tuple(metrics))
```
owtaylor, line 153: This would allow a LLM to smuggle in other options, like output files, breaking the readonly nature. The metrics names need to be validated. Something like:

 Start a-zA-Z
 Continues: a-zA-Z0-9_.

As a regexp: `'^[a-zA-Z][a-zA-Z0-9_.]*$`

I would also use `pmrep -o csv`, since pmrep doesn't always get column widths right:

```
  f.free  f.free
  /dev/s  /dev/s
   Kbyte   Kbyte
     N/A     N/A
     xxx  195420
     xxx  195420
```

vs:

```
Time,"filesys.free-/dev/sda3","filesys.free-/dev/sda2"
2026-09-09 15:52:31,,
2026-09-09 16:52:31,13333556,195420
2026-09-09 17:52:31,13233428,195420
2026-09-09 18:52:31,13231464,195420
```

I *think* I would also reformat as json:

```json
[{
   "time": "2026-09-09T15:52:31-0400",
   "metrics": {
        "filesys.free-/dev/sda3": "13333556"
        "filesys.free-/dev/sda2": "195420"
 },{
   "time": "2026-09-09 17:52:31-0400",
   "metrics": {
        "filesys.free-/dev/sda3": "13233428"
        "filesys.free-/dev/sda2": "195420"
}]
```

This obviously increases tokens a lot, but my (complete guess) is that an LLM will have an easier time instead of a wall of numbers. 

Whether we return CSV or JSON to the LLM, we need to strictly control the amount of data returned to avoid context filling surprises. My proposal would be:

 - If all three of start/end/interval are provided, you get what you ask for, even if it is huge
 - Otherwise, we fill in the missing start/end/interval to give ~20 items

Algorithm is roughtly:

```
    set timespan to be 1h, or if interval is provided interval * 20
    default start to timespan ago, or timespan before end, if end is provided
    default end timestamp after start
    default interval to "round number greater than (end - start) / 20" ... 5s, 10s, 1m, 5m, 10m, 30m...
```

src/linux_mcp_server/tools/pcp.py
```diff
@@ -0,0 +1,206 @@
...
+    tags={"fixed", "performance", "pcp"},
+    annotations=ToolAnnotations(readOnlyHint=True),
+)
+@log_tool_call
+@disallow_local_execution_in_containers
+async def get_performance_summary(
+    archive_path: t.Annotated[
+        str | None,
+        Field(
+            description="Optional PCP archive path for historical summary. If not provided, shows live system summary.",
```
owtaylor, line 185: Optional timestamp "for summary at a particular time" - not archive path.

owtaylor (changes requested):
This looks quite good - various comments below. One of the main things is that I think we should completely hide archive files from the interface - only time should matter, and the only host we should care about is localhost.

I'd also rename the tools so that they start with pcp_* rather than having pcp in the middle somewhere, and have a note on all of them "Use this only if get_system_information() indicates PCP is available" (see my proposal for what get_system_information should return.)

src/linux_mcp_server/tools/pcp.py
panyamkeerthana: 100% agree returning this large of a metric set can be expensive. I tried `pminfo -t {keyword}` but that isn't a text search its filtering by metric prefix so I went with this substring filtering approach

 I think **pmsearch** is more in line with what we're looking for since it does full text search but it requires a valkey server with the valkey-search module configured. Do you think we should consider using it?

For the model knowing the right metric I'm not sure how reliably that would work across different models but the subagent approach seems promising, I'll look into both🫡

src/linux_mcp_server/tools/pcp.py
panyamkeerthana: pcp_archive_dir is a little misleading 😅 Initially I used ` pcp | grep pmlogger | awk '{print $NF}'` to get the parent archive directory from the active pmlogger archive path but switched to running `pcp` and doing the parsing directly to stay consistent w the codebase and avoid shell piping.

src/linux_mcp_server/tools/pcp.py
panyamkeerthana: pcp outputs local system timestamps without an timezone offset `-z` sets the rendering context to the hosts archive timezone, but leaves out offset strings

PR #570, opened 2026-09-04
ci: Use locked versions for functional tests

owtaylor:
Instead of having the functional tests use the latest versions from PyPI, use the locked versions, in the same way as the rest of our tests.

This:

 - Separates concerns. While we do care about working with latest pypi versions (for users that do `uv tool install`), the functional tests are meant to answer "does our code work with a real RHEL system"
 - Avoids pull request CI failing for versioning problems unrelated to the pull request.
 - Gives better consistency with our containers, which use the locked dependency versions. (We'll shortly be building RPMs that way as well.)


owtaylor:
@coderabbitai review

owtaylor:
@coderabbitai review

PR #580, opened 2026-09-08
ci: don't run functional tests on changes to irrelevant directories

owtaylor:
The functional tests require manual approval for changes for non-core contributors and wait for a test system to be assigned for a long time, so skip them for changes that obviously won't affect the result.


panyamkeerthana (approved):
LGTM!

PR #598, opened 2026-09-16
fix: execute the newly-supplied script on confirmation mismatch

### LINKED ISSUE #594: run_script_with_confirmation executes the stale stored script, not the newly-supplied one, on parameter mismatch
### Summary

`run_script_with_confirmation` executes the **stale, originally-validated script** instead of the newly-supplied one whenever a caller reuses a token but supplies different script/description/readonly parameters. The gatekeeper correctly re-validates the *new* script and returns OK, but the command actually wrapped and executed is still the *old* one. No error is raised anywhere in the chain.

### Version

`linux-mcp-server` 1.6.0 (also present on current `main`, confirmed against commit `22de54f3`).

### Root cause

`src/linux_mcp_server/tools/run_script.py`, `run_script_with_confirmation`:

```python
execute_details = new_details if details_changed else script_details   # correctly computed...

if details_changed:
    gatekeeper_result = await check_run_script(description, script_type, ..., readonly=readonly)  # ...validates the NEW script...
    ...

try:
    command = _wrap_script(script_details)                       # <-- BUG: wraps the OLD script
    returncode, stdout, stderr = await execute_command(command, host=execute_details.host)  # only .host is new
```

`execute_details` is computed specifically to hold whichever `ScriptDetails` should actually be acted on, and is used correctly for `host=execute_details.host` on the very next line — but the script content passed into `_wrap_script()` still comes from the stale `script_details`. The `readonly` flag used for sandboxing is also taken from the stale object, not the new value.

### How it was introduced

`git blame`/`git show` against this repo:

- **Correct, from #417** (`f0b66ba9`): `command = _wrap_script(execute_details.script_type, execute_details.script)`.
- **Regressed in `658c14de`** ("Merge commit from fork", 2026-09-04) — that commit's actual goal was unrelated (fixing readonly-sandbox args), and changed `_wrap_script()`'s signature from `(script_type, script)` to a single `ScriptDetails` object across 4 call sites. Three were updated correctly. The fourth, in `run_script_with_confirmation`, became `_wrap_script(script_details)` instead of `_wrap_script(execute_details)` — the one call site with two candidate objects in scope, and the refactor picked the wrong one.

### Why the existing test doesn't catch it

`TestRunScriptWithConfirmationMCP.test_mismatch_revalidates_and_executes` mocks `execute_command`'s return value to a fixed `(0, "out", "")` regardless of input, and only asserts on that return value plus the fact the gatekeeper was called once — it never inspects `execute_command.call_args` to check which script was actually passed through. It passes identically whether the bug is present or not. The sibling test for `run_script_interactive`'s mismatch path (`test_mismatch_revalidates_new_id`) *does* correctly assert the new content is used — this test just never got the equivalent assertion.

A test that would catch it:
```python
args, kwargs = patch_execute_command.call_args
wrapped_command = args[0]
assert "<new script marker>" in wrapped_command[2]
assert "<old script marker>" not in wrapped_command[2]
```

### Impact

Under a realistic trigger (an AI agent or script reusing a token while polling/rechecking status, supplying a new — often read-only — script each time), the server silently re-executes whatever was originally validated for that token instead. In our own testing this produced two concrete failure modes:
1. A harmless reused token (`rpm -q ...`) caused five subsequent, materially different calls to silently no-op while reporting `status=success`.
2. A token originally issued for a `systemctl stop <unit>` cleanup script got reused for what the caller believed was a read-only status-poll call — the gatekeeper approved the new read-only script, but the actual command executed was the stale `systemctl stop`, killing a genuinely-running, long-lived job (confirmed via `SIGTERM`/`status=15` in the unit's own journal, matching the stale script's content exactly, not a generic timeout signature).

### Suggested fix

```diff
     try:
-        command = _wrap_script(script_details)
+        command = _wrap_script(execute_details)
         returncode, stdout, stderr = await execute_command(command, host=execute_details.host)
```

Plus hardening `test_mismatch_revalidates_and_executes` to assert on `execute_command.call_args` (see above) so this exact class of regression can't silently reappear in a future refactor of this function.

### Reproduction

Verified deterministically via direct MCP protocol calls (bypassing any LLM/agent client): `validate_script` a script A, then `run_script_with_confirmation` reusing that token with a different script B. Pre-fix: script A's effect is observed, not B's. Post-fix (one-line change above, applied locally): script B's effect is observed, not A's.

owtaylor:
Fixes #594.

## Problem

`run_script_with_confirmation` executed the **stale, originally-validated script** instead of the newly-supplied one whenever a caller reused a token with different script/description/readonly parameters. The gatekeeper correctly re-validated the *new* script and returned OK, but `_wrap_script()` was handed the stale `script_details` rather than `execute_details` — so the old script ran, under the old `readonly` sandbox flag, with no error anywhere in the chain.

Regressed in `658c14de`, which changed `_wrap_script()`'s signature from `(script_type, script)` to a single `ScriptDetails` object across four call sites; this was the one call site with two candidate objects in scope, and the refactor picked the wrong one.

## Fix

```diff
-        command = _wrap_script(script_details)
+        command = _wrap_script(execute_details)
```

`execute_details` is already the object computed to hold whichever details should be acted on, and is used on the very next line for `host=`.

## Test

`test_mismatch_revalidates_and_executes` previously mocked `execute_command`'s return value and only asserted on that value plus a single gatekeeper call — it passed identically with or without the bug. It now inspects `execute_command.call_args` and asserts the new script body is what got wrapped, the stale one is not, and that the new `readonly=True` drives the systemd sandbox args.

Verified TDD: the hardened test fails against the unfixed code (`assert 'print(99)' in "...SCRIPT='print(3)'..."`) and passes with the fix. All 33 tests in `tests/tools/test_run_script.py` pass.


panyamkeerthana (approved):
LGTM. Great fix and test improvements!

PR #599, opened 2026-09-16
fix: pass a static --description to systemd-run

### LINKED ISSUE #595: systemd-run's auto-derived Description= corrupts the transient unit file for any multi-line wrapped script
### Summary

`systemd-run`'s auto-derived `Description=` doesn't escape embedded newlines, so any **multi-line** script routed through `run_script`/`run_script_with_confirmation`'s sandboxed execution path (`sudo systemd-run ... bash -c "$SCRIPT"`) corrupts its own transient unit file's `[Unit]` section. The corruption stays dormant until something re-parses the file — in practice, whenever the wrapped script itself calls `systemctl daemon-reload`, which then fails re-parsing its own now-corrupted definition and desyncs `systemd-run --wait`'s completion tracking for that unit.

### Version

`linux-mcp-server` 1.6.0 (also present on current `main`, confirmed against commit `22de54f3` — `SYSTEMD_RUN_ARGS` still has no `--description`).

### Root cause

`src/linux_mcp_server/tools/run_script.py`:

```python
SYSTEMD_RUN_ARGS = [
    "--quiet",
    "--pipe",
    "--collect",
    "--wait",
    "--property=WorkingDirectory=/tmp",
    "--property=PrivateTmp=true",
    "--property=NoNewPrivileges=true",
]
SYSTEMD_RUN_COMMAND = "/usr/bin/sudo /usr/bin/systemd-run {args}"

WRAPPER_TEMPLATE = """\
set -euo pipefail
SCRIPT={script}
if command -v sudo >/dev/null 2>&1 && command -v systemd-run >/dev/null 2>&1 && sudo -l whoami >/dev/null 2>&1; then
  exec {systemd_run_command} {script_type} -c "$SCRIPT"
else
  exec {script_type} -c "$SCRIPT"
fi
"""
```

No `--description` is ever passed to `systemd-run`, so it auto-derives the transient unit's `Description=` from the command it's about to run — here, `bash -c "$SCRIPT"` where `$SCRIPT` is the caller's (possibly multi-line) script text. `systemd-run`'s auto-derivation does **not** escape embedded newlines the way its own `ExecStart=` line correctly does (`\n` literal escapes) — so a multi-line command produces a `Description=` field with raw embedded newlines, splitting one directive into several bogus non-`key=value` lines in the generated `/run/systemd/transient/run-uNNN.service`.

### Concrete evidence — an actual corrupted unit file

A 4-line "clear stale unit" script (ending in `systemctl daemon-reload`) produced this generated unit file, pulled directly off the host after a real failed run:

```ini
[Unit]
Description=/bin/bash -c set -euo pipefail; systemctl stop leapp-preupgrade-helper 2>/dev/null || true
systemctl reset-failed leapp-preupgrade-helper 2>/dev/null || true
rm -f /run/systemd/transient/leapp-preupgrade-helper.service /etc/systemd/system/leapp-preupgrade-helper.service
systemctl daemon-reload
CollectMode=inactive-or-failed

[Service]
WorkingDirectory=/tmp
PrivateTmp=yes
NoNewPrivileges=yes
ExecStart=
ExecStart=@/bin/bash "/bin/bash" "-c" "set -euo pipefail; systemctl stop leapp-preupgrade-helper 2>/dev/null || true\nsystemctl reset-failed leapp-preupgrade-helper 2>/dev/null || true\nrm -f /run/systemd/transient/leapp-preupgrade-helper.service /etc/systemd/system/leapp-preupgrade-helper.service\nsystemctl daemon-reload"
```

Note the contrast within the same file: `ExecStart=` correctly represents the multi-line script as a single line with literal `\n` escapes; `Description=` does not, and its raw content spills across 4 lines — one of which (`systemctl reset-failed ...`) contains no `=` character at all, an invalid `[Unit]` directive.

### Why it stays dormant until `daemon-reload`

The corrupted file is inert once written — starting the unit from it doesn't itself trigger a re-parse. But the script's own `systemctl daemon-reload` does, from *inside* the still-running transient unit whose own definition is the corrupted one:

```
systemd[1]: /run/systemd/transient/run-u222.service:4: Missing '='.
systemd[1]: Reloading.
systemd[1]: /run/systemd/transient/run-u222.service:4: Missing '='.
systemd[1]: run-u222.service: Current command vanished from the unit file, execution of the command list won't be resumed.
systemd[1]: run-u222.service: Succeeded.
```

The reload re-parses the unit, hits the malformed line, and loses track of the currently-executing `ExecStart` command for it — consistent with `systemd-run --wait` not reliably signaling completion afterward, producing a client-side command timeout on an otherwise-successful script. Any multi-line script through this execution path is at risk; `daemon-reload` is simply the most common realistic trigger (an ordinary thing for a "clear this unit before relaunching" script to call), not the only possible one.

### Impact

In our own testing, this corruption produced a client-side timeout on a genuinely-successful script, which an AI agent responded to by fabricating fictional follow-up job runs (including fake PIDs) instead of reporting the real error — the underlying corruption is what created the confusing failure surface, though that's a separate model-behavior concern from this bug itself.

### Suggested fix

One-line addition to `SYSTEMD_RUN_ARGS`:

```diff
 SYSTEMD_RUN_ARGS = [
     "--quiet",
     "--pipe",
     "--collect",
     "--wait",
+    "--description=linux-mcp-server-wrapped-script",
     "--property=WorkingDirectory=/tmp",
     "--property=PrivateTmp=true",
     "--property=NoNewPrivileges=true",
 ]
```

Gives `systemd-run` a fixed, single-line description so it never needs to auto-derive one from the (possibly multi-line) wrapped command. No other behavior changes — `ExecStart=` (and therefore what the script does) is unaffected. (A more thorough fix would be newline-escaping in `systemd-run`'s own `Description=` auto-derivation, but that's upstream in `systemd` itself and out of scope here.)

### Reproduction

Verified deterministically via direct MCP protocol calls (bypassing any LLM/agent client), using the exact "clear stale unit" script above against a real host. Pre-fix: journal shows the `Missing '='`/`vanished from the unit file` messages on every attempt. Post-fix (same call, same script, same host, service restarted with the one-line patch applied): clean journal, no error. Regression-checked with a single-line readonly script, a different multi-line readonly script with no `daemon-reload`, and a readonly script against a second host — all clean.

owtaylor:
Fixes #595.

When `systemd-run` isn't given an explicit `--description`, it derives `Description=` for the transient unit from the command line it is about to run. Before systemd commit [9ce3440a6fb3](https://github.com/systemd/systemd/commit/9ce3440a6fb3fb785e0fa3a6fcf60af7ef652440) that derivation didn't escape embedded newlines, even though the `ExecStart=` line it writes right below does. That fix is in RHEL 10, but was not backported to RHEL 8 or RHEL 9.

Because we run scripts as `bash -c "$SCRIPT"`, any multi-line script makes `Description=` spill across several lines of the generated `/run/systemd/transient/run-uNNN.service`, leaving lines with no `=` in them. The bad file is inert until something re-parses it — in practice the script's own `systemctl daemon-reload`, issued from inside the very transient unit whose definition is corrupt. systemd then logs `Missing '='` followed by `Current command vanished from the unit file, execution of the command list won't be resumed`, loses track of the running command, and `systemd-run --wait` never reports completion — so a script that actually succeeded surfaces to the client as a timeout.

Passing a fixed description avoids the auto-derivation entirely.

Doing this only for scripts that would break parsing was considered and rejected: it would depend on the exact behavior of the buggy systemd versions and on systemd's parsing rules, and it isn't obviously better anyway — having the unit say the command came from linux-mcp-server is useful on its own.

## Testing

`tests/tools/test_run_script.py` gains a case asserting the wrapped command carries the fixed, shell-quoted description; the full file passes.


panyamkeerthana (commented):
LGTM on the fix, passing a static `--description` prevents the corruption

panyamkeerthana (approved):

Issue #590, opened 2026-09-11
run_script_with_confirmation silently executes locally when host is omitted


lifetoldme:
### Summary

When a script is validated with `validate_script(host="<target>", readonly=false)` and then executed via `run_script_with_confirmation(...)` **without** repeating `host`, the script runs **locally on the MCP host** instead of the validated target. It should either error on the mismatch or fall back to the stored host.

### Version

`linux-mcp-server` 1.5.0 (pipx install; client: opencode).

### Steps to reproduce

1. `validate_script(description="...", script_type="bash", script="hostname", host="target-host", readonly=false)` returns a token.
2. `run_script_with_confirmation(description="...", script_type="bash", script="hostname", readonly=false, token=<token>)` - note: no `host` argument (schema default is `null`).
3. Observe: the script runs on the MCP host (prints its hostname), not `target-host`.

### Expected

Execution on `target-host` (the host stored with the validation token), or a clear error that the parameters do not match the validated script.

### Actual

Execution locally (`host=None`). In `tools/run_script.py`, `run_script_with_confirmation` builds `new_details` from the incoming arguments and compares it with the stored `script_details`. Since `host` defaults to `None`, `details_changed` is true; the code revalidates and executes `execute_details = new_details`, i.e. `host=None` -> local execution via `execute_command(command, host=None)`. `run_script_interactive` has the same pattern.

### Impact

A mutating script can appear to succeed while touching the wrong machine (or fail with confusing errors like `zpool: command not found`) - easy to miss because the output is otherwise normal. `run_script` (read-only path) is unaffected: it has no `host` parameter and always uses the stored host, which made the inconsistency harder to spot.

### Suggested fix

One of:
- Default `host` to the stored `script_details.host` when the caller omits it (only treat genuinely different values as a change), or
- Treat a host omission/mismatch as a validation error instead of silently executing the new details, or
- Make `host` required on `run_script_with_confirmation` / `run_script_interactive` so the mismatch cannot happen silently.

### Workaround

Repeat `host` on the confirmation call exactly as it was passed to `validate_script`.

pbuchan-rh:
Confirming this from a different angle — we hit the same behavior and traced it one layer further down, to the authorization/policy layer rather than just `run_script.py` itself.

[...]

PR #602, opened 2026-09-17
feat(config): add LINUX_MCP_HOST_MODE to control where tools may run

owtaylor:
Stacked on #601 — base is `hostname-required`, so review that one first. Only the second commit belongs to this PR.

Every tool takes a `host`, and until now the server accepted whatever the model passed and only found out later whether it could act on it. Where the answer was known up front — in a container, or on a platform where local execution isn't implemented — nothing told the model, so it would try `localhost` and get an error back (#590).

This adds a three-way `LINUX_MCP_HOST_MODE`: `local-only`, `remote-only`, or `any`.

The mode shapes what the model is *told* as well as what it may do:

- it picks the **Target host** bullet in the server instructions;
- under `remote-only`, the `host` parameter in `tools/list` is described as the remote system to reach over SSH, and says the server cannot run anything locally;
- under `local-only`, the `host` parameter is dropped from the tool schemas entirely. There is only one value it could have, and asking the model to recite it on every call is only a chance to get it wrong. The middleware fills it in, so a client working from an older schema that still passes `host: localhost` keeps working.

A host outside the mode is refused, never redirected — running somewhere other than where the model asked would answer a question nobody posed.

The default is `remote-only` when the server runs in a container or on a platform other than Linux, and `any` otherwise. A container sees its own filesystem, process table and units rather than the host's, so local execution there gives the user nothing useful.

That default replaces `@disallow_local_execution_in_containers`, which had to be repeated on all 22 tools and was easy to leave off a new one. The mode is checked once, in `resolve_target_host()`, which every tool call already goes through; `restrict()` additionally takes the corresponding capability out of the `ExecutionContext`, so the mode still holds if anything ever reaches execution by another route.

Everything mode-related — schema adjustment, instructions text, error wording, and enforcement — lives together in `target_host.py`, so the picture given to the model and what the server actually does can't drift apart.

Tested with new unit tests in `tests/test_host_mode.py` and functional tests in `tests/functional/general/test_host_mode.py`.


owtaylor:
This is meant to be a stacked PR, but it ended up cross-repo with the first branch in origin/ and the second branch in my fork - which doesn't work for stacked PR's. Closing, will refile in a bit.

